PR #23, opened 2018-04-15
Reducing exception from fatal to notification on shipping method problem

mccottry:
We do not want the code to completely fail if there is a problem with a single shipping option.

Reasoning:
The idea of our plugin is to increase conversions.  It is not to make sure that FedEx, or ShippingHQ, or UPS, or whatever plugin is working.  There can be any number of reasons why one maybe down, like internet problems, or strict address validation for a particular region, whereas other shipping methods will not fail.

If any given shipping method fails for whatever reason, it is good to note it, but we must present the customer with all available shipping options so that they can check out of Bolt successfully and increase our conversion rates.

When making a decision like this, it may be good to consider what the underlying system would do.  In this instance, we can ask ourselves, "If FedEx was down, would the Magento one page checkout completely breakdown and leave the customer with no way to checkout, or would it give the customer all the available options that it is aware of?"

Options are better than no options in this instance.

### DIFF
--- a/app/code/community/Bolt/Boltpay/Helper/Api.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Api.php
@@ -793,7 +793,8 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
 
         foreach ($rates as $rate) {
             if ($rate->getErrorMessage()) {
-                throw new Exception("Error getting shipping option for " .  $rate->getCarrierTitle() . ": " . $rate->getErrorMessage());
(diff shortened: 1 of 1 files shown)


johnmccombs1:
Yep. I'm in complete agreement with this. I do think we should probably have another task for a long term fix of giving the customer some notification that their address may be wrong. For example with ATBBQ, if they have a failure with ShipperHQ none of those shipping methods will come back, but if they had flat rate setup, it may return a result. This will make it look like there is only one valid shipping rate that might take awhile to ship, when the reality is that they just put in a bad address. I could see a notification in cases where there was an error with any shipping method that we say "We were unable to calculate some shipping methods. You may want to check your address and try again..."

daisy1754:
@johnmccombs1 I raised that to Bolt SF team.

PR #25, opened 2018-04-17
Added logging for shipping rates.

johnmccombs1:


app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -235,7 +235,7 @@ public function createOrder($reference, $session_quote_id = null)
             }
 
             if (!$is_shipping_set) {
-                $errorMessage = 'Shipping method not found';
+                $errorMessage = 'Shipping method not found. $service: ' . $service . ' $shipping_address: ' . var_export($shipping_address->debug(), true);
```
daisy1754: let's put variables in metadata instead of dumping it into message body, so it's easier to read in bugsnag

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -273,7 +273,19 @@ public function createOrder($reference, $session_quote_id = null)
...
+
+        try {
+            $service->submitAll();
+        } catch (Exception $e) {
+            $errorMessage = 'Error submitting order';
+            $metaData = array(
+                'transaction'   => json_encode((array)$transaction),
+                'quote_address' => var_export($quote->getShippingAddress()->debug(), true)
+            );
+            Mage::helper('boltpay/bugsnag')->notifyException(new Exception($errorMessage), $metaData);
```
daisy1754: (non-blocking) It's ok for now but I think this would cause multiple error reports (one from here, and the another from upstream). I think what we can do in the long term is defining custom error type so we can propagate all info to upper level and we bugsnag at the top level.

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -807,9 +819,15 @@ public function getShippingAndTaxEstimate( $quote )
...
+            $rateCode = $rate->getCode();
+
+            if(empty($rateCode)) {
+                Mage::helper('boltpay/bugsnag')->notifyException(new Exception('Rate code is empty. ' . var_export($rate->debug(), true)));
+            }
+
             $option = array(
                 "service"   => $label,
-                "reference" => $rate->getCarrier() . '_' . $rate->getMethod(),
+                "reference" => $rateCode,
```
daisy1754: This seems functional change, not logging change. Maybe this fixes the issue? 
The reference field is currently used later at 
> $quote->getShippingAddress()->setShippingMethod($referenceShipmentMethod)->save();

johnmccombs1: I couldn't find any examples where the rate code wasn't the same as the carrier and method name, so I swapped with this since the rate code is actually what Magento expects as far as I can tell.

app/code/community/Bolt/Boltpay/Helper/Api.php
johnmccombs1: I'll change and re-push. Good call.

PR #31, opened 2018-04-23
Added partial capture feature.

johnmccombs1:


app/code/community/Bolt/Boltpay/controllers/ApiController.php
```diff
@@ -178,4 +180,11 @@ public function hookAction()
         }
     }
 
+    protected function getCaptureAmount($transaction) {
+        if(isset($transaction->capture->amount->amount) && is_numeric($transaction->capture->amount->amount)) {
+            return $transaction->capture->amount->amount/100;
+        }
+
+        return null;
```
daisy1754: if you return null, will it cause exception in later execution of 
> $order->getGrandTotal() > $captureAmount
?

PR #32, opened 2018-04-24
Shipping and tax prefetch

YevhenShBolt:
I added an ability to fill the estimate form after getting a request from freegeoip.
Then if repeated call exists to "prefetchEstimateAction" and in "quote" exists an address data, it should return the empty array


YevhenShBolt:
Added the cache key generation for prefetch.

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -168,39 +171,69 @@ public function indexAction()
...
+     * @return string
+     */
+    public function generateCacheKey($quoteId, $_shippingAddress, $totalAmount)
+    {
+        $quoteTotalAmount = round($totalAmount*100);
+        $countryCode = $_shippingAddress->getCountryId();
+        $region = $_shippingAddress->getRegionCode();
+        $postCode = $_shippingAddress->getPostcode();
+
+        $key = $quoteId . '_' . $quoteTotalAmount . '_' . $countryCode.$region.$postCode;
```
daisy1754: Does $region ever contain any unsafe character for cache key (eg., whitespace)? Maybe we want to hash this string here?

Also nit I prefer to insert '_' between countrycode-region and region-postcode for consistency with

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -168,39 +171,69 @@ public function indexAction()
...
+        $this->getResponse()->setBody($response);
+    }
+
+    /**
+     * @param $quoteId
+     * @param $_shippingAddress
+     * @param $totalAmount
+     * @return string
+     */
+    public function generateCacheKey($quoteId, $_shippingAddress, $totalAmount)
```
daisy1754: sorry I'm keep saying this but I prefer we include skus here. Think about $50 gift card vs $50 chair. Price is same, shipping cost would be different.

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -168,39 +171,69 @@ public function indexAction()
...
+            $quote->getBillingAddress()->addData($address_data);
+
+            try {
+                $estimate_response = Mage::helper('boltpay/api')->getShippingAndTaxEstimate($quote);
+            } catch (Exception $e) {
+                $estimate_response = null;
+            }
+
+            Mage::app()->getCache()->save(serialize($address_data), "quote_location_" . $cacheKey);
+            Mage::app()->getCache()->save(serialize($estimate_response), "quote_shipping_and_tax_estimate_" . $cacheKey);
```
daisy1754: can you also include timestamp so we won't use stale cache? (let's say it'll expire in 10min)

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -132,11 +134,12 @@ public function indexAction()
             // Check session cache for estimate.  If the shipping city or postcode, and the country code match,
             // then use the cached version.  Otherwise, we have to do another calculation
             ////////////////////////////////////////////////////////////////////////////////////////
-            $cached_address = unserialize(Mage::app()->getCache()->load("quote_location_".$quote->getId()));
+            $cacheKey = $this->generateCacheKey($quote->getId(), $quote->getShippingAddress(), $quote->getGrandTotal());
```
daisy1754: should we also include if user is logged into magento? (I assume special customer class has different shipping treatment?)

daisy1754: Also we probably need to consider discount - freeshipping discount don't change ground total but changes shipping

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -132,11 +134,12 @@ public function indexAction()
...
+            $cached_address = unserialize(Mage::app()->getCache()->load("quote_location_" . $cacheKey));
 
             if ($cached_address && ((strtoupper($cached_address["city"]) == strtoupper($address_data["city"])) || ($cached_address["postcode"] == $address_data["postcode"])) && ($cached_address["country_id"] == $address_data["country_id"])) {
                 //Mage::log('Using cached address: '.var_export($cached_address, true), null, 'shipping_and_tax.log');
-                $response = unserialize(Mage::app()->getCache()->load("quote_shipping_and_tax_estimate_".$quote->getId()));
+                $response = unserialize(Mage::app()->getCache()->load("quote_shipping_and_tax_estimate_" . $cacheKey));
```
daisy1754: can you add HTTP response header `X-Bolt-Cache-Hit: Hit`?

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -132,11 +134,12 @@ public function indexAction()
...
             } else {
                 //Mage::log('Generating address from quote', null, 'shipping_and_tax.log');
                 //Mage::log('Live address: '.var_export($address_data, true), null, 'shipping_and_tax.log');
```
daisy1754: can you add HTTP response header `X-Bolt-Cache-Hit: Miss`?

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
danac-gs: What is the reason behind caching two values for every cache key? Could it be one, with $cacheKey as key and the shipping and tax options as the value?

app/design/frontend/base/default/template/boltpay/prefetch_shipping_quote.phtml
```diff
@@ -29,8 +29,42 @@ if($this->isBoltActive()) {
...
+                            } else if (region.tagName.toLowerCase() === 'select') {
+                                evt.initEvent('change', true, true);
+                                region.value = _data.region;
+                                country.dispatchEvent(evt);
+                            }
+                        }
+                    }
+                }
+            };
             xmlhttp.send(JSON.stringify(<?=$this->getLocationEstimate()?>));
```
danac-gs: Changing the page element values with the geo-location estimated ones might not be the expected functionality in many cases. Maybe a check if the values are empty before the change is needed. That seems to me like a no need feature, but I might be wrong. If I am not, I'd move the whole functionality out into the controller, and just call it from the page with a GET request. To be more precise, I'd make the controller responsible for geo-ip-location estimation call and calling the shipping and tax estimation prefetch with the result. That way you can later extend the prefetch to other addresses, not just geoloc one, but logged in customer address, and also the quote shipping address that might be already saved (user entered it on the checkout-shipping page, and then returned to cart).

daisy1754: I think I agree with Daniel here. we don't probably need change in this file

YevhenShBolt: Yes, it makes sense. But current functionality changes the value in estimate form only after prefetch was done. So, if in quote exist some shipping address data the form values do not be filled. This form is always empty if the customer did not enter any address previously.

Another I want to notice that if we moved this code from js, we can catch some delay before page render because we should wait for a response from GeoIp service.

Maybe I'm aggravating and I wrong, but i guess that this async request does not freeze checkout cart page for now.

app/design/frontend/base/default/template/boltpay/prefetch_shipping_quote.phtml
```diff
@@ -29,8 +29,42 @@ if($this->isBoltActive()) {
...
+                    var result = JSON.parse(xmlhttp.responseText);
+
+                    if (result.address_data && Object.keys(result.address_data).length > 0) {
+                        var evt = document.createEvent('HTMLEvents'),
+                            _data = result.address_data,
+                            postcode = document.getElementById('postcode'),
+                            country = document.getElementById('country'),
+                            region = document.getElementById('region');
+
+                        if (postcode) {
```
daisy1754: check if value already exists; not fill if it's there

app/design/frontend/base/default/template/boltpay/prefetch_shipping_quote.phtml
daisy1754: chatted off-the-thread. I'm ok to have this if you feel strongly about this (though I'm still not sure if it's our plugin's job to do this), but we need to check if the field already contain text. Currently, it is possible that user starts entering something in the field and the plugin asynchronously fetch address from server and override user's input.

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -168,39 +172,116 @@ public function indexAction()
...
+     * @return string
+     */
+    public function generateCacheKey($quote, $addressData)
+    {
+        /* @var $shippingAddress Mage_Sales_Model_Quote_Address */
+        $shippingAddress = $quote->getShippingAddress();
+        $itemIDs = $quote->getItemsCollection()->getAllIds();
+        $quoteTotalAmount = round($quote->getGrandTotal()*100);
+        $isCustomerGuest = $quote->getCustomerIsGuest();
+        $discount = round(abs($shippingAddress->getDiscountAmount()) * 100);
```
daisy1754: nit: this should be discount code, not amount

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -168,39 +172,116 @@ public function indexAction()
...
+            } catch (Exception $e) {
+                $estimate_response = null;
+            }
+
+            $lifeTime = 600; //10 minutes
+            Mage::app()->getCache()->save(
+                serialize($address_data),
+                'quote_location_' . $cacheKeyLocation,
+                array('BOLT_QUOTE_LOCATION'),
+                $lifeTime
```
daisy1754: I assume you want to save `current time + 10min`, not `10min`.

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -132,11 +134,13 @@ public function indexAction()
             // Check session cache for estimate.  If the shipping city or postcode, and the country code match,
             // then use the cached version.  Otherwise, we have to do another calculation
             ////////////////////////////////////////////////////////////////////////////////////////
-            $cached_address = unserialize(Mage::app()->getCache()->load("quote_location_".$quote->getId()));
+            $cacheKeyLocation = $quote->getId() . '_' . $address_data['country_id'];
+            $prefetchCacheKey = $this->generateCacheKey($quote, $address_data);
+            $cached_address = unserialize(Mage::app()->getCache()->load('quote_location_' . $cacheKeyLocation));
 
             if ($cached_address && ((strtoupper($cached_address["city"]) == strtoupper($address_data["city"])) || ($cached_address["postcode"] == $address_data["postcode"])) && ($cached_address["country_id"] == $address_data["country_id"])) {
```
daisy1754: need to check if current time vs cache expiration

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
YevhenShBolt: I am not sure that i understand you. I set the lifetime of the current cache 10 minutes.

As you previously wrote:
> can you also include timestamp so we won't use stale cache? (let's say it'll expire in 10min)

As i see (maybe i wrong), but lifetime set in seconds and it determine how long cache will be active after creation.
There is no needs to setup it with timestamp + 10minutes i guess.
By default Zend_Cache_Core class set cache lifetime to 3600 as i see.

daisy1754: please disregard this comment. I didn't understand how `Mage::app()->getCache()->save(` works

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
daisy1754: please ignore this comment too.

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -168,39 +172,119 @@ public function indexAction()
...
+            // process fails or is aborted, which will force the actual index action
+            // to get fresh data instead of reading from the session cache
+            ////////////////////////////////////////////////////////////////////////
+            $cacheKeyLocation = $quote->getId() . '_' . $address_data['country_id'];
+            $prefetchCacheKey = $this->generateCacheKey($quote, $address_data);
+            Mage::app()->getCache()->remove('quote_location_' . $cacheKeyLocation);
+            Mage::app()->getCache()->remove($prefetchCacheKey);
+            ////////////////////////////////////////////////////////////////////////
+
+            $quote->getShippingAddress()->addData($address_data);
```
daisy1754: Does this address John's concern? It seems we always override address in quote, which we should avoid - if quote already has address info, it shouldn't be overridden by IP-based inferred address
https://app.asana.com/0/436430054069245/633055371214186/f

YevhenShBolt: A bit above we have 
if ($shippingAddress && empty($countryId) && $countItems)

IF in quote exist $countryId we do not do a recalculation.

daisy1754: maybe I'm a bit paranoid but is there any situation where $countryId is not set but region etc. are set?

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -168,39 +172,119 @@ public function indexAction()
...
+                array('BOLT_PREFETCH_SHIPPING_AND_TAX'),
+                $lifeTime
+            );
+        } else {
+            $address_data = array();
+        }
+
+        $response = Mage::helper('core')->jsonEncode(array('address_data' => $address_data));
+        $this->getResponse()->setHeader('Content-type', 'application/json');
+        $this->getResponse()->setHeader('X-Bolt-Cache-Hit', 'HIT');
```
daisy1754: this should come to line 147. Idea is to change this based on whether we read it from cache or not (so if cache doesn't hit pleaes call `$this->getResponse()->setHeader('X-Bolt-Cache-Hit', 'MISS');`

daisy1754:
LGTM - I'll defer approval to @johnmccombs1

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
YevhenShBolt: added

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
johnmccombs1: @daisy1754 I did check a recent database to confirm that country id is set whenever there was an address. I'm not sure if this is always the case, but I think it's a good assumption. No reason we couldn't just add that though to be safe.

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -168,39 +172,119 @@ public function indexAction()
...
+    public function generateCacheKey($quote, $addressData)
+    {
+        /* @var $shippingAddress Mage_Sales_Model_Quote_Address */
+        $shippingAddress = $quote->getShippingAddress();
+        $itemIDs = $quote->getItemsCollection()->getAllIds();
+        $quoteTotalAmount = round($quote->getGrandTotal()*100);
+        $isCustomerGuest = $quote->getCustomerIsGuest();
+        $discountCode = str_replace(' ', '', $shippingAddress->getDiscountDescription());
+
+        $key = $quote->getId() . '_' . $isCustomerGuest . '_' . $discountCode . '_' .$quoteTotalAmount;
```
johnmccombs1: I'm a little worried about using $quoteTotalAmount because in edge cases the customer could swap out items that are the same price but a different weight, in which case we would need to recalculate. What if we instead run through all skus and quantities and create a checksum? That way we are ensuring that the actual items are all accounted for?

johnmccombs1: $quoteTotalAmount is also important still since the total could change based on other factors (customer group, shopping cart price rule that doesn't depend on coupon, etc). But I think my point was that we need to make sure that the items in the cart are the same.

johnmccombs1: Disregard my last comments. Just saw that you're adding them at the end :)

YevhenShBolt: @daisy1754, Should I add product weight and qty there?

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -132,11 +134,13 @@ public function indexAction()
             // Check session cache for estimate.  If the shipping city or postcode, and the country code match,
             // then use the cached version.  Otherwise, we have to do another calculation
             ////////////////////////////////////////////////////////////////////////////////////////
-            $cached_address = unserialize(Mage::app()->getCache()->load("quote_location_".$quote->getId()));
+            $cacheKeyLocation = $quote->getId() . '_' . $address_data['country_id'];
+            $prefetchCacheKey = $this->generateCacheKey($quote, $address_data);
```
johnmccombs1: I think we need to see if the $prefetchCacheKey actually exists in the cache. I just tested on my local and it doesn't, so the response returned is empty and no shipping rates are returned at all. I think we might be over complicating this. I would suggest that we create an array that contains all of the fields that we care about (e.g. quote id, street address, city, state, postal code, country, grand total, skus, etc), populate it in generateCacheKey(), and store that as a single checksum rather than keeping two cache keys. If it exists then we can know whether or not the quote is exactly the same or not, rather than just checking that the city, postcode, and country are the same. If so we use that checksum to return the response. If not, we create a new one.

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
daisy1754: I think weight is implied via sku. adding qty is probably a good idea.

YevhenShBolt:
@daisy1754 I refactored the whole class, please review.
As I see now it cached properly and Bolt checkout popup used that cached data.
Also, I measured the time executions for controller action methods with cache and without cache.

The prefetchEstimateAction() method:
without Cache: ~ 220 - 260 ms
with Cache: ~145 - 160ms

The indexAction() method:
without Cache: ~ 345 - 360 ms
with Cache: ~190 - 220ms

PR #34, opened 2018-04-25
Fixed bug with partial capture throwing exception for capture amount …

johnmccombs1:
…when it wasn't supposed to be sent.

### DIFF
--- a/app/code/community/Bolt/Boltpay/controllers/ApiController.php
+++ b/app/code/community/Bolt/Boltpay/controllers/ApiController.php
@@ -185,12 +185,6 @@ class Bolt_Boltpay_ApiController extends Mage_Core_Controller_Front_Action
             return $transaction->capture->amount->amount/100;
         }
 
-        Mage::helper('boltpay/bugsnag')->addMetaData(
-            array(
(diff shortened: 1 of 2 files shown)


app/code/community/Bolt/Boltpay/controllers/ApiController.php
```diff
@@ -185,12 +185,6 @@ protected function getCaptureAmount($transaction) {
             return $transaction->capture->amount->amount/100;
```
daisy1754: should we check existence of `$transaction->capture` and `$transaction->capture->amount`?

johnmccombs1: I think we only care about the amount, right? isset() will return true only if all variables in the chain are defined I believe. So if $transaction->capture is undefined, isset($transaction->capture->amount->amount) will return false and won't error out. Is that what you're asking?

daisy1754: aha - PHP is such a generous language. I was worried about potential nullpointerexception kind of error but it sounds like it won't happen. thanks for clarification.

PR #41, opened 2018-05-02
Fixed bug where percentage discounts on order were causing +/- $0.01 …

johnmccombs1:
…differences

### DIFF
--- a/app/code/community/Bolt/Boltpay/Helper/Api.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Api.php
@@ -320,6 +320,17 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
         return $rateDebuggingData;
     }
 
+    /**
+     * Determines whether the discount amount from Bolt is off by $0.01 compared to the Magento quote discount amount
+     *
+     * When $quote->collectTotals calls Mage_SalesRule_Model_Validator->process it uses a singleton to instantiate the
+     * validator so when it gets called each time, the  _roundingDeltas variable persists previous data and causes off
+     * by $0.01 rounding errors. Each call to collectTotals either sets it to the correct amount or to an amount that
+     * is off by $0.01. This function detects this problem.
+     *
+     * @param $transaction  Transaction data sent by Bolt
+     * @param Sales_Model_Service_Quote $quote     Quote derived from transaction data
+     */
     protected function isDiscountRoundingDeltaError($transaction, $quote) {
         $boltDiscountAmount = $this->getBoltDiscountAmount($transaction);
         $quoteDiscountAmount = $quote->getShippingAddress()->getDiscountAmount();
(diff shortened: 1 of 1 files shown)


app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -316,6 +320,54 @@ protected function getRatesDebuggingData($rates) {
...
+            $difference = round(abs($boltDiscountAmount - $quoteDiscountAmount), 2);
+            if($difference <= 0.01) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    protected function fixQuoteDiscountAmount($transaction, $quote) {
```
daisy1754: Can you add comment (and maybe link to site describing magento's bug) to explain why this is needed and how it fixes the issue?

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -316,6 +320,54 @@ protected function getRatesDebuggingData($rates) {
...
+        $quote->setTotalsCollectedFlag(false)->collectTotals()->save();
+
+        if($this->isDiscountRoundingDeltaError($transaction, $quote)) {
```
daisy1754: We should instead check `$boltDiscountAmount != $quoteDiscountAmount`

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -316,6 +320,54 @@ protected function getRatesDebuggingData($rates) {
...
+                    $boltDiscountAmount -= $discount->amount->amount / 100;
+                }
+            }
+        }
+
+        return $boltDiscountAmount;
+    }
+
+    protected function isOnePennyRoundingError($boltDiscountAmount, $quoteDiscountAmount) {
+        if($boltDiscountAmount != $quoteDiscountAmount) {
```
daisy1754: are these float? What if `$boltDiscountAmount` is 10.00 and `$quoteDiscountAmount` is 10.000001 ?

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -316,6 +320,54 @@ protected function getRatesDebuggingData($rates) {
...
+            $difference = round(abs($boltDiscountAmount - $quoteDiscountAmount), 2);
```
daisy1754: I prefer we don't assume sign of boltdiscountamount of quotediscountamount so `round(abs(A) - abs(B), 2)`

app/code/community/Bolt/Boltpay/Helper/Api.php
johnmccombs1: That's essentially what this is doing, but it's checking to see if they are only off by $0.01. If they are off by more than this, then there is a different issue. Calling collectTotals could pull in other artifacts that could cause this number to be off by more than $0.01 (admin adds a new discount amount, etc) which would be a different error and would be caught by validateSubmittedOrder

app/code/community/Bolt/Boltpay/Helper/Api.php
johnmccombs1: Both Bolt and Sales_Model_Service_Quote are rounded to two decimal places. In Magento collectTotals is always rounded just before returning the totals. If you think we should be more explicit we could round them to 2 decimal places here just to confirm. Thoughts?

app/code/community/Bolt/Boltpay/Helper/Api.php
johnmccombs1: I considered this. But what about the case where the Bolt discount is positive and the Magento quote is negative (we're currently allowing this with shipping rate discounts). So for example, let's say we have a Bolt discount amount of $8.01 and a Magento discount amount of -$8.00. If we do round(abs(A) - abs(B), 2) we will end up with $8.01 - $8.00 = $0.01 which will incorrectly flag this error. Instead it should be $8.01 - (-$8.00) = $16.01 which is accurate and won't flag this as a rounding error.

app/code/community/Bolt/Boltpay/Helper/Api.php
johnmccombs1: Pushed new change to address this.

PR #45, opened 2018-05-10
Cleanup bugsnag reporting for API error

chirag-taraviya:


app/code/community/Bolt/Boltpay/Helper/Bugsnag.php
```diff
@@ -163,4 +163,11 @@ protected function getBoltTraceId()
 
         return null;
     }
+
+    
+    public function setMetaData(array $metaData)
```
daisy1754: `set` sounds like it would override existing value. can you rename it to `addMetaData`?

chirag-taraviya: @daisy1754 addMetaData already used for breadcrumb , so i rename to addBoltMetaData

daisy1754:
I just realized we already had `addMetaData` method. can you use this instead?

chirag-taraviya:
addMetaData function will add meta to breadcrumbs_ , we needs two another tab so i think we can not use addMetaData

daisy1754:
@chirag-taraviya can you send test exception to bugsnag and post screenshots of how report will look like for 1) when youuse  addMetaData and when you use your new addBoltMetaData?

daisy1754:
@chirag-taraviya thank you! Can you create a separate pull request to rename `addMetaData` to `addBreadcrumb`? Then let's use name `addMetaData` here.

chirag-taraviya:
addMetaData is used many places so its bigger call to change everywhere , still needs to change addMetaData  to addBreadcrumb?  most of data log added in Breadcrumb. Please suggest

daisy1754:
To clarify my proposal is just renaming. Search and replace all `addMetaData` to `addBreadcrumb` - no behaviour change. Also please do it as separate pull request as you mentioned it'll be bigger change in terms of number of files changed.

chirag-taraviya:
ok will do it

daisy1754:
@chirag-taraviya As I mentioned above, please create a separate pull request for renaming addMetaData to addBreadcrumb.

You'll need to create a separate branch - ie.,
```
git checkout development
git checkout -b rename_addbreadcrumb
git cherry-pick 9dbe204a656478d607561ee5e77e4655abd3fc27
```


Also, for this pull request, please refer to my first comment "I think there are other code that makes http request. can you update other places too?"

daisy1754:
Please merge with dev and resolve conflict once it's ready for review. Thanks.

chirag-taraviya:
https://app.bugsnag.com/bolt/magento/errors/5ac74508b06e3d0019b34d18?filters[event.since][0]=30d&filters[error.status][0]=open

daisy1754:
something doesn't look right. can you check?
https://app.bugsnag.com/bolt/magento/errors/5ac74508b06e3d0019b34d18?filters%5Bevent.since%5D%5B%5D=30d&filters%5Berror.status%5D%5B%5D=open&event_id=5afd7c50508af1001b927f21

screen shot 2018-05-17 at 8 26 20 am

chirag-taraviya:
we can add last response instead of all bolt response , here is out put if we need only last execute bolt response its look like , Please let me know so i can modified my code 
image

daisy1754:
yes last response is what we need.

app/code/community/Bolt/Boltpay/Helper/Bugsnag.php
```diff
@@ -164,4 +164,15 @@ protected function getBoltTraceId()
         return null;
     }
 
+   /* add metaData for to create new tab
+    *
+    * @param array $metaData
+    * @param Boolean $merge
+    *
+    */
+    public function addMetaData(array $metaData, $merge = false)
```
daisy1754: @chirag-taraviya what does `$merge` param do? Can we remove this and hard-code true at line 175?

PR #46, opened 2018-05-11
order reference used in place of display id to retrieve quotes

mccottry:
https://app.asana.com/0/436430054069245/614370607072692

### DIFF
--- a/app/code/community/Bolt/Boltpay/Helper/Api.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Api.php
@@ -160,7 +160,7 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
      * Processes Magento order creation. Called from both frontend and API.
      *
      * @param string    $reference           Bolt transaction reference
-     * @param int       $session_quote_id    Quote id, used if trigger from shopping session context,
+     * @param int       $session_quote_id    Quote id, used if triggered from shopping session context,
      *                                       This will be null if called from within an API call context
      *
      * @return Mage_Sales_Model_Order   The order saved to Magento
@@ -189,12 +189,10 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
             throw new Exception("The Bolt order reference does not match the current cart ID.");
         }
 
-        $display_id = $transaction->order->cart->display_id;
+        $order_id = $transaction->order->cart->display_id;
 
-        $quote = Mage::getModel('sales/quote')
-            ->getCollection()
-            ->addFieldToFilter('reserved_order_id', $display_id)
-            ->getFirstItem();
+        /* @var Mage_Sales_Model_Quote $quote */
+        $quote = Mage::getModel('sales/quote')->loadByIdWithoutStore($quote_id);
 
         // adding guest user email to order
         if (!$quote->getCustomerEmail()) {
@@ -259,14 +257,19 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
 
         $quote->setTotalsCollectedFlag(false)->collectTotals()->save();
 
-        $existingOrder = Mage::getModel('sales/order')->loadByIncrementId($display_id);
+        /**
+         * TODO: Investigate if this is a problem.  Logically, it makes no sense that the order
+         *       exists at this point if we assume that the id is permanently reserved.  If the
+         *       id becomes available for use by another quote, this will be a problem.
+         */
+        $existingOrder = Mage::getModel('sales/order')->loadByIncrementId($order_id);
         if (sizeof($existingOrder->getData()) > 0) {
             Mage::app()->getResponse()->setHttpResponseCode(200);
             Mage::app()->getResponse()->setBody(
                 json_encode(
                     array(
                     'status' => 'success',
-                    'message' => "Order increment $display_id already exists."
+                    'message' => "Order increment $order_id already exists."
                     )
                 )
             );
--- a/app/code/community/Bolt/Boltpay/Helper/Api.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Api.php
@@ -189,7 +189,7 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
             throw new Exception("The Bolt order reference does not match the current cart ID.");
         }
 
-        $order_id = $transaction->order->cart->display_id;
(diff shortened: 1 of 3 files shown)


daisy1754:
I'd defer this to @danac-gs or @johnmccombs1

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -259,14 +257,19 @@ public function createOrder($reference, $session_quote_id = null)
...
         $quote->setTotalsCollectedFlag(false)->collectTotals()->save();
 
-        $existingOrder = Mage::getModel('sales/order')->loadByIncrementId($display_id);
+        /**
+         * TODO: Investigate if this is a problem.  Logically, it makes no sense that the order
+         *       exists at this point if we assume that the id is permanently reserved.  If the
+         *       id becomes available for use by another quote, this will be a problem.
+         */
+        $existingOrder = Mage::getModel('sales/order')->loadByIncrementId($order_id);
         if (sizeof($existingOrder->getData()) > 0) {
```
YevhenShBolt: I guess we should do something like
if (!empty($order->getData())) {

AS you do it in app/code/community/Bolt/Boltpay/controllers/ApiController.php::85

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -189,12 +189,10 @@ public function createOrder($reference, $session_quote_id = null)
...
 
-        $display_id = $transaction->order->cart->display_id;
+        $order_id = $transaction->order->cart->display_id;
 
-        $quote = Mage::getModel('sales/quote')
-            ->getCollection()
-            ->addFieldToFilter('reserved_order_id', $display_id)
-            ->getFirstItem();
+        /* @var Mage_Sales_Model_Quote $quote */
+        $quote = Mage::getModel('sales/quote')->loadByIdWithoutStore($quote_id);
```
danac-gs: If the new code works correctly with the hooks, I am OK with the change.

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -189,12 +189,10 @@ public function createOrder($reference, $session_quote_id = null)
             throw new Exception("The Bolt order reference does not match the current cart ID.");
         }
 
-        $display_id = $transaction->order->cart->display_id;
+        $order_id = $transaction->order->cart->display_id;
```
johnmccombs1: I wonder if we should call it what it is on the order and use increment_id, or what it is on the quote and call it reserved_order_id. When I think of order_id, I think of entity_id on sales_flat_order which isn't what order_id is here. I'm good either way here, but my preference would be to call it either reserved_order_id or increment_id.

mccottry: Good point.  How I look at the naming of the variable is and the line of code is to give as much context to the next developer as possible.  So, in this line of code, if there is a new developer, he or she will see that the `order_id` is equivalent to the Bolt concept of `display_id`.  However, I fail to convey to him that in Magento, the displayed id is actually the `increment_id` in Magento speak.  I believe the Magento poorly named this and is inherently confusing, so I propose calling the variable `$order_increment_id` so that all of the above ideas are conveyed.

daisy1754: +1 
I think I like `reservedOrderId` or `incrementId`.

app/code/community/Bolt/Boltpay/Helper/Api.php
daisy1754: Plugin should assume hook could be called multiple times. If HTTP request times out, Bolt server would assume it was failure and try again later. 
My assumption is this if block exists to handle 1) rare hook retries as described above and 2) race condition that leads to calling this method from both frontend and via hook.

daisy1754: Re: `sizeof vs empty`, I slightly prefer `empty` for readability but IMO not a big deal.

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -259,14 +257,19 @@ public function createOrder($reference, $session_quote_id = null)
 
         $quote->setTotalsCollectedFlag(false)->collectTotals()->save();
 
-        $existingOrder = Mage::getModel('sales/order')->loadByIncrementId($display_id);
+        /**
+         * TODO: Investigate if this is a problem.  Logically, it makes no sense that the order
```
daisy1754: What is the next step here? If you have clear idea on how to investigate, maybe create asana task and put link here? If not, I suggest we don't add comment.

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -78,18 +78,15 @@ public function indexAction()
                 'telephone' => $shipping_address->phone
             );
 
-            $display_id = $request_data->cart->display_id;
-
+            $quote_id = $request_data->cart->order_reference;
```
daisy1754: nit: going forward let's use camelCase within codebase for consistency (and to follow magento's convention).

app/code/community/Bolt/Boltpay/Helper/Api.php
YevhenShBolt: +1 @daisy1754

mccottry: ok.  `reservedOrderId`, camel_cased seems like a winner.  Look close, there is a funny joke in there. :dancing_women:

app/code/community/Bolt/Boltpay/Helper/Api.php
mccottry: Ok.  The suggestion makes sense and I will factor in the public nature of the code and repository to convey these ideas.  A mix of GitHub, Asana, and Slack could easily be used instead of inline code comments for this sort of dialogue.

app/code/community/Bolt/Boltpay/Helper/Api.php
mccottry: > Re: sizeof vs empty, I slightly prefer empty for readability but IMO not a big deal.

Me too.  But, let's take it to the next level and use the Magento method which is meant for this exact purpose:
`if (!$order->isEmpty())`

This will also be in the separate branch and pull request listed above.

mccottry:
>Feel free to merge once comments are addressed

Comments were addressed.  Merging changes.

PR #47, opened 2018-05-11
Bug collect totals shipping rate

johnmccombs1:


app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -951,15 +951,24 @@ public function getShippingAndTaxEstimate( $quote )
     public function applyShippingRate($quote, $shippingRateCode) {
         $shippingAddress = $quote->getShippingAddress();
 
-        if(!empty($shippingAddress)) {
+        if (!empty($shippingAddress)) {
             // Unsetting address id is required to force collectTotals to recalculate discounts
```
daisy1754: Is this comment still valid?

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -951,15 +951,24 @@ public function getShippingAndTaxEstimate( $quote )
...
+            $shippingAddress->isObjectNew(true);
             $shippingAddressId = $shippingAddress->getData('address_id');
-            $shippingAddress->unsetData('address_id');
 
             $shippingAddress->setShippingMethod($shippingRateCode);
+
+            $quoteItems = $quote->getAllItems();
+            foreach ($quoteItems as $item) {
```
daisy1754: mind adding comment why you need to re-set discount?

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -951,15 +951,24 @@ public function getShippingAndTaxEstimate( $quote )
...
+                $item->setData('discount_amount', $item->getOrigData('discount_amount'));
+                $item->setData('base_discount_amount', $item->getOrigData('base_discount_amount'));
+            }
+
             $quote->setTotalsCollectedFlag(false)->collectTotals();
 
-            $shippingAddress->setData('address_id', $shippingAddressId);
+            if(!empty($shippingAddressId) && $shippingAddressId != $shippingAddress->getData('address_id')) {
```
daisy1754: when `$shippingAddressId != $shippingAddress->getData('address_id')` becomes true?

johnmccombs1: This is for a hypothetical case where $address->isObjectNew() may cause some future code to think that address doesn't exist and attempts to create a new one. I believe this would happen if the quote was saved as this flag is used to determine on "before save" whether or not to create a new address id. It's possibly that a future total collector could call save on the quote and cause an issue here. We could modify this to check this case and create a bugsnag entry if it ever happens.

PR #49, opened 2018-05-15
Removes reliance on shopping cart and ensures Bolt/Magento sync

mccottry:
Pending 
- addition of price verification that replaces complete_authorize
- production readiness full testing
- resolution of the git reported conflicts with `develop`


app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -174,7 +170,54 @@ public function getCartDataJs($multipage = true)
...
+                // Only shipping, tax and discounts can change, and only if the shipping, tax
+                // and discount calculations change on the Magento server
+                ////////////////////////////////////////////////////////////////////////////////
+
+                /*********************************************************/
+                /* Clean up resources that may have previously been saved
+                /* @var Mage_Sales_Model_Quote[] $expired_quotes */
+                $expired_quotes = Mage::getModel('sales/quote')
+                    ->getCollection()
+                    ->addFieldToFilter('parent_quote_id', $quote->getId());
```
johnmccombs1: I don't see where this column has been added to the sales_flat_quote_table. Are you missing an installer script or am I just not seeing all of the files?

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -174,7 +170,54 @@ public function getCartDataJs($multipage = true)
...
+                        ->setBillingAddress($quote->getBillingAddress())
+                        ->setShippingAddress($quote->getShippingAddress())
+                        ->getShippingAddress()
+                            ->setShippingMethod($quote->getShippingAddress()->getShippingMethod())
+                            ->save();
+                }
+
+                $immutableQuote
+                    ->setCustomer($quote->getCustomer())
+                    ->merge($quote)
```
johnmccombs1: I'm not familiar with the merge function. Does this take anything from the quote that is passed in and assign anything that isn't already set on the immutableQuote? Just curious.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -174,7 +170,54 @@ public function getCartDataJs($multipage = true)
...
+                    // Generate new increment order id and associate it with current quote, if not already assigned
+                    ->reserveOrderId()
```
johnmccombs1: I don't know if this is an issue or not, but won't this increment the reserveOrderId on every cart or checkout page refresh. I'm not sure if that's an issue, but I know that before I realized that a reserved order id didn't always have a one-to-one mapping with an order entity_id, I would often look at the increment id to gauge how many orders were in the system. I know this is stupid, but I think it would cause some people to think there is something wrong if we drastically increase the gap between increment ids. However, this may not be an issue. Just pointing it out.

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -299,13 +285,34 @@ public function createOrder($reference, $session_quote_id = null)
...
+                ->setLastQuoteId($quote->getId())
+                ->setLastSuccessQuoteId($quote->getId());
+
+            // add order information to the session
+            $checkout_session->setLastOrderId($order->getId())
+                ->setRedirectUrl('')
+                ->setLastRealOrderId($order->getIncrementId());
+        }
+
+        $quote->delete();
```
johnmccombs1: Just want to make sure that i'm following. We are deleting the immutable quote here, but is that the quote that is associated with the order? If so, shouldn't this be the one that we set to inactive and delete the parent quote? Otherwise the order will have reference to a quote id that doesn't exist.

app/code/community/Bolt/Boltpay/controllers/ApiController.php
```diff
@@ -80,9 +80,29 @@ public function hookAction()
...
+             * failed to deliver on user experience and we should go ahead and treat
+             * it as an orphaned transaction and create the order.
+             ***************************************************************************/
+            ignore_user_abort(true);
+            set_time_limit( 60 );
+            $remaining_attempts = 6;
+            do {
+                /* @var Mage_Sales_Model_Order $order */
+                $order = Mage::getModel('sales/order')->loadByIncrementId($orderId);
+                sleep(5 );
```
johnmccombs1: Would it be simpler to just increase the time that Bolt waits before the initial attempt to call the hook api? I seem to recall that Kazuki can change this on the Bolt side. Just worried that if we do it here we disable their ability to do it on the Bolt side.

daisy1754: Yes let's increase delay for calling hook from bolt. It's currently 10sec. What do you suggest? 30sec? 60sec?

Sleeping is not a good idea because 
1) process keeps holding resources (eg., TCP connections) and 2) caller (in this case Bolt server) may timeout.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -174,7 +170,54 @@ public function getCartDataJs($multipage = true)
...
+                    ->setStoreId($quote->getStoreId())
+                    ->setParentQuoteId($quote->getId());
+
+                $orderCreationResponse = $this->createBoltOrder($immutableQuote, $multipage);
+                $immutableQuote->setBoltOrderToken($orderCreationResponse->token)
```
danac-gs: How the token remains persistent within the quote? Is this enough or the additional field definition is required via installer?

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: @johnmccombs1 Correct.  The install scripts were not in the commit.  They are included now.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry:
```
    /**
     * Merge quotes
     *
     * @param   Mage_Sales_Model_Quote $quote
     * @return  Mage_Sales_Model_Quote
     */
    public function merge(Mage_Sales_Model_Quote $quote)
    {
        Mage::dispatchEvent(
            $this->_eventPrefix . '_merge_before',
            array(
                 $this->_eventObject=>$this,
                 'source'=>$quote
            )
        );

        foreach ($quote->getAllVisibleItems() as $item) {
            $found = false;
            foreach ($this->getAllItems() as $quoteItem) {
                if ($quoteItem->compare($item)) {
                    $quoteItem->setQty($quoteItem->getQty() + $item->getQty());
                    $found = true;
                    break;
                }
            }

            if (!$found) {
                $newItem = clone $item;
                $this->addItem($newItem);
                if ($item->getHasChildren()) {
                    foreach ($item->getChildren() as $child) {
                        $newChild = clone $child;
                        $newChild->setParentItem($newItem);
                        $this->addItem($newChild);
                    }
                }
            }
        }

        /**
         * Init shipping and billing address if quote is new
         */
        if (!$this->getId()) {
            $this->getShippingAddress();
            $this->getBillingAddress();
        }

        if ($quote->getCouponCode()) {
            $this->setCouponCode($quote->getCouponCode());
        }

        Mage::dispatchEvent(
            $this->_eventPrefix . '_merge_after',
            array(
                 $this->_eventObject=>$this,
                 'source'=>$quote
            )
        );

        return $this;
    }
```

mccottry:
```
    /**
     * Merge quotes
     *
     * @param   Mage_Sales_Model_Quote $quote
     * @return  Mage_Sales_Model_Quote
     */
    public function merge(Mage_Sales_Model_Quote $quote)
    {
        Mage::dispatchEvent(
            $this->_eventPrefix . '_merge_before',
            array(
                 $this->_eventObject=>$this,
                 'source'=>$quote
            )
        );

        foreach ($quote->getAllVisibleItems() as $item) {
            $found = false;
            foreach ($this->getAllItems() as $quoteItem) {
                if ($quoteItem->compare($item)) {
                    $quoteItem->setQty($quoteItem->getQty() + $item->getQty());
                    $found = true;
                    break;
                }
            }

            if (!$found) {
                $newItem = clone $item;
                $this->addItem($newItem);
                if ($item->getHasChildren()) {
                    foreach ($item->getChildren() as $child) {
                        $newChild = clone $child;
                        $newChild->setParentItem($newItem);
                        $this->addItem($newChild);
                    }
                }
            }
        }

        /**
         * Init shipping and billing address if quote is new
         */
        if (!$this->getId()) {
            $this->getShippingAddress();
            $this->getBillingAddress();
        }

        if ($quote->getCouponCode()) {
            $this->setCouponCode($quote->getCouponCode());
        }

        Mage::dispatchEvent(
            $this->_eventPrefix . '_merge_after',
            array(
                 $this->_eventObject=>$this,
                 'source'=>$quote
            )
        );

        return $this;
    }
```

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: @johnmccombs1 I think it is helpful for me to review code from the diff view.  It helps me answer questions like this visually and you might find it to be useful so I will provide a link.

https://github.com/BoltApp/bolt-magento1/pull/49/files

Anyway, likely, you are already doing this, but just in case, see above.

Now, to answer your questions, if you look at the diff in the files, you will notice that `reservedOrderId()` is called exactly the same number of times at the same point in the order flow as it was before.  Therefore, there is no change to how order Ids are being reserved and thus, there is no difference in what it will appear like in the underlying database.

Then we must ask the question, "is the preceding implementation wrong?"  The answer is that the proceeding implementation is correct, as is its continued use.  This is because calling this function will only generate a new reserved order id if one is not already assigned to the quote and the id is not in use by a current order.  You can look this up in the Magento source code.

However, perhaps your observation would be correct if the quote was not saved after the call to reserve order id.  It is saved on line 219.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: @danac-gs I did not include the installer script in the original push.  It has been added.  Good call.  Thanks.

app/code/community/Bolt/Boltpay/Helper/Api.php
mccottry: If the code reaches this point, then a Magento order has been created.  There is no longer a need for any quote, but, if there was, for some sort of auditing purposes, then the original quote theoretically should be the same.

For security purposes, once an immutable quote is used, it must no longer be available for any processing.  If it existed after its use, then if someone managed to get the transaction id or order token could attempt to reuse those to fool the system by maliciously reusing previous transactions.  Furthermore, on a more minor level, it is kept and set to inactive as to not use unnecessary space in the sales_flat_quote table.

app/code/community/Bolt/Boltpay/controllers/ApiController.php
mccottry: Most importantly, I will first say that a Bolt side implementation for increased delay would definitely be ideal. Historically, It started at 0 seconds, moved to 5 seconds, and was quickly switched to 10 seconds.  Still, none of these are enough, and I recall Kazuki mentioning to me that he has seen times well above that.  In the code above, We give it 30 seconds and I believe 30 seconds is more than enough.  If it is taking that long, there are other merchant-side problems to consider.

Now, if in reality, it is difficult to get this change switched on Bolt-side for whatever reason, then it appears useful to explain some things about the code above.

@johnmccombs1 In no way can this disable the ability on Bolt-side to create a delay.  In the average case scenario, when Bolt arrives, there is no delay.  I would guess in the high 90 percents and maybe much more.  In fact, if caching of shipping and tax is done with foresight, and used here, then I think that this would reach near 100%.  The next most common scenario is Bolt time plus 5 seconds.

@daisy1754  Bolt-side solution is certainly desirable across all plugins, and if that code is set to 30 seconds, then this fallback will likely never be triggered.  But, if for whatever reason it is triggered:

1.)  PHP `sleep` does not take up any CPU resources.  The only resources would be the memory footprint used by Magento and we would hope that this is negligable in a mature software framework, and perhaps, expect it.

2.) The code assumes that Bolt will drop the connection before it finishes.
` ignore_user_abort(true);` handles this.  It is coupled with a maximum allowed execution time of 60 seconds by `set_time_limit( 60 );`  TCP connections end as soon as Bolt ends them, which I also believe max time is 10 seconds.

Finally, in practice, this code should be triggered very rarely, but when used, it makes the current Bolt-side 10 second delay practical (or even better a 5 second delay) and can reduce the auth confirmation time from the proposed 30 delay + plus magento processing time to something closer to de facto real time.  
Most importantly, I will first say that a Bolt side implementation for increased delay would definitely be ideal. Historically, It started at 0 seconds, moved to 5 seconds, and was quickly switched to 10 seconds.  Still, none of these are enough, and I recall Kazuki mentioning to me that he has seen times well above that.  In the code above, We give it 30 seconds and I believe 30 seconds is more than enough.  If it is taking that long, there are other merchant-side problems to consider.

Now, if in reality, it is difficult to get this change switched on Bolt-side for whatever reason, then it appears useful to explain some things about the code above.

@johnmccombs1 In no way can this disable the ability on Bolt-side to create a delay.  In the average case scenario, when Bolt arrives, there is no delay.  I would guess in the high 90 percents and maybe much more.  In fact, if caching of shipping and tax is done with foresight, and used here, then I think that this would reach near 100%.  The next most common scenario is Bolt time plus 5 seconds.

@daisy1754  Bolt-side solution is certainly desirable across all plugins, and if that code is set to 30 seconds, then this fallback will likely never be triggered.  But, if for whatever reason it is triggered:

1.)  PHP `sleep` does not take up any CPU resources.  The only resources would be the memory footprint used by Magento and we would hope that this is negligable in a mature software framework, and perhaps, expect it.

2.) The code assumes that Bolt will drop the connection before it finishes.
` ignore_user_abort(true);` handles this.  It is coupled with a maximum allowed execution time of 60 seconds by `set_time_limit( 60 );`  TCP connections end as soon as Bolt ends them, which I also believe max time is 10 seconds.

Finally, in practice, this code should be triggered very rarely, but when used, it makes the current Bolt-side 10 second delay practical (or even better a 5 second delay) and can reduce the auth confirmation time from the proposed 30 delay + plus magento processing time to something closer to de facto real time.

daisy1754: I meant HTTP connection resource. Even if machine is idle, it'll keep HTTP connection for 5+ seconds which is not ideal.
Also IMO this makes code more complex to deal with theoretical scenario.


Also I'm not sure what we're optimizing here for. If for some reason hook creates order first but frontend makes AJAX calls magento for order creation later, shouldn't AJAX call returns success response and frontend should show success message?

mccottry: @daisy1754 https://github.com/BoltApp/bolt-magento1/pull/54

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -302,13 +291,34 @@ public function createOrder($reference, $session_quote_id = null)
...
+                ->setLastRealOrderId($order->getIncrementId());
+        }
+
+        // Close out session by deleting the parent quote and deactivating the immutable quote so that it can no
+        // longer be used.
+        /* @var Mage_Sales_Model_Quote $parentQuote */
+        $parentQuote = Mage::getModel('sales/quote')->loadByIdWithoutStore($quote->getParentQuoteId());
+        $parentQuote->delete();
+        $quote->setIsActive(false)
+            ->setBoltOrderToken(null)
```
daisy1754: why we need to nullify orderToken? Is it useful for debugging if we keep it as is?

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -302,13 +291,34 @@ public function createOrder($reference, $session_quote_id = null)
...
+            $checkout_session->setLastOrderId($order->getId())
+                ->setRedirectUrl('')
+                ->setLastRealOrderId($order->getIncrementId());
+        }
+
+        // Close out session by deleting the parent quote and deactivating the immutable quote so that it can no
+        // longer be used.
+        /* @var Mage_Sales_Model_Quote $parentQuote */
+        $parentQuote = Mage::getModel('sales/quote')->loadByIdWithoutStore($quote->getParentQuoteId());
+        $parentQuote->delete();
```
daisy1754: why not just `setIsActive(false)`?

daisy1754: If we create a new quote at `getCartDataJs` and does not delete it unless order is created in magento, deleting here doesn't seem to help much in terms of disk usage, as most of quote are abandonded.

app/code/community/Bolt/Boltpay/Model/Observer.php
```diff
@@ -78,54 +78,43 @@ private function getBoltApiHelper()
...
-
-            try {
-                if (!Mage::getStoreConfig('payment/boltpay/disable_complete_authorize'))  {
-                    $this->sendCompleteAuthorizeRequest($complete_authorize_request);
-                }
-            } catch (Exception $e) {
+
+            if ( (int)($order->getGrandTotal()*100) !== $transaction->amount->amount)  {
+
                 $message = "THERE IS A MISMATCH IN THE ORDER PAID AND ORDER RECORDED.<br>PLEASE COMPARE THE ORDER DETAILS WITH THAT RECORD IN YOUR BOLT MERCHANT ACCOUNT AT: ";
```
daisy1754: Let's include `$order->getGrandTotal()*100` and `$transaction->amount->amount` in message.

app/code/community/Bolt/Boltpay/Model/Observer.php
```diff
@@ -78,54 +78,43 @@ private function getBoltApiHelper()
...
                 $message .= Mage::getStoreConfig('payment/boltpay/test') ? "https://merchant-sandbox.bolt.com" : "https://merchant.bolt.com";
                 $message .= "/transaction/$reference";
+
                 $order->setState(Mage_Sales_Model_Order::STATE_HOLDED, true, $message)
                     ->save();
 
                 $metaData = array(
```
daisy1754: Let's include `$order->getGrandTotal()*100` and `$transaction->amount->amount` here too.

app/code/community/Bolt/Boltpay/controllers/OrderController.php
```diff
@@ -59,22 +59,22 @@ public function saveAction()
...
-
-            // add order information to the session
-            $checkout_session->setLastOrderId($order->getId())
-                ->setRedirectUrl('')
-                ->setLastRealOrderId($order->getIncrementId());
+            if ($order->isEmpty()) {
+                $session_quote = $checkout_session->getQuote();
+                $boltHelper->createOrder($reference, $session_quote->getId());
+            }
+            /////////////////////////////////////////////////////////
```
daisy1754: do we need this line?

app/code/community/Bolt/Boltpay/sql/bolt_boltpay_setup/upgrade-0.0.17-1.1.0.php
```diff
@@ -0,0 +1,62 @@
...
+ * This installer adds the Bolt order token and parent quote id for bolt order tracking.
+ */
+/* @var Mage_Sales_Model_Resource_Setup $installer */
+$installer = Mage::getResourceModel('sales/setup', 'sales_setup');
+$installer->startSetup();
+
+Mage::log('Installing Bolt 1.1.0 updates', null, 'bolt_install.log');
+
+$installer->addAttribute(
+    "quote", "bolt_order_token", array(
```
daisy1754: unused?

app/code/community/Bolt/Boltpay/Helper/Api.php
mccottry: I was hoping to use the order token as a nonce.  However, it appears that the order token returned by calling the Bolt Create Order API is different than the one returned by calling the Bolt Fetch API.  If this is true, it will be removed in entirety.

app/code/community/Bolt/Boltpay/controllers/OrderController.php
mccottry: Do you mean the line 77?  The other three lines, yes.

app/code/community/Bolt/Boltpay/sql/bolt_boltpay_setup/upgrade-0.0.17-1.1.0.php
mccottry: Ultimately, if the token returned from the Bolt API create order endpoint is different from the token sent in a Bolt API fetch call, there will be no real need for this.  If they are the same, then it will be used as a nonce and is a better way to look up quotes as the sequential quote id/order reference is many times easier to guess.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -168,20 +163,113 @@ public function getCartDataJs($multipage = true)
...
+                    // and the chosen shipping method at order save time.
+                    $immutableQuote
+                        ->setBillingAddress($sessionQuote->getBillingAddress())
+                        ->setShippingAddress($sessionQuote->getShippingAddress())
+                        ->getShippingAddress()
+                        ->setShippingMethod($sessionQuote->getShippingAddress()->getShippingMethod())
+                        ->save();
+                }
+
+                $already_set_by_merge = array(
```
daisy1754: why we need this?

What is the problem of doing like:

```
foreach($sessionQuote->getData() as $key => $value ) {
  $immutableQuote->setData($key, $value);
}
```

app/code/community/Bolt/Boltpay/controllers/OrderController.php
```diff
@@ -88,18 +125,13 @@ public function saveAction()
      */
     public function lockcartAction() 
     {
-        ////////////////////////////////////////////////////////////////////////
-        // To be uncommented once the lock cart endpoint has been implemented
```
daisy1754: ?

app/code/community/Bolt/Boltpay/controllers/OrderController.php
```diff
@@ -72,13 +72,50 @@ public function saveAction()
...
+                        )
+                )
+            ;
 
         } catch (Exception $e) {
             Mage::helper('boltpay/bugsnag')->notifyException($e);
-            throw $e;
+
+            /////////////////////////////////////////////////////////
+            /// Since the order failed, we must cancel the Bolt Order
```
daisy1754: Please please don't do this. In case of inventory missing, invicta team needs to check items and may do partial capture.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: When running through the array completely, a critical element is being overwritten that causes the order creation to fail as the cart total passed to Bolt create order is set to zero.  All of these data points are set by merge in a working condition and do not need to be reset.

Each item can be checked one by one to see which one or which ones are causing the problem, but, they could all just as easily be ignored and not reset.  I chose the latter option.

The reason for going through the array and setting the values not listed is because other plugins may have added extra fields to the quote table like us, and we want to preserve this.

mccottry: So, it is not the prettiest looking piece of code, but, I can't really justify going through every element in the array that is already set when I know that simply not overwriting them will not produce the error that occurs when they are overwritten.  It may be that overwriting certain values trigger certain events, or other things that causes problems -- and the problems are clearly introduced by the unchecked
`foreach($sessionQuote->getData() as $key => $value ) {
  $immutableQuote->setData($key, $value);
}`

daisy1754: how about moving this and foreach below into subfunction named `copyAdditionalData` and add comment ther ethat just coping all data via foreach can cause weird result?

PR #53, opened 2018-05-21
Bug fix collecttotals percentage discount bug

johnmccombs1:
This is to fix the discount rounding issues with collectTotals. The primary goal is to make a public function for resetting Mage_SalesRule_Model_Validator->_roundingDeltas. Everything else is just to make this simpler (added a helper for this) and remove the workarounds for not having this in place before.


app/code/community/Bolt/Boltpay/Helper/Data.php
```diff
@@ -89,4 +89,25 @@ public function canUseForCountry($country)
...
+    /**
+     * Resets rounding deltas before calling collect totals which fixes bug in collectTotals that causes rounding errors
+     * when a percentage discount is applied to a quote
+     *
+     * @param Mage_Sales_Model_Quote $quote
+     * $param boolean $clearTotalsCollectedFlag Determines whether to set setTotalsCollectedFlag to false
+     *
+     * @return Mage_Sales_Model_Quote
+     */
+    public function collectTotals($quote, $clearTotalsCollectedFlag = false) {
```
daisy1754: @johnmccombs1 can you write unit test for this? (please push back if it'd be too complicated)

johnmccombs1:
@mccottry I did apply this change to every place that we call collectTotals. I am hesitant to override Mage_Sales_Model_Quote with this change as it may have unintended consequences since collectTotals is one of the most invasive functions in Magento. I like that we are controlling where it's used instead of blanket modifying it. You are right that there is still the potential for issues if a third party overrides how orders are saved and calls collectTotals in the before save event which could add back in the rounding error. However, I think unless this becomes a problem we should limit this change to our own code to reduce the risk of this change. Worst case we need to override the core call to collectTotals and the modification will be minor on our part. For now I'm going to wait.

app/code/community/Bolt/Boltpay/Helper/Data.php
johnmccombs1: I'll be creating a regression test for this in the future, but in order to expedite this I'm going to go ahead and migrate this.

PR #57, opened 2018-05-24
Update order note after order creation

dumega:
Currently when order is created via plugin, there are two ways, ajax and api, and we can not distinguish from each other, so we update the order note for this. and when issue happened (such as duplicated orders), we can get more details.


daisy1754:
@dumega FYI I addressed my own comment and merging now.

PR #62, opened 2018-05-29
Add bolt button to the default MiniCart popup

YevhenShBolt:
Added config values:
- Add Bolt Button Everywhere

Fixed the layout boltpay.xml - because "ifconfig" attribute seems does not work properly.


app/code/community/Bolt/Boltpay/Helper/Data.php
```diff
@@ -32,6 +32,15 @@
  */
 class Bolt_Boltpay_Helper_Data extends Mage_Core_Helper_Abstract
 {
+    /**
+     * @var string The Bolt sandbox url for the javascript
+     */
+    const JS_URL_TEST = 'https://connect-sandbox.bolt.com';
```
daisy1754: I'm pretty sure this is already defined somewhere.

app/code/community/Bolt/Boltpay/Helper/Data.php
```diff
@@ -89,4 +98,75 @@ public function canUseForCountry($country)
 
         return true;
     }
+
+    /**
+     * Get config value
+     *
+     * @return bool
+     */
+    public function isNeedAddButtonToMiniCart()
```
daisy1754: Do we need this config? Can we always replace minicart if it presents? We can add config later if there is merchant who use minicart but don't want Bolt there.

app/code/community/Bolt/Boltpay/Model/Checkout/Cart.php
```diff
@@ -0,0 +1,342 @@
...
+            Mage::helper('boltpay/bugsnag')->notifyException($e);
+        }
+    }
+
+    /**
+     * @param $cartData
+     * @param $hintData
+     * @return string
+     */
+    public function generateBoltCheckoutJavascript($cartData, $hintData)
```
daisy1754: can we reuse existing getCartDataJs method?

app/code/community/Bolt/Boltpay/Helper/Data.php
YevhenShBolt: Yeah it was already defined in Checkout_Boltpay block. I plan to refactor Block in the separate PR.

app/code/community/Bolt/Boltpay/Helper/Data.php
YevhenShBolt: I think we should add it because every merchant has own classes in the MiniCart buttons.

app/design/frontend/base/default/layout/boltpay.xml
```diff
@@ -29,22 +29,26 @@
...
+      <block type="core/template" name="bolt.extra.minicart-styles" as="extra-minicart-styles" template="boltpay/minicart/styles.phtml" />
     </reference>
   </default>
 
   <checkout_cart_index>
     <reference name="head">
       <block type="boltpay/checkout_boltpay" name="prefetch_shipping_quote" as="prefetch_shipping_quote" template="boltpay/prefetch_shipping_quote.phtml" />
       <block type="boltpay/checkout_boltpay" name="connect_js_script" as="connect_js" template="boltpay/connectjs.phtml" />
-      <block type="boltpay/checkout_boltpay" name="replace_script" as="replace_script" template="boltpay/replace.phtml" />
+      <!--<block type="boltpay/checkout_boltpay" name="replace_script" as="replace_script" template="boltpay/replace.phtml" />-->
```
daisy1754: no comment out please. let's remove if we don't need it.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -557,6 +557,15 @@ function getConfigSelectors()
         return json_encode(array_filter(explode(',', Mage::getStoreConfig('payment/boltpay/selectors'))));
     }
 
+    /**
+     * Returns the Replace Button Selectors configuration.
+     * @return string
+     */
+    public function getMiniCartSelector()
+    {
+        return json_encode(array_filter(explode(',', Mage::getStoreConfig('payment/boltpay/minicart_selector'))));
```
daisy1754: I don't see minicart_selector is defined. Is this used?

app/design/frontend/base/default/layout/boltpay.xml
```diff
@@ -29,22 +29,26 @@
 
   <default>
     <reference name="head">
+      <action method="addItem" ifconfig="payment/boltpay/add_button_to_minicart"><type>skin_css</type><name>css/boltpay/minicart.css</name><params/></action>
       <block type="boltpay/checkout_boltpay" name="track_script" as="track_script" template="boltpay/track.phtml" />
+      <block type="boltpay/checkout_boltpay" name="connect_js_script" as="connect_js" template="boltpay/connectjs.phtml" />
```
daisy1754: Since we have `add_button_to_minicart` config, can we make sure this only included when `add_button_to_minicart` is true or in cart page?

app/design/frontend/base/default/template/boltpay/replace.phtml
```diff
@@ -42,23 +44,28 @@ if(!Mage::helper('boltpay')->canUseBolt($this->getQuote(), false)) return;
...
 
-            var replaceRetries  = 20;
-            var replaceInterval = 500;
-
+            let replaceRetries  = 20,
+                replaceInterval = 500;
 
             // replaceable button selectors and types of Bolt buttons to replace them with
-            var selectors = {
+            let selectors = {
```
daisy1754: I tend to be conservative introducing new ES syntax (well let is not "new" but there are a few legacy browsers that don't support it). I suggest we stick to 'var'.

app/code/community/Bolt/Boltpay/Helper/Data.php
```diff
@@ -89,4 +98,75 @@ public function canUseForCountry($country)
...
+            return Mage::helper('core')->decrypt($key);
+        }
+
+        return $key;
+    }
+
+    /**
+     * @return array
+     */
+    public function getReplacementButtonSelectorsInMiniCart()
```
daisy1754: unused?

app/code/community/Bolt/Boltpay/etc/system.xml
```diff
@@ -355,11 +355,39 @@ This function is called when the Bolt checkout modal is closed.
...
+            <add_button_to_minicart translate="label">
+              <label>Add Bolt Button to the MiniCart Popup</label>
+              <frontend_type>select</frontend_type>
+              <source_model>adminhtml/system_config_source_yesno</source_model>
+              <sort_order>120</sort_order>
+              <show_in_default>1</show_in_default>
+              <show_in_website>1</show_in_website>
+              <show_in_store>0</show_in_store>
+            </add_button_to_minicart>
+            <replace_minicart_button_selectors>
```
daisy1754: Do we need to distinguish `replace_minicart_button_selectors` and `selectors` at line 195? Maybe it's simpler to always use `selectors`?

app/code/community/Bolt/Boltpay/etc/system.xml
```diff
@@ -355,11 +355,39 @@ This function is called when the Bolt checkout modal is closed.
               <show_in_store>0</show_in_store>
               <sort_order>106</sort_order>
             </close>
+            <add_button_to_minicart translate="label">
+              <label>Add Bolt Button to the MiniCart Popup</label>
```
daisy1754: I'm wondering whether we should name this in more generic way. For instance argento theme has popup with checkout button. It's not minicart per se. 

The way I see is we need an option to decide whether we only inject connect.js to cart page or we inject it everywhere

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
YevhenShBolt: The method uses in app/design/frontend/base/default/template/boltpay/replace.phtml::56 string

app/design/frontend/base/default/template/boltpay/replace.phtml
YevhenShBolt: ok

app/code/community/Bolt/Boltpay/Helper/Data.php
YevhenShBolt: ok, we will replace every button if it present.

app/code/community/Bolt/Boltpay/Helper/Data.php
YevhenShBolt: it now used

app/code/community/Bolt/Boltpay/Helper/Data.php
```diff
@@ -110,4 +111,74 @@ public function collectTotals($quote, $clearTotalsCollectedFlag = false) {
 
         return $quote;
     }
+
+    /**
+     * Get config value
+     *
+     * @return bool
+     */
+    public function isNeedAddButtonEverywhere()
```
daisy1754: nit: "is need" is a bit weird English. Maybe `shouldAddButtonEverywhere` ?

app/code/community/Bolt/Boltpay/Helper/Data.php
```diff
@@ -97,9 +96,11 @@ public function canUseForCountry($country)
      * @param Mage_Sales_Model_Quote $quote
      * $param boolean $clearTotalsCollectedFlag Determines whether to set setTotalsCollectedFlag to false
```
daisy1754: nit: remove this line

app/code/community/Bolt/Boltpay/etc/system.xml
```diff
@@ -355,11 +355,20 @@ This function is called when the Bolt checkout modal is closed.
               <show_in_store>0</show_in_store>
               <sort_order>106</sort_order>
             </close>
+            <add_button_everywhere translate="label">
+              <label>Add Bolt Button Everywhere</label>
+              <frontend_type>select</frontend_type>
+              <source_model>adminhtml/system_config_source_yesno</source_model>
```
daisy1754: <comment>By default, Bolt checkout button is displayed in cart page. Enable this option if you plan to display Bolt checkout outside of cart page, such as mini cart.</comment>

app/code/community/Bolt/Boltpay/Helper/Data.php
```diff
@@ -110,4 +111,74 @@ public function collectTotals($quote, $clearTotalsCollectedFlag = false) {
...
+     *
+     * @param $path
+     * @return bool
+     */
+    public function canShowEverywhere($path)
+    {
+        $active = Mage::getStoreConfigFlag('payment/boltpay/active');
+        $isEverywhere = $this->isNeedAddButtonEverywhere();
+
+        if (!$active) {
```
daisy1754: How about simply `return $active && $isEverywhere`?

app/design/frontend/base/default/layout/boltpay.xml
```diff
@@ -30,14 +30,47 @@
...
       <block type="boltpay/checkout_boltpay" name="prefetch_shipping_quote" as="prefetch_shipping_quote" template="boltpay/prefetch_shipping_quote.phtml" />
-      <block type="boltpay/checkout_boltpay" name="connect_js_script" as="connect_js" template="boltpay/connectjs.phtml" />
-      <block type="boltpay/checkout_boltpay" name="replace_script" as="replace_script" template="boltpay/replace.phtml" />
+
+      <block type="boltpay/checkout_boltpay" name="connect_js_script.cart" as="connect_js.cart" />
+      <block type="boltpay/checkout_boltpay" name="replace_script.cart" as="replace_script.cart" />
+
+      <reference name="connect_js_script.cart">
+        <action method="setTemplate">
+          <template helper="boltpay/data/canShowOnCheckoutCart">
```
daisy1754: I found this a bit confusing, as currently canShowOnCheckoutCart returns false when showEverywhere is true. I think the root of confusion is logic here is basically depends upon what we do in above `default` block.

Can we instead, always insert connect_js_script (and override one defined above, if there is anything already defined)?

app/code/community/Bolt/Boltpay/etc/system.xml
YevhenShBolt: Adde this as a comment to the config

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -311,6 +311,11 @@ public function getCartDataJs($multipage = true)
             $success = Mage::getStoreConfig('payment/boltpay/success');
             $close = Mage::getStoreConfig('payment/boltpay/close');
 
+            /*
+             * If we enable the config option "Add Bolt Button everywhere" - we should check the quote items
+             * before Bolt Popup open.
+             */
+            $isEmptyQuote = (!($sessionQuote->getItemsCollection()->count())) ? 'true' : 'false';
```
daisy1754: should we have this check earlier? For instance,`$this->createBoltOrder` at line 268 should not be called if cart is empty

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -687,20 +697,91 @@ public function getQuote()
...
+        return ($canAddEverywhere || $isAllowedOnCurrentPage);
+    }
+
+    /**
+     * Checking if allow the replace script.
+     *
+     * @return bool
+     * @throws Exception
+     */
+    public function isAllowedReplaceScriptOnCurrentPage()
```
daisy1754: maybe this method can be simplified as

```
return !$isFireCheckoutPage && this->isAllowedConnectJsOnCurrentPage();
```

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
YevhenShBolt: @daisy1754 
But it show when "replace.phtml" is initializing:
app/code/community/Bolt/Boltpay/Helper/Data.php::canUseBolt()

Current method do not check the cart items number. So if we have enabled the option "Add the Bolt Button Everywhere" the script will init the button even we do not have any items in the cart. This issue does not appear if, as like in the Luma theme, the buttons do not appear, but if the checkout button does not hide so we have such issue. We see the Bolt Button.

daisy1754: Just to clarify, I agree we need `$sessionQuote->getItemsCollection()->count()`

My proposal is move this check to above, around line 268. If cart is empty we should not call `$this->createBoltOrder` method

app/code/community/Bolt/Boltpay/Helper/Data.php
```diff
@@ -110,4 +117,51 @@ public function collectTotals($quote, $clearTotalsCollectedFlag = false) {
...
+    public function shouldAddButtonEverywhere()
+    {
+        return Mage::getStoreConfigFlag('payment/boltpay/add_button_everywhere');
+    }
+
+    /**
+     * @param bool $decrypt
+     * @return string
+     */
+    public function getPublishableKeyMultiPageKey($decrypt = false)
```
daisy1754: is this method used?

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
YevhenShBolt: now i understand, thanks, i will move it up

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
YevhenShBolt: simplified.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -312,20 +315,26 @@ public function getCartDataJs($multipage = true)
             $success = Mage::getStoreConfig('payment/boltpay/success');
             $close = Mage::getStoreConfig('payment/boltpay/close');
 
-
             //////////////////////////////////////////////////////
             // Generate and return BoltCheckout javascript.
             //////////////////////////////////////////////////////
+            $immutableQuoteId = ($immutableQuote) ? $immutableQuote->getId() : '';
```
dumega: is this necessary? I think if $immutableQuote is null there is no chance to step here

PR #63, opened 2018-05-29
Explicitly set quote belong to guest if customer id does not exist

dumega:
when processing Magento order creation in API,  getCustomerIsGuest is false even the customer isnot logged in, so we need to explicitly set it to false for guest. to avoid misjudgement in thrid-party modules

### DIFF
--- a/app/code/community/Bolt/Boltpay/Helper/Api.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Api.php
@@ -204,9 +204,9 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
             $immutableQuote->save();
         }
         // explicitly set quote belong to guest if customer id does not exist
-        if (!$quote->getCustomerId()) {
-            $quote->setCustomerIsGuest(true);
-            $quote->save();
+        if (!$immutableQuote->getCustomerId()) {
+            $immutableQuote->setCustomerIsGuest(true);
+            $immutableQuote->save();
         }
 
         $immutableQuote->getShippingAddress()->setShouldIgnoreValidation(true)->save();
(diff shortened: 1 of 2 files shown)


app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -203,6 +203,11 @@ public function createOrder($reference, $sessionQuoteId = null, $isAjaxRequest =
             $immutableQuote->setCustomerEmail($email);
             $immutableQuote->save();
         }
+        // explicitly set quote belong to guest if customer id does not exist
```
daisy1754: Can we move this to inside `getCartDataJs`? To me this seems initialization of `$immutableQuote`

daisy1754: @mccottry let's try not to optimize for corner cases. 

I prefer we set as much as information in immutableQuote at creation time

dumega: @daisy1754  @mccottry  yes, to add `setCustomerIsGuest` before initiate the order does make sense. and I still suggest to check and update it again in the order creation process. Between the initiation and creation of order, there can be a lot of other complicated functions that may effect, not only just consideration of `a second tab or window,`, so I think it is always safe to check it again.

daisy1754: @dumega sounds good please make these changes.

mccottry: @dumega It certainly doesn't hurt to check it again, however, understand that once values are  set in the `$immutableQuote`, all these values cannot be changed by the Magento system by design.  So, no matter how complex the actions are between bolt order creation, (i.e. initiation) , and Magento order save (i.e. creation of order), the value that you set during bolt order creation will be the same at Magento order save time on the `$immutable_quote`.

dumega: @mccottry sure, the magento system wont change it, but I mean other third-party modules,  eg. when we get quote from checkout session before create order, there is an event `custom_quote_process`, and we never know if there is any other third-party module would hook into this event and do something then update the `customerisguest`.  then that's why I suggest to still check it again

dumega: @daisy1754 @mccottry  after some testing, I found that we can not depend on the status of `customerisguest` of the parent quote, but still need to check if its parent quote has valid `customerid`, please check my last commit

PR #64, opened 2018-05-29
This eliminates firecheckout being loaded on non-firecheckout pages

mccottry:


app/design/frontend/base/default/template/boltpay/replace_firecheckout.phtml
```diff
@@ -0,0 +1,249 @@
...
+    var onloadCallback = function(){
+
+        /****************************************************************************************************
+         * Fire Checkout extension
+         ****************************************************************************************************/
+        var fireCheckout = function() {
+
+            if (typeof FireCheckout !== "function") return false;
+
+            <?php
```
daisy1754: nit: this if statement should be moved to around line 33

PR #67, opened 2018-06-01
Feature partial capture from magento

johnmccombs1:


daisy1754:
this test needs to be updated
https://github.com/BoltApp/bolt-magento1/blob/develop/tests/unit/testsuite/Bolt/Boltpay/Model/PaymentTest.php

app/code/community/Bolt/Boltpay/Model/Payment.php
```diff
@@ -280,6 +282,14 @@ public function capture(Varien_Object $payment, $amount)
             } elseif ($transactionStatus != self::TRANSACTION_COMPLETED) {
                 $message = sprintf('Capture attempted denied. Transaction status: %s', $transactionStatus);
```
YevhenShBolt: Maybe we can use helper for support translation
Something like:
Mage::helper('boltpay')->__('Capture attempted denied. Transaction status: %s', $transactionStatus)

daisy1754: Great point. Since it's not part of this PR I'll go ahead merging this PR and create asana task regarding translation

daisy1754: Actually I made this as future issue. Unless it's use-facing string, I think it's ok to only support English for near future
https://app.asana.com/0/436430054069245/695464419655877/f

PR #68, opened 2018-06-02
Prevents on-hold status from being changed by webhooks

mccottry:
Note: getSubtotalWithDiscount was used over trying to subtract tax and shipping from Grand Total due to the various number of ways that Grand Total can be configured, most notably, to be displayed without tax.

Also fixed shipping and tax issue.


app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -848,7 +865,7 @@ public function getShippingAndTaxEstimate( $quote )
         $shippingAddress = $quote->getShippingAddress();
         $shippingAddress->setCollectShippingRates(true)->collectShippingRates()->save();
 
-        $origTotalWithoutShippingOrTax = $this->getTotalWithoutTaxOrShipping($quote);
+        $original_discounted_price = $quote->getSubtotalWithDiscount();
```
daisy1754: nit: let's stick to camelCase

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -165,121 +165,141 @@ public function verify_hook($payload, $hmacHeader)
...
-        $payment->setAdditionalInformation('bolt_merchant_transaction_id', $transaction->id);
-        //add if called by ajax request, this info would be used for log
-        $payment->setAdditionalInformation('is_ajax_request', $isAjaxRequest);
-        $payment->setTransactionId($transaction->id);
-
-        Mage::helper('boltpay')->collectTotals($immutableQuote, true)->save();
-      
-        // a call to internal Magento service for order creation
-        $service = Mage::getModel('sales/service_quote', $immutableQuote);
+        } catch ( Exception $e ) {
```
daisy1754: should these try block cover until like 355? Maybe we want to re-activate quote when exception happens at `Mage::getModel('boltpay/payment')->handleOrderUpdate($order)` for instance.

app/code/community/Bolt/Boltpay/Model/Observer.php
```diff
@@ -157,6 +158,55 @@ public function addMessageWhenCapture($observer)
...
+    public function updateBoltTransactionStatus($observer)
+    {
+        /* @var Mage_Sales_Model_Order $order */
+        $order = $observer->getOrder();
+
+        /** @var Mage_Sales_Model_Order_Payment $payment */
+        $payment = $order->getPayment();
+
+        $method = $payment->getMethod();
+        if (strtolower($method) == Bolt_Boltpay_Model_Payment::METHOD_CODE) {
```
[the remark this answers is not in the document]
dumega: @daisy1754 I think Leon just try update the bolt transaction statut for order on merchant server, not bolt server

app/code/community/Bolt/Boltpay/Model/Payment.php
```diff
@@ -41,6 +41,7 @@ class Bolt_Boltpay_Model_Payment extends Mage_Payment_Model_Method_Abstract
     const TRANSACTION_CANCELLED = 'cancelled';
     const TRANSACTION_COMPLETED = 'completed';
     const TRANSACTION_PENDING = 'pending';
+    const TRANSACTION_ON_HOLD = 'on-hold';
```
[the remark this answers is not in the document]
dumega: I think this is a status container rather than a real status

app/code/community/Bolt/Boltpay/Model/Observer.php
```diff
@@ -95,13 +95,14 @@ public function verifyOrderContents($observer)
         if (strtolower($method) == Bolt_Boltpay_Model_Payment::METHOD_CODE) {
 
             $reference = $payment->getAdditionalInformation('bolt_reference');
+            $magento_total = Mage::getStoreConfig('tax/sales_display/grandtotal', $order->getStoreId()) ? (int)($order->getGrandTotal()*100) : (int)(($order->getGrandTotal()+$order->getTaxAmount())*100);
```
dumega: In my mind the transaction amount would always include tax, doesn't it?

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -871,12 +888,12 @@ public function getShippingAndTaxEstimate( $quote )
...
             }
 
-            $shippingDiscountModifier = $this->getShippingDiscountModifier($origTotalWithoutShippingOrTax, $quote);
+            $new_discounted_price = $quote->getSubtotalWithDiscount();
 
             $option = array(
                 "service" => $label,
                 "reference" => $rateCode,
-                "cost" => round(($quote->getShippingAddress()->getShippingAmount() - $shippingDiscountModifier) * 100),
+                "cost" => round(($quote->getShippingAddress()->getShippingAmount() + ($original_discounted_price - $new_discounted_price)) * 100),
```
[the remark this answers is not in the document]
johnmccombs1: Technically this should be the same, and we haven't had any instances that I'm aware of that break this. The only reason I was doing it differently is that Magento allows you to add a new total collector that can do pretty much anything you want. My code was meant to allow for total collectors that modify the quote outside of shipping, tax, and subtotal. So for example, if a 3rd party extension added a special handling fee for expedited shipping and it wasn't applied to the subtotal, the code here may give invalid results. It's definitely an edge case, but that was the reason for the more complex logic. We could leave it out and wait until we actually come across a problem like this. I have never seen something like this in Magento, but I was planning for future problems that may never occur.

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -165,121 +165,141 @@ public function verify_hook($payload, $hmacHeader)
...
+                throw new Exception("The quote has expired." );
+            }
+          
+            // check if this order is currently being proccessed.  If so, throw exception
+            /* @var Mage_Sales_Model_Quote $parentQuote */
+            $parentQuote = Mage::getModel('sales/quote')->loadByIdWithoutStore($immutableQuote->getParentQuoteId());
+            if ($parentQuote->isEmpty() || !$parentQuote->getIsActive()) {
+                throw new Exception("The quote is currently being processed.");
+            } else {
+                $parentQuote->setIsActive(false);
```
johnmccombs1: Should we save the parent quote to the database so the a web hook that comes later will be able to pull the active status?

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -165,121 +165,141 @@ public function verify_hook($payload, $hmacHeader)
...
-        $payment->setTransactionId($transaction->id);
-
-        Mage::helper('boltpay')->collectTotals($immutableQuote, true)->save();
-      
-        // a call to internal Magento service for order creation
-        $service = Mage::getModel('sales/service_quote', $immutableQuote);
+        } catch ( Exception $e ) {
+            // Order creation failed, so mark the parent quote as active so webhooks can retry it
+            if ($parentQuote) {
+                $parentQuote->setIsActive(true);
```
johnmccombs1: Should we call save here so that setting it to active will be restored in the database?

app/code/community/Bolt/Boltpay/Model/Observer.php
johnmccombs1: To Daniel's point, I don't think I've ever seen a grand total that didn't include tax, but if you have I'd be interested in knowing what caused it. Also, casing should probably be camel case.

app/code/community/Bolt/Boltpay/Helper/Api.php
mccottry: No.  The code is correct as is.  If there is an exception prior to 297, it means the order was never created in Magento and the web hooks should be free to try to create it.

If there is an exception after that point, it means the order is already in Magento.  Web hooks should not be able to create a new order.

app/code/community/Bolt/Boltpay/Model/Observer.php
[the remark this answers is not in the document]
mccottry: `boltTransactionStatus` is a legacy field that appeared to save the transaction status at only order creation time.  More importantly, this field has been exclusively used to determine what are valid **Magento** order transitions.  So, the code is mixing the ideas Bolt order status and Magento order status and lacked mappings for cases many cases like 'on-hold' which is not a 
Bolt Status, and other subtle ones like deferred.

Important things are though it is not semantically correct, the mappings are functionally correct with how the system uses this field.  If we wish for it to be semantically correct, then we drop this field and simply rely on Magento order state, removing all `Payment::TRANSACTION_*` constants. and updating `Payment::$_validStateTransitions`, and methods `Payment::translateHookTypeToTransactionStatus`, and `Payment::handleTransactionUpdate` at minimum.  Because this seemed to me to be a very heavy-handed approach compared to the simpler and stable solution presented, I evaluated the risk of such a change for semantic value out-weighing the benefits, I opted not to pursue this route.

However, both approaches will function the same.  Perhaps renaming the field can be the happy medium if this naming presents a problem with code comprehension.

app/code/community/Bolt/Boltpay/Model/Payment.php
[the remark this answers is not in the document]
mccottry: @daisy1754 That is true.  And these transactions are all being used as "Magento order statuses" and are not exactly what the original code is implying in the name.  As the code is written, the payment class is best thought of as a status interpreter or container rather than reflecting real Bolt status.  In this light, it lacked an interpretation of the Magento 'on-hold' order status, and this as well as the other parts of the code, add this to the plugin.

app/code/community/Bolt/Boltpay/Model/Payment.php
```diff
@@ -478,7 +480,7 @@ public function handleTransactionUpdate(Mage_Payment_Model_Info $payment, $newTr
                 //Mage::log(sprintf("Valid next states from %s: %s", $prevTransactionStatus, implode(",",$validNextStatuses)), null, 'bolt.log');
 
                 if (!in_array($newTransactionStatus, $validNextStatuses)) {
```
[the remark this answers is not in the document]
mccottry: I mentioned in the comment above that this is very old code that I saw to update instead of re-design.  Absolutely, Magento's order status can be used by making the updates described above.

app/code/community/Bolt/Boltpay/Model/Observer.php
mccottry: @dumega In my mind too.
@johnmccombs1 I never saw it before until I ran into a case where it made it clear.  The case is in the configuration listed above.

+1 for camelCase.  The change will be applied.

app/code/community/Bolt/Boltpay/Helper/Api.php
mccottry: The parent quote is automatically saved to the database.  Web hooks should not pull from the parent quote, but from the parent quote clone.  This is the entire idea of not relying on front end manipulated data at order creation time.

mccottry: good call.

app/code/community/Bolt/Boltpay/Helper/Api.php
mccottry: good call.

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -871,12 +888,12 @@ public function getShippingAndTaxEstimate( $quote )
                 Mage::helper('boltpay/bugsnag')->notifyException(new Exception('Rate code is empty. ' . var_export($rate->debug(), true)));
             }
 
-            $shippingDiscountModifier = $this->getShippingDiscountModifier($origTotalWithoutShippingOrTax, $quote);
+            $new_discounted_price = $quote->getSubtotalWithDiscount();
```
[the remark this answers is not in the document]
mccottry: @daisy1754 Everything is possible.  Here, particularly, the incorrect shipping cost and therefore grand_total cost is at the root of orders being placed directly to on-hold, and is, in a sense, at the core of the original problem that this PR is trying to solve.  Let me know if a separate PR is how you want to proceed or if the two or three related lines of code are best reviewed in this PR for a merge that solves the all the problems which needed to be addressed.

app/code/community/Bolt/Boltpay/Model/Payment.php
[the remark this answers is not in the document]
daisy1754: ak ok so it's not bolt's status. It is Payment status defined in Boltpay module.

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -165,121 +165,141 @@ public function verify_hook($payload, $hmacHeader)
...
+            // check if quote has already been used
+            if ( !$immutableQuote->getIsActive() ) {
+                throw new Exception("The quote has expired." );
+            }
+          
+            // check if this order is currently being proccessed.  If so, throw exception
+            /* @var Mage_Sales_Model_Quote $parentQuote */
+            $parentQuote = Mage::getModel('sales/quote')->loadByIdWithoutStore($immutableQuote->getParentQuoteId());
+            if ($parentQuote->isEmpty() || !$parentQuote->getIsActive()) {
+                throw new Exception("The quote is currently being processed.");
```
daisy1754: 'The quote'. $immutableQuote->getParentQuoteId() .'is currently being processed for is already processed'

app/code/community/Bolt/Boltpay/Model/Observer.php
```diff
@@ -157,6 +158,55 @@ public function addMessageWhenCapture($observer)
         }
     }
 
+    /**
+     * Updates the Bolt transaction status on order status change.
+     *
```
daisy1754: Add comment like "note this transaction status is not necessarily same as order status in Bolt server. bolt_transaction_status field keeps track of payment status from magento plugin perspective."

app/code/community/Bolt/Boltpay/etc/config.xml
```diff
@@ -126,8 +134,34 @@
           </deferred>
         </states>
       </order>
+      <!-- Adds total sorting refinement due to the changes in uasort between PHP5 and PHP7 -->
```
daisy1754: is this part of this on-hold change? If not, can you create a separate PR (I can approve instantly).

mccottry: In PHP7, for Magento 1.9.x, the grand total will always be wrong in the store backend without this being set, so, for me, this is part of the same pull request.  FYI, nnly in Wordpress/WooCommerce, for total code structure change have I placed several unrelated code changes into the same pull request.  Outside of this, I always have and always will separate unrelated code changes as separate PR as my standard practice.

johnmccombs1:
I added a fix for the mismatch error we were seeing.

daisy1754:
This PR is now 800 lines and it's hard to review. I extracted some changes as separate PRs #105 #106 #107 #108. I recommend you to do similar if there is another change that can be extracted.

daisy1754:
Okay let's merge this...

daisy1754:
🎉

PR #72, opened 2018-06-04
Bypass getShippingAndTaxEstimate if post code is invalid.

johnmccombs1:
Here is the new branch and pull request for overriding shipping and tax estimate when prefetch makes request with invalid zip code.

See also: https://github.com/BoltApp/bolt-magento1/pull/55

### DIFF
--- a/app/code/community/Bolt/Boltpay/controllers/ShippingController.php
+++ b/app/code/community/Bolt/Boltpay/controllers/ShippingController.php
@@ -221,7 +221,12 @@ class Bolt_Boltpay_ShippingController extends Mage_Core_Controller_Front_Action
             try {
                 /** @var Bolt_Boltpay_Helper_Api $helper */
                 $helper = Mage::helper('boltpay/api');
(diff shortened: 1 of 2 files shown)


app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -221,7 +221,12 @@ public function prefetchEstimateAction()
             try {
                 /** @var Bolt_Boltpay_Helper_Api $helper */
                 $helper = Mage::helper('boltpay/api');
-                $estimateResponse = $helper->getShippingAndTaxEstimate($quote);
+
+                if (@$addressData['postcode']) {
+                    $estimateResponse = $helper->getShippingAndTaxEstimate($quote);
+                } else {
+                    $estimateResponse = $helper->getDefaultShippingAndTaxResponse();
```
daisy1754: Let's just return here (no cache). I don't want to cache default response when `@$addressData['postcode']` is missing for some glitch

PR #79, opened 2018-06-11
Implement virtual terminal for magento1

mccottry:
https://app.asana.com/0/544708310157130/681387159814229


app/design/adminhtml/default/default/template/boltpay/form.phtml
```diff
@@ -24,15 +24,77 @@
...
         </div>
     </li>
 </ul>
+
+<script type="text/javascript">
+    (function() {
+
+        //////////////////////////////////////////////////
+        //  JSONP function to reliable load connect.js
+        //  after ajax reloads of this block's template
```
daisy1754: Why this is needed?

daisy1754:
@mccottry I tried this PR but keep getting "Please specify a shipping method." error after authorizng the payment.

daisy1754:
Also in case of guest, looks like email is missing.

app/code/community/Bolt/Boltpay/controllers/ApiController.php
```diff
@@ -87,17 +87,37 @@ public function hookAction()
                 //Mage::log('Order Found. Updating it', null, 'bolt.log');
                 $orderPayment = $order->getPayment();
 
+                // set meta payment data
```
daisy1754: why this is needed? If it's not new should't these parameter already set?

app/design/frontend/base/default/template/boltpay/replace.phtml
```diff
@@ -153,7 +153,7 @@ if(!$this->isAllowedReplaceScriptOnCurrentPage() || !$this->canUseBolt() ) retur
                                      * Creating the order at Bolt end, if applicable (cart not empty).
                                      * Generating BoltCheckout code.
                                      *****************************************************************/
-									<?=$this->getCartDataJs(); ?>
+									<?=$this->getCartDataJs('one-page'); ?>
```
daisy1754: I think this is 'multi-page'?

app/design/frontend/base/default/template/boltpay/track.phtml
```diff
@@ -34,7 +34,7 @@ $version = (string)$versionElm[0];
             id="bolt-track"
             type="text/javascript"
             src="<?= $this->getTrackJsUrl(); ?>"
-            data-publishable-key="<?=$this->getPublishableKey(); ?>">
+            data-publishable-key="<?=$this->getPublishableKey('multi-page'); ?>">
```
daisy1754: We can live with this for now but ideally this should be either one-page or multi-page, depending on which one is set.

app/design/adminhtml/default/default/template/boltpay/form.phtml
mccottry: This is the same pattern used on the onepage checkout, which is very similar to the admin page.  So, the quick answer is for reliable convenience. 

But also, it is true that in the past, it was the way to style and attach events to buttons.  It is true that calling BoltConnect could be enough, but to load it with a dynamically assigned publishable key would require rewriting the head of the admin page, creating an additional or overriding block, creating additional templates, a lot of XML configuration, etc. or some similar tricks.  So, need is too strong a word.  

It is done this way because it is easier with little cost, and it is also status quo.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -328,16 +327,35 @@ public function getCartDataJs($multipage = true)
                     json_cart,
                     $jsonHints,
                     {
-                      check: function() {
+                      check: function() {           
+                        $check
```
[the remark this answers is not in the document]
mccottry: When I first implemented this some months back, I consciously made the decision to have the `$check` and all merchant code injected to be executed as soon as possible, including before our code to provide them with the means to override our behavior.  Placing it after always eliminates that option, and it is the opposite of what we desire to do. 

Back in the day, I had decided that the check for a Bolt order token was imperative and if missing was an irrecoverable error.  Therefore, I made what was intended to be the one exception there, but in seeing how people have added to the code and your question, probably should have internally documented the reasoning.  Some months later, and seeing the code in practice, especially in using it to do merchant side overrides, I came to realize that a missing order token was not irrecoverable and that it is always better to give as much power to the merchant's developers as possible.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -328,16 +327,35 @@ public function getCartDataJs($multipage = true)
...
+                        if ((typeof editForm !== 'undefined') && (typeof editForm.validate === 'function')) {
```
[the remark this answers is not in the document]
mccottry: Not a problem at all.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -363,7 +381,13 @@ public function getCartDataJs($multipage = true)
                         $onPaymentSubmit
                       },
                       
-                      success: function(transaction, callback) {
+                      success: function(transaction, callback) { 
+                        // order and order.submit will exist for admin
+                        if ((typeof order !== 'undefined' ) && (typeof order.submit === 'function')) {
```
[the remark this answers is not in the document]
mccottry: no problem.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -381,6 +405,19 @@ function() {
                       
                       close: function() {
                          $close
+                         //////////////////
+                         // admin logic
+                         //////////////////
+                         if (order_completed && (typeof order !== 'undefined' ) && (typeof order.submit === 'function')) {
```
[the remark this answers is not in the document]
mccottry: yep.

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -317,9 +314,38 @@ public function createOrder($reference, $sessionQuoteId = null, $isAjaxRequest =
...
+    }
+
+    /**
+     * Sets the order to being in the new state and annotates it with creation meta data
+     *
+     * @param Mage_Sales_Model_Order    $order          the newly created order
+     * @param string                    $reference      the Bolt transaction reference
+     * @param bool                      $wasCreatedByFrontend  true if order was created via ajax, false if via webhook
+     */
+    private function markOrderAsNew($order, $reference, $wasCreatedByFrontend) {
```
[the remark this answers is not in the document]
mccottry: This was done in another branch and can be merged here.  I didn't read this until now time, but it is exactly as you describe here with a small difference in the name, which is `setInitialOrderStatus`

app/code/community/Bolt/Boltpay/controllers/Adminhtml/Sales/Order/CreateController.php
```diff
@@ -0,0 +1,78 @@
...
+            'last_name'       => $shippingAddress['lastname'],
+            'locality'        => $shippingAddress['city'],
+            'region'          => Mage::getModel('directory/region')->load($shippingAddress['region_id'])->getCode(),
+            'postal_code'     => $shippingAddress['postcode'],
+            'country_code'    => $shippingAddress['country_id'],
+            'phone'           => $shippingAddress['telephone'],
+            'phone_number'    => $shippingAddress['telephone'],
+        );
+
+        if ($email) {
```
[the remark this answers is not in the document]
mccottry: true, will do.

app/design/frontend/base/default/template/boltpay/replace.phtml
mccottry: I know that you are right. Updating.

app/design/frontend/base/default/template/boltpay/track.phtml
mccottry: Sure.  It has always been the way it is now.  I was only going through the code and explicitly writing out implicit parameters.  No need to wait.  We can implement that now.  What it means is that in production, there are probably some cases where people only have a payment only key and their track.js is not working properly.

PR #88, opened 2018-06-19
Meta stored in observer

mccottry:
https://app.asana.com/0/544708310157130/713020115254919


app/code/community/Bolt/Boltpay/Model/Observer.php
```diff
@@ -173,4 +173,50 @@ protected function _addMagentoOrderIdToMessage($incrementId)
...
+
+        $reference = Mage::getSingleton('core/session')->getBoltReference();
+        $transaction = Mage::getSingleton('core/session')->getBoltTransaction() ?: Mage::helper('boltpay/api')->fetchTransaction($reference);
+
+        $boltCartTotal = $transaction->amount->currency_symbol. ($transaction->amount->amount/100);
+        $orderTotal = $order->getGrandTotal();
+
+        $hostname = Mage::getStoreConfig('payment/boltpay/test') ? "merchant-sandbox.bolt.com" : "merchant.bolt.com";
+        if(Mage::getSingleton('core/session')->getWasCreatedByMagento()){ // order is create via AJAX call
+            $msg = sprintf(
```
daisy1754: nit: you can move this outside of if statement, and append extra text in case of hook.

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -263,8 +263,25 @@ public function createOrder($reference, $sessionQuoteId = null, $isAjaxRequest =
         $service = Mage::getModel('sales/service_quote', $immutableQuote);
 
         try {
+            ///////////////////////////////////////////////////////
+            /// These values are used in the observer after successful
+            /// order creation
+            ///////////////////////////////////////////////////////
+            Mage::getSingleton('core/session')->setBoltTransaction($transaction);
+            Mage::getSingleton('core/session')->setBoltReference($reference);
+            Mage::getSingleton('core/session')->setWasCreatedByMagento($isAjaxRequest);
```
daisy1754: order is always created via magento. It's either via Ajax request or via hook.
Maybe instead `wasCreatedByHook`?

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -307,13 +307,23 @@ public function getCartDataJs($checkoutType = 'multi-page')
...
             // opportunity to do full overrides
             //////////////////////////////////////////////////////
-            $check = Mage::getStoreConfig('payment/boltpay/check');
-            $onCheckoutStart = Mage::getStoreConfig('payment/boltpay/on_checkout_start');
-            $onShippingDetailsComplete = Mage::getStoreConfig('payment/boltpay/on_shipping_details_complete');
-            $onShippingOptionsComplete = Mage::getStoreConfig('payment/boltpay/on_shipping_options_complete');
-            $onPaymentSubmit = Mage::getStoreConfig('payment/boltpay/on_payment_submit');
-            $success = Mage::getStoreConfig('payment/boltpay/success');
-            $close = Mage::getStoreConfig('payment/boltpay/close');
+            if (($checkoutType === 'admin') && !Mage::getStoreConfig('payment/boltpay/use_javascript_in_admin')) {
```
daisy1754: How about we always keep these empty if checkouttype is admin? (so we don't add `use_javascript_in_admin`). 
We can add flag later when really needed, but once we add flag it's our burden to maintain.

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -263,8 +263,25 @@ public function createOrder($reference, $sessionQuoteId = null, $isAjaxRequest =
...
+
             $service->submitAll();
         } catch (Exception $e) {
+            ///////////////////////////////////////////////////////
+            /// Unset session values set above
+            ///////////////////////////////////////////////////////
+            Mage::getSingleton('core/session')->unsBoltTransaction();
+            Mage::getSingleton('core/session')->unsBoltReference();
+            Mage::getSingleton('core/session')->unsWasCreatedByMagento();
+            ///////////////////////////////////////////////////////
```
YevhenShBolt: I recommend moving this three lines to the separate method.

app/code/community/Bolt/Boltpay/Helper/Bugsnag.php
```diff
@@ -154,19 +154,14 @@ public function setBoltTraceId($traceId)
...
-        }
-
-        $traceId = @$_SERVER['HTTP_X_BOLT_TRACE_ID'];
-
-        if(!empty($traceId)) {
-            return $traceId;
-        }
-
-        return null;
+        return $this->boltTraceId ?: @$_SERVER['HTTP_X_BOLT_TRACE_ID'];
```
YevhenShBolt: I sugget to move $_SERVER to the separate method and then we can use something like that:

Create di.xml and
```xml
<?xml version="1.0"?>
<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:noNamespaceSchemaLocation="urn:magento:framework:ObjectManager/etc/config.xsd">
    <!-- Restore original visitor IP from CloudFlare header -->
    <type name="Magento\Framework\HTTP\PhpEnvironment\RemoteAddress">
        <arguments>
            <argument name="alternativeHeaders" xsi:type="array">
                <item name="http_cf_connecting_ip" xsi:type="string">HTTP_CF_CONNECTING_IP</item>
            </argument>
        </arguments>
    </type>
</config>
```

and then we can use something like:
```php
    /**
     * @var \Magento\Framework\HTTP\PhpEnvironment\RemoteAddress
     */
    protected $_remoteAddressInstance;

...

    /**
     * Used to get remote address, in which cookie data is set
     *
     * @return string returns remote address
     */
    public function getRemoteAddress()
    {
        return $this->_remoteAddressInstance->getRemoteAddress();
    }
```

app/code/community/Bolt/Boltpay/Model/Observer.php
```diff
@@ -173,4 +173,50 @@ protected function _addMagentoOrderIdToMessage($incrementId)
...
+            return;
+        }
+
+        $reference = Mage::getSingleton('core/session')->getBoltReference();
+        $transaction = Mage::getSingleton('core/session')->getBoltTransaction() ?: Mage::helper('boltpay/api')->fetchTransaction($reference);
+
+        $boltCartTotal = $transaction->amount->currency_symbol. ($transaction->amount->amount/100);
+        $orderTotal = $order->getGrandTotal();
+
+        $hostname = Mage::getStoreConfig('payment/boltpay/test') ? "merchant-sandbox.bolt.com" : "merchant.bolt.com";
```
YevhenShBolt: Can we move it to the model and then use it like constant?

Like `Bolt_Boltpay_Model_Payment::URL_MERCHANT_SANDBOX`

app/code/community/Bolt/Boltpay/Helper/Bugsnag.php
mccottry: @YevhenShBolt  Why would I make something that is one line of code into something that is four lines of code spanning across multiple methods?

To me, that is highly unnecessary and makes the code harder to read and needlessly more complex.

I liken your suggestion to `goto` construct, that was used in the late 50s, 60s, 70s, and 80s.  In that style, we are jumping around in the code for no good reason.  

If there was some sort of sanitizing of values that were needed in looking into the array, then there is a point.  But, in this case, there seems to be no point, and we do not need to wrap a method around an array to pull its raw values.

My point is this:
Look at the old code, 8 lines.
Look at the new code, equivalent and only 1 line.

And what I understand the suggestion to be is  something like this:

```
...
return $this->boltTraceId ?: $this->_getHeaderField('HTTP_X_BOLT_TRACE_ID');
...

/**
 * Method to an element from the $_SERVER array
 */
 private function _getHeaderField( $key ) {
     return @$_SERVER[$key];
}
```

So, one line now becomes six with a fairly useless and logically obscuring method.

I disagree with this approach and consider it an anti-pattern.

app/code/community/Bolt/Boltpay/Model/Observer.php
mccottry: @daisy1754 I basically copied and pasted this code, and I can now streamline line it too!

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php

Currently, I have set the default value of this to be `false`.

So, for all active merchants, they will have to actively turn this on. If they actively turn this on, it would probably be for a reason good reason, like calling an external API.

I do not see this as causing a superfluous burden, but rather staying consistent with our docs and adding a simple point of override if when necessary for those who actively enable this feature.

What do you think?

app/code/community/Bolt/Boltpay/Helper/Api.php
mccottry: yep!

app/code/community/Bolt/Boltpay/Helper/Api.php
YevhenShBolt: My goals are:
First, I recommended to move the logic to separate method where we can manage what we want to unset.
Second, it will useful for PHPUnit test to cover it.

I thought so. Basically, you can skip my recommendation and maybe it would be better to refactor it when it needed.

app/code/community/Bolt/Boltpay/Helper/Bugsnag.php
YevhenShBolt: My goal is to simply it.
I want to make something like:

```php
...
return $this->boltTraceId ?: $this->_getServerBoltTraceId();
...

/**
 * Method to get an element from the $_SERVER array
 */
 private function _getServerBoltTraceId() {
     return $_SERVER['HTTP_X_BOLT_TRACE_ID'];
}
```

Here, we can modify it as we want and will not touch the base view of `public function getBoltTraceId();`
For example, if the merchant will use the Cloudflare or varnish the headers will be removed and we should use another name or logic to determine which header we needed.
In such way, i see that such method will be useful.

@daisy1754  and @mccottry . Guys, if you think that for now, it will be fairly useless - please ignore my recommendation. Maybe I'm getting paranoid ;-)

app/code/community/Bolt/Boltpay/Helper/Api.php
mccottry: @YevhenShBolt Feel free to refactor it when you think necessary.  When I thought of how the code would look after this, I cannot justify such a change myself.

app/code/community/Bolt/Boltpay/Helper/Bugsnag.php
mccottry: @YevhenShBolt IMO, we should stop adding methods like this to the code for the reasons I explained above.  In your example with Cloudflare and Varnish, if it were on a merchant's site and this sort of problem arose, then you still would have to adjust the method `public function getBoltTraceId();` because the private method could not be overridden.

Obviously, you could make your helper method of the helper method protected or public.  But, just re-read that... we have a one-line helper method for another one-line helper method that create 6 to 11 lines of code for something that can effectively be accomplished in one simple line.

When making a helper method, we should ask ourselves at least three fundamental questions.

1.)  Does this method significantly reduce the complexity of a code block?
2.)  Is this method likely to be reused in other than one place in the code?
3.)  Is it likely that I will need to override this specific functionality?

If you can make a strong case for any of these things, then a helper method is warranted.  Here, I think the method you will want to override would be `public function getBoltTraceId()`.  

For the cases you described above, (i.e. headers being rewritten), then I think the scope of these functions is too narrow for the projected problem and I would recommend handling that in another branch/pull request.

YevhenShBolt: ok

PR #93, opened 2018-06-22
assures that email is added to the prefill values for admin modal

mccottry:
https://app.asana.com/0/544708310157130/713162692866472

### DIFF
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -459,19 +459,18 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
             }
         }
 
-        // If address exists populate the hints array with existing address data.
-        if ($address) {
-            if (@$address->getEmail())     $hints['email']        = $address->getEmail();
-            if (@$address->getFirstname()) $hints['firstName']    = $address->getFirstname();
-            if (@$address->getLastname())  $hints['lastName']     = $address->getLastname();
-            if (@$address->getStreet1())   $hints['addressLine1'] = $address->getStreet1();
-            if (@$address->getStreet2())   $hints['addressLine2'] = $address->getStreet2();
-            if (@$address->getCity())      $hints['city']         = $address->getCity();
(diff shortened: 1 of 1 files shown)


app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -436,29 +436,30 @@ function() {
...
-        ///////////////////////////////////////////////////////////////
+        /////////////////////////////////////////////////////////////////////////
         $address = $quote->getShippingAddress();
-
-        if ($session && $session->isLoggedIn()) {
-            /** @var Mage_Customer_Model_Customer $customer */
-            $customer = Mage::getModel('customer/customer')->load($session->getId());
-            $address = $customer->getPrimaryShippingAddress();
-            $hints['email'] = $customer->getEmail();
+        if (!$address->getStreet1()) {
```
daisy1754: what is this check for?

mccottry: https://github.com/BoltApp/bolt-magento1/blob/af9f8d6f24c62efc27a415d13bec0c644afd584d/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php#L448-L454

The checks do exactly what the comments above say that they do.
The first checks if there is a valid shipping address set in the quote instead of an empty address with null values.  This is accomplished by checking if the mandatory field of *street1* is set.

If this is not set, then the next line check if the customer is logged in.  If so, then we get the address data from the primary shipping address saved by the customer.

https://github.com/BoltApp/bolt-magento1/blob/af9f8d6f24c62efc27a415d13bec0c644afd584d/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php#L463-L473

line 463, as I read it, will always evaluate to true.  Additionally, the `@` error message suppression operators in lines 464-473 are also unnecessary, as they are called on a `Varien_Object` that has a magic __get method that handles this and will return null without message.  I would normally remove both of these to get the pull request merged more quickly, but, I think now that it is better to make the changes and answer all questions which arise in order to continually improve the codebase.

daisy1754: If we know address is always non-null, and you make change here at line 453 to account for that, let's cleanup line 463 as well.

PR #96, opened 2018-06-24
Adds unhold logic while preventing hooks from unholding orders

mccottry:
https://app.asana.com/0/544708310157130/707109894457835
https://app.asana.com/0/544708310157130/687916964711091

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Payment.php
+++ b/app/code/community/Bolt/Boltpay/Model/Payment.php
@@ -45,6 +45,7 @@ class Bolt_Boltpay_Model_Payment extends Mage_Payment_Model_Method_Abstract
     const TRANSACTION_REJECTED_REVERSIBLE = 'rejected_reversible';
     const TRANSACTION_REJECTED_IRREVERSIBLE = 'rejected_irreversible';
     const TRANSACTION_NO_NEW_STATE = 'no_new_state';
+    const TRANSACTION_ALL_STATES = 'all_states';
 
     const HOOK_TYPE_AUTH = 'auth';
     const HOOK_TYPE_CAPTURE = 'capture';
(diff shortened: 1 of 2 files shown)


app/code/community/Bolt/Boltpay/Model/Payment.php
```diff
@@ -102,6 +103,19 @@ class Bolt_Boltpay_Model_Payment extends Mage_Payment_Model_Method_Abstract
...
+
+    /**
+     * Bolt_Boltpay_Model_Payment constructor.
+     *
+     * Allows transitions from on-hold from the non-webhook context
+     */
+    public function __construct()
+    {
+        if (!Bolt_Boltpay_Helper_Api::$fromHooks) {
+            $this->_validStateTransitions[self::TRANSACTION_ON_HOLD] = array(self::TRANSACTION_ALL_STATES);
```
daisy1754: instead of introducing new magic state, can we do:
`array_push($this->_validStateTransitions[self::TRANSACTION_ON_HOLD], self::TRANSACTION_COMPLETED, self::TRANSACTION_PENDING, self::TRANSACTION_AUTHORIZED))`

mccottry: @daisy1754 That was my first thought.  I believe the magic state is better because clients may have several states that they wish to transition manually which may lie outside any bounds which we may define.

In anticipation of this, the second magic state was introduced, like the TRANSACTION::NO_NEW_STATE

danac-gs: I agree with @mccottry on magic states.

daisy1754: Hmm so it sounds like we should allow transition from *any* state to *any* state for manual update, shouldn't we?

mccottry: @daisy1754 I believe that is true and constitutes a change request that is created in a different branch.  It is not related to unholding.

johnmccombs1:
Sorry for being late on this, but I've reviewed all changes and feel comfortable with them.

PR #99, opened 2018-06-26
Firecheckout/OnePageCheckout prevent Bolt Order Creation if no shipping method is specified

mccottry:
This is for non-virtual carts

### DIFF
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -254,6 +254,17 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
                         ->setParentQuoteId($sessionQuote->getId())
                         ->save();
 
+                    ///////////////////////////////////////////////////////////////////////////////
+                    // if we are missing a shipping method in one-page for a non-virtual cart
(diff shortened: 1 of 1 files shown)


app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -254,6 +254,17 @@ public function getCartDataJs($checkoutType = 'multi-page')
                         ->setParentQuoteId($sessionQuote->getId())
                         ->save();
 
+                    ///////////////////////////////////////////////////////////////////////////////
+                    // if we are missing a shipping method in one-page for a non-virtual cart
+                    // then, we will return a '"needs_shipping";' string meaning that no call to
+                    // BoltCheckout.configure will be made because a shipping address is required
+                    // under these conditions in order to make that call
+                    ///////////////////////////////////////////////////////////////////////////////
+                    if ($checkoutType !== 'multi-page' && !$immutableQuote->isVirtual() && !$immutableQuote->getShippingAddress()->getShippingMethod() ) {
```
daisy1754: should we bugsnag?

It's up to you.  

Pro:  When this happens unexpectedly, there is a note in Bugsnag saying that this occurred.

Con: This expected to happen very often in refresh cases where only partial address info has been filled out and the user is filling out extra address info.  So, bugsnag would be cluttered and the reason I initially reasoned not to include it.

johnmccombs1:
I'm good with these changes as well.

PR #104, opened 2018-06-29
Off by penny fix

mccottry:
Also includes 
- merge to `develop` corrections
- $isEmptyQuote fix

### DIFF
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -130,6 +130,7 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
             $isEmptyQuote = (!($sessionQuote->getItemsCollection()->count())) ? true : false;
 
             if (!$isEmptyQuote) {
+
                 ///////////////////////////////////////////////////////////////////////////////////////
                 // Merchant scope: get "bolt_user_id" if the user is logged in or should be registered,
                 // sign it and add to hints.
@@ -139,6 +140,7 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
                     $signRequest = array(
                         'merchant_user_id' => $reservedUserId,
                     );
+
                     $signResponse = $boltHelper->transmit('sign', $signRequest);
 
                     if ($signResponse != null) {
@@ -190,6 +192,8 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
                     $boltHelper->applyShippingRate($sessionQuote, $shippingMethod);
                 }
 
+            } else {
+                return '';
             }
 
             $authCapture = (Mage::getStoreConfig('payment/boltpay/auto_capture') == self::AUTO_CAPTURE_ENABLED);
@@ -559,7 +563,7 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
     {
         return Mage::getStoreConfig('payment/boltpay/skip_payment');
     }
-    
+
     /**
      * Returns whether enable merchant scoped account.
      * @return string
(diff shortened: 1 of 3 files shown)


app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -190,6 +192,8 @@ public function getCartDataJs($checkoutType = 'multi-page')
                     $boltHelper->applyShippingRate($sessionQuote, $shippingMethod);
                 }
 
+            } else {
+                return '';
```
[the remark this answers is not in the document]
mccottry: @YevhenShBolt This is based from your change and without this, it is throwing an error.  I am very unfamiliar with the "Add Bolt Button Everywhere", so can you give a better explanation of what the problem is when using that new feature?  Otherwise, can you make the change that you are suggesting as a Pull Request to this branch?

From what I understand, you want to return a javascript alert.  My concern is that having a javascript alert pop up on the shopping cart page the already says *Shopping Cart is Empty* will be perceived as an intrusive Bolt introduced problem to most merchants.

My proposal is to make the *Add Bolt Button Everywhere* conform to receiving an empty string, as the point of `getCartDataJs` is to return an executable javascript string. and, if the quote is empty, we shouldn't be generating any cart JS.  If the quote is empty, then either it does not call 
`getCartDataJs`, or if it calls this method, and it receives an empty string, then the Bolt Button should not be rendered.

I assume the *Add Bolt Button Everywhere* button is the mini-cart and not an "Instant Buy" type button.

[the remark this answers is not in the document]
daisy1754: @mccottry "add bolt button everywhere" is a mini-cart. For minicart, we want to keep showing bolt button and show alert when user clicks Bolt button.
If you look at existing code below, we have something like below, here we leverage `check` function to check isEmptyQuote flag and show alert.

```
                BoltCheckout.configure(
                    json_cart,
                    $jsonHints,
                    {
                      check: function() {           
                        $check
                        $checkForAdmin
                        if (isEmptyQuote) {
                            alert('{$boltHelper->__('Your shopping cart is empty. Please add products to the cart.')}');
                            return false;
                        }
                        if (!json_cart.orderToken) {
                            alert(json_cart.error);
                            return false;
                        }
                        return true;
                      },
```

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -77,23 +71,42 @@ public function getTrackJsUrl()
...
+    private function _createBoltOrder($quote, $checkoutType)
     {
-        // Load the required helper class
         $boltHelper = Mage::helper('boltpay/api');
+        $isMultiPage = $checkoutType === 'multi-page';
 
         $items = $quote->getAllVisibleItems();
 
-        if (empty($items)) return json_decode('{"token" : ""}');
+        $has_admin_shipping = false;
```
johnmccombs1: Should we use camel case for this?

app/code/community/Bolt/Boltpay/Model/Observer.php
```diff
@@ -94,11 +94,21 @@ public function verifyOrderContents($observer)
...
                 $message .= "<br/>Bolt reports ".($transaction->amount->amount/100).'. Magento expects '.$magentoTotal/100;
 
-                // TODO: Properly adjust amount if it is off by only one cent
-                $order->setHoldBeforeState($order->getState());
-                $order->setHoldBeforeStatus($order->getStatus());
-                $order->setState(Mage_Sales_Model_Order::STATE_HOLDED, true, $message)
-                    ->save();
+                # Adjust amount if it is off by only one cent, likely due to rounding
+                $difference = $transaction->amount->amount - $magentoTotal;
+                if ( abs($difference) <= 1) {
```
daisy1754: I prefer to keep add one more condition to be explicit:

```
if ($difference == 0) {
  // no-op
} else if (abs($difference) == 1) {
  // adjust
} else {
  // on-hold
}
```

mccottry: @daisy1754 

Please check this code:
https://github.com/BoltApp/bolt-magento1/blob/92a5f7eeba9b7f2ee3bea9e16a54379d6afe6443/app/code/community/Bolt/Boltpay/Model/Observer.php#L90-L99

Do you want this even though the parent if block implicitly checks for this and therefore makes the case `if ($difference == 0)` to be impossible to ever be true?  I changed `if (abs($difference) == 1)` as you asked because the numbers should always be integers.

PR #105, opened 2018-07-02
Move clone quote to helper

daisy1754:


app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -181,85 +177,8 @@ public function getCartDataJs($checkoutType = 'multi-page')
                     foreach ($expiredQuotes as $expiredQuote) {
```
mccottry: This is wrong, and I believe that trying to break this down into many pull request is going to lead to many errors.  Expired quotes should not be deleted here.

daisy1754: @mccottry intention here is this PR only does refactoring - we keep existing broken behaviour as is. Fix will come with your main PR (or another PR).

PR #108, opened 2018-07-02
Update error code for hooks

daisy1754:


daisy1754:
ok let https://github.com/BoltApp/bolt-magento1/pull/68 take care of this

PR #113, opened 2018-07-05
Replace freegeoip code to ipstack & make prefetchEstimate work

dumega:
https://app.asana.com/0/436430054069245/733119829814643/f

### DIFF
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -44,7 +44,6 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
      */
     const AUTO_CAPTURE_ENABLED = 1;
 
-
     const CSS_SUFFIX = 'bolt-css-suffix';
 
     /**
@@ -663,8 +662,12 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
         $locationInfo = Mage::getSingleton('core/session')->getLocationInfo();
 
         if (empty($locationInfo)) {
-            $locationInfo = $this->url_get_contents("http://freegeoip.net/json/".$this->getIpAddress());
-            Mage::getSingleton('core/session')->setLocationInfo($locationInfo);
+            //To receive the API results in the old freegeoip format, we need ipstack Access Key
+            $ipstackAccessKey = Mage::helper('core')->decrypt(Mage::getStoreConfig('payment/boltpay/ipstack_key'));
+            if(!empty($ipstackAccessKey)){
+               $locationInfo = $this->url_get_contents("http://api.ipstack.com/".$this->getIpAddress()."?access_key=".$ipstackAccessKey."&output=json&legacy=1");
+               Mage::getSingleton('core/session')->setLocationInfo($locationInfo);
+            }
         }
 
         return $locationInfo;
--- a/app/code/community/Bolt/Boltpay/controllers/ShippingController.php
+++ b/app/code/community/Bolt/Boltpay/controllers/ShippingController.php
@@ -217,17 +217,17 @@ class Bolt_Boltpay_ShippingController extends Mage_Core_Controller_Front_Action
 
             // ----------^_^----------- //
             $shippingAddress = array(
-                'city'       => @$shippingAddressOriginal['city'],
-                'region'     => @$shippingAddressOriginal['region'],
-                'region_id'  => @$shippingAddressOriginal['region_id'] ? $shippingAddressOriginal['region_id'] : null,
-                'postcode'   => @$shippingAddressOriginal['postcode'],
-                'country_id' => @$shippingAddressOriginal['country_id'],
+                'city'       => @($shippingAddressOriginal['city']),
+                'region'     => @($shippingAddressOriginal['region']),
+                'region_id'  => @($shippingAddressOriginal['region_id']),
+                'postcode'   => @($shippingAddressOriginal['postcode']),
+                'country_id' => @($shippingAddressOriginal['country_id']),
             );
             unset($shippingAddressOriginal);
 
             $addressData = $this->mergeAddressData($geoLocationAddress, $shippingAddress);
 
-            if(@$addressData['postcode']) {
+            if(isset($addressData['postcode'])) {
                 $cacheIdentifier = $this->getPrefetchCacheIdentifier($quote, $addressData);
                 $this->saveAddressCache($addressData, $cacheIdentifier);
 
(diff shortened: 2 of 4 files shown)


app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -43,6 +43,11 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
      * @var int flag that represents if the capture is automatically done on authentication
      */
     const AUTO_CAPTURE_ENABLED = 1;
+    
+    /**
+     * @var string To receive the API results in the old freegeoip format, we need ipstack Access Key
+     */
+    const IPSTACK_ACCESS_KEY = '18beed19bdef35bf26f49f2e53c63842';
```
daisy1754: I'm little uncomfortable putting this in public github repo, since we're charged based on number of request limit. can you add config in admin for this?

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -663,8 +662,12 @@ function getLocationEstimate()
         $locationInfo = Mage::getSingleton('core/session')->getLocationInfo();
 
         if (empty($locationInfo)) {
-            $locationInfo = $this->url_get_contents("http://freegeoip.net/json/".$this->getIpAddress());
-            Mage::getSingleton('core/session')->setLocationInfo($locationInfo);
+            //To receive the API results in the old freegeoip format, we need ipstack Access Key
+            $ipstackAccessKey = Mage::helper('core')->decrypt(Mage::getStoreConfig('payment/boltpay/ipstack_key'));
```
YevhenShBolt: I advice to move it to the Helper/Data.php
Like - getPublishableKeyOnePage()

It provides us an ability to use it in different places if needed.

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -217,11 +217,11 @@ public function prefetchEstimateAction()
 
             // ----------^_^----------- //
             $shippingAddress = array(
-                'city'       => @$shippingAddressOriginal['city'],
-                'region'     => @$shippingAddressOriginal['region'],
+                'city'       => @$shippingAddressOriginal['city'] ? $shippingAddressOriginal['city'] : null,
```
daisy1754: I'm PHP noob. Isn't `@$shippingAddressOriginal['city']` return `null` if 'city' does not exist?

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
daisy1754: I feel like we only access this key from here

app/code/community/Bolt/Boltpay/etc/system.xml
```diff
@@ -102,6 +102,16 @@
               <show_in_website>1</show_in_website>
               <show_in_store>0</show_in_store>
             </signing_key>
+            
+            <ipstack_key translate="label">
+              <label>Ipstack Access Key</label>
+              <frontend_type>obscure</frontend_type>
```
daisy1754: Add comment:
> If provided, Bolt plugin performs address lookup to optimize the performance. Obtain API key via https://ipstack.com/ or contact Bolt

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -329,22 +329,30 @@ public function getGeoIpAddress()
...
+            $countryObj = Mage::getModel('directory/country')->loadByCode($addressData['country_id']);
+            $isRegionAvailable = ($countryObj->getRegionCollection()->getSize() > 0);
+    
+            if (!$isRegionAvailable) {
+                // If country does not have region options for dropdown.
+                $addressData['region'] = $addressData['region_name'];
+            }
+            elseif(!empty($addressData['region'])){
+                $regionModel = Mage::getModel('directory/region')->loadByCode($addressData['region'], $addressData['country_id']);
+                if($regionModel){
```
daisy1754: nice!

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: I tend to lean more toward what @daisy1754 has said. 

 It is a matter of taste.  One may view a helper method as a simplification and clearer, reusable representation of the action which is being performed, and another may view it as code clutter, preferring to make simple one line call.

However, if reused, the extra decryption process does justify the helper method.

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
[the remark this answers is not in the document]
dumega: @daisy1754  @mccottry  Personally, I also would like to change it to the style `'city'       => isset($shippingAddressOriginal['city']) ? $shippingAddressOriginal['city'] : null,` , it is much readable and error suppression is not a good idea in such an array
```
Error suppression should be avoided if possible as it doesn't just suppress the error that you are trying to stop, but will also suppress errors that you didn't predict would ever occur. This will make debugging a nightmare.
```

also we can not set it to empty string here, cause if so, when we refresh the cart page, the region id would become 0, not empty anymore, that would cause issue.

dumega: @daisy1754 yes, you are right if would return null if `$shippingAddressOriginal['city']` does not exsit when try to throw exception. and I would like to make it more readable in the new commit https://github.com/BoltApp/bolt-magento1/pull/113/commits/d20d831e4561142c5aef605e58f992772f9b47c1

daisy1754: To my understanding, do you  want to replace
> 'city'       => @$shippingAddressOriginal['city'],

with 

> 'city'       => isset($shippingAddressOriginal['city']) ? $shippingAddressOriginal['city'] : null,

for better readability? 

I think you should assume whoever works on this repo has decent PHP knowledge, and with that, IMO `@$shippingAddressOriginal['city']` is more readable.

app/design/frontend/base/default/template/boltpay/prefetch_shipping_quote.phtml
```diff
@@ -32,21 +32,33 @@ if($this->isBoltActive()) {
...
+                            evt.initEvent('change', true, true);
+                            region_id.value = _data.region_id;
+                            region_id.dispatchEvent(evt);
+                        }
 
-                        if (region && !region.value.length) {
+                        if (region && !region.value.length && _data.region != null) {
                             if (region.tagName.toLowerCase() === 'input') {
-                                region.value = _data.region_name;
+                                region.value = _data.region;
```
daisy1754: wanted to double check if this is intentional. I'd assume if it's text input user would enter region name, not a region code?

dumega: This is a hidden field, not for user input

dumega: and actually _data.region is region name in our script

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -329,22 +329,30 @@ public function getGeoIpAddress()
...
+                    $addressData['region'] = $regionModel->getName();
+                    $addressData['region_id'] = $regionModel->getId();
```
dumega: we get the address info from ipstack, and in the response, it has two fields for region `region_code` and `region_name` (eg. for California, the region_code is CA and region_name is California). But in magento 1, when calculate the shipping cost, it does not have such a field for code, and instead it has another two field for region, `region` (actually the name) and `region_id` (internal ID). then if I try to convert the response from ipstack into mangeto 1 format

PR #117, opened 2018-07-11
Initial test button configuration.

johnmccombs1:
Here is the task description:

Currently we have settings in magento admin. System->Configuration->Payment Method->Bolt Pay

Add "Check configuration" here to run following checks:
 - If API key, signing secret, and at least one publishable key is set
 - Make API request to merchant/order API with fake data. Show error if response is not 2xx
 - Verify db has all schema defined in https://github.com/BoltApp/bolt-magento1/tree/develop/app/code/community/Bolt/Boltpay/sql/bolt_boltpay_setup_

### DIFF
--- a/app/code/community/Bolt/Boltpay/Helper/Api.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Api.php
@@ -542,7 +542,7 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
      */
     public function getApiUrl()
     {
-        return Mage::getStoreConfig('payment/boltpay/test') ?
+        return Mage::getStoreConfigFlag('payment/boltpay/test') ?
             self::API_URL_TEST :
             self::API_URL_PROD;
     }
--- a/app/code/community/Bolt/Boltpay/Helper/Api.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Api.php
@@ -201,7 +201,7 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
                 throw new Exception("The quote ". $immutableQuote->getParentQuoteId() ." is currently being processed or has been processed.");
             } else {
                 $parentQuote->setIsActive(false)->save();
-            }          
+            }
 
             // adding guest user email to order
             if (!$immutableQuote->getCustomerEmail()) {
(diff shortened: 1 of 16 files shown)


app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -372,10 +372,11 @@ protected function validateSubmittedOrder($order, $quote) {
...
      * @param string $object   defines part of endpoint url which is normally/always??? set to merchant
      * @param string $type     Defines the endpoint type (i.e. order|transactions|sign) that is used as part of the url
+     * @param null $storeId
      * @throws  Mage_Core_Exception  thrown if an error is detected in a response
      * @return mixed           Object derived from Json got as a response
      */
-    public function transmit($command, $data, $object='merchant', $type='transactions')
+    public function transmit($command, $data, $object='merchant', $type='transactions', $storeId = null)
     {
         $url = $this->getApiUrl() . 'v1/';
```
daisy1754: do you have to pass storeId here?

app/code/community/Bolt/Boltpay/controllers/ConfigurationController.php
```diff
@@ -0,0 +1,216 @@
...
+        );
+        curl_setopt($ch, CURLOPT_HTTPHEADER, $headerInfo);
+        curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+        curl_setopt($ch, CURLOPT_HEADER, true);
+
+        curl_exec($ch);
+        $response = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+        curl_close($ch);
+
+        return $response == 200;
```
daisy1754: can you change this to something like int($response/100) == 2?
We may use different response code like 201 created in the future

daisy1754:
One more thing - can you bugsnag all failures?
If this button shows config is incorrect, next action item for merchant is probably contacting us. We should be able to know details of what went wrong.

app/code/community/Bolt/Boltpay/Helper/Api.php
johnmccombs1: This is the cleanest way to apply the store lookup. If it's not done this way it will require some ugly code to pull it.

johnmccombs1:
@daisy1754 I updated code based on your suggestions. Please review.

PR #118, opened 2018-07-13
Catch and report post success forward errors

mccottry:
After the order is saved to Bolt and Magento, the browser is then forwarded to the success page.  Before this, Magento triggers a `checkout_onepage_controller_success_action` event which several plugins may hook into and in some cases fail.

Instead of allowing the program to crash, for Bolt orders, we catch the exception and report it to Bugsnag.

### DIFF
--- a/app/code/community/Bolt/Boltpay/controllers/OnepageController.php
+++ b/app/code/community/Bolt/Boltpay/controllers/OnepageController.php
@@ -14,79 +14,43 @@
  * @copyright  Copyright (c) 2018 Bolt Financial, Inc (https://www.bolt.com)
  * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-
 require_once 'Mage/Checkout/controllers/OnepageController.php';
 
 /**
  * Class Bolt_Boltpay_OnepageController
  *
- * @deprecated This will be removed after the skip payment option is resolved.
  */
 class Bolt_Boltpay_OnepageController extends Mage_Checkout_OnepageController
 {
-    public function saveShippingMethodAction() 
+
+    /**
+     * Order success action
+     */
+    public function successAction()
     {
-        try {
-            if ($this->_expireAjax()) {
-                return;
-            }
-
-            if ($this->getRequest()->isPost()) {
-                $data = $this->getRequest()->getPost('shipping_method', '');
-                $result = $this->getOnepage()->saveShippingMethod($data);
-                // $result will contain error data if shipping method is empty
-                if (!$result) {
-                    if (Mage::getStoreConfig('payment/boltpay/skip_payment')) {
-                        try {
-                            $data = array('method' => Bolt_Boltpay_Model_Payment::METHOD_CODE);
-                            $result = $this->getOnepage()->savePayment($data);
-                        } catch (Mage_Payment_Exception $e) {
-                            if ($e->getFields()) {
-                                $result['fields'] = $e->getFields();
-                            }
-
-                            $result['error'] = $e->getMessage();
-                            Mage::helper('boltpay/bugsnag')->notifyException($e);
-                        } catch (Mage_Core_Exception $e) {
-                            $result['error'] = $e->getMessage();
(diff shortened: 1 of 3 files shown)


app/code/community/Bolt/Boltpay/controllers/OnepageController.php
```diff
@@ -14,79 +14,47 @@
...
-                $this->getResponse()->setBody(Mage::helper('core')->jsonEncode($result));
-            }
-        } catch (Exception $e) {
+        try {
+            Mage::dispatchEvent('checkout_onepage_controller_success_action', array('order_ids' => array($lastOrderId)));
+        } catch ( Exception $e ) {
             Mage::helper('boltpay/bugsnag')->notifyException($e);
-            throw $e;
+            
+            if (strtolower($this->_getOrder()->getPayment()->getMethod()) !== Bolt_Boltpay_Model_Payment::METHOD_CODE) {
```
daisy1754: hmm is it safe to swallow error? can you give me an example of case we want to ignore exception?

dumega: I think we would better let them go. At this point the bolt plugin actually already complete its task and order is alread successfully placed, and even there is any exception from other plugins, we would better let them handle in their own way.

mccottry: @daisy1754 
>At this point the bolt plugin actually already complete its task 
>and the order is already successfully placed

The only thing left for Bolt to do on Bolt orders is to show the order success screen to the end user.  Any error that comes up at this point throws an error screen to the end user, giving Bolt the appearance of being buggy and not succeeding.  

A real case for this is Indique Dev and possibly Indique live where certain plugins are performing post order save task that seeks conditions for their payment methods regardless of other payment methods being used.

In our code, for Bolt, we let our plugin succeed. For other payment methods or
> other plugins, we ... let them handle in their own way

daisy1754:
I'm now a bit concerned about overriding core class. There are a few tricky questions like:
 - what if different version of magento have different implementation of OnepageController
 - what if other plugin also overrides OnepageController

Maybe we should keep this as merchant-specific customization. Thought?

mccottry:
@daisy1754 
*Maybe we should keep this as merchant-specific customization. Thought?*
The place where this became an issue, Indique, has this implemented as *local* customization, so, yes, a local customization is always a good practice.

*What if different version of magento have different implementation of OnepageController?*
A few observations:
- The class that we override should always subclass the core file, preserving the overall logic.  When appropriate or whenever possible, the class should invoke the parent method.
-   In the case above, the logic is very limited in scope.  It is simply to clear, the session, notify observers that the success page is being forwarded to, and forwarding to the said page, so, there is not much possibility important code to be happening here outside of honoring those task.
- Most or all of our Magento clients are on a 1.9.x version.  1.91 - 1.9.3 seem to be the same. There will be no higher version than 1.9.3.x
- If we were to override this controller, *the ideal solution is to set up automated integration and/or unit test in all the targeted Magento versions.*  Alternatively, because of the limited scope of this, we could look to see the source code in many version of Magento for this method.  

**What if other plugins also overrides OnepageController?**
In this case, it's a question of who wants it more, or who wants it first.  A first class plugin may assume that it is authoritative and it should be the one that other plugins conform to, like FireCheckout, and in the case of Indique, `LucidPath_Salesrep_OnepageController`.  In the ideal future world, people have Bolt first and extend these plugins work with us.  But, in the real world, we would quickly identify this in a standard integration.  Either we are forwarding to the success page, or we are not.  If not, if the reason is because the override did not take place, we are in the same position that we would be in otherwise, which is implementing this as a local extension.

--------------------

All this said, this is not a critical update.  It is good for the purpose of drawing this to the attention of others.  If you chose, closing this PR and addressing this in an issue I brought up may be just as good of a solution.

https://app.asana.com/0/436430054069245/749651341671997

daisy1754:
okay let's keep it as merchant-specific customization unless this comes up a lot

PR #120, opened 2018-07-17
Looks up order by order reference

mccottry:
Looks up order by `order_reference` which is much more reliable than `increment_id/display_id`


mccottry:
This was refactored in https://github.com/BoltApp/bolt-magento1/pull/123

It is presented here to simplify by separating the logic review from the refactor.

PR #121, opened 2018-07-17
Generate new increment id if detected as already in use

mccottry:
Several times we have observed that the reserved order id (increment id) can be used even though a quote has "reserved" it.  This is also implied as an expectation in core Magento code in class `Mage_Sales_Model_Quote`

```
    /**
     * Generate new increment order id and associate it with current quote
     *
     * @return Mage_Sales_Model_Quote
     */
    public function reserveOrderId()
    {
        if (!$this->getReservedOrderId()) {
            $this->setReservedOrderId($this->_getResource()->getReservedOrderId($this));
        } else {
            //checking if reserved order id was already used for some order
            //if yes reserving new one if not using old one
            if ($this->_getResource()->isOrderIncrementIdUsed($this->getReservedOrderId())) {
                $this->setReservedOrderId($this->_getResource()->getReservedOrderId($this));
            }
        }
        return $this;
    }
```

This fix is needed for SwissGear and would benefit the Magento 1 plugin on the whole.

***NOTE:***
_The complete solution calls for a simple, and reliable way to reset display_id on Bolt.
If it requires not setting the display id until we are sure of what it will be, then we should 
do this in the Magento 1 code, and enable it in the Bolt side code for Magento 1_

**_These means that my proposal is to remove sending display_id from the Bolt order creation routine, then to send it to Bolt in the response of the first webhook call, and finally have this feature enabled in Magento 1 like in all other plugins._**

### DIFF
--- a/app/code/community/Bolt/Boltpay/Helper/Api.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Api.php
@@ -274,11 +274,12 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
             $preExistingOrder = Mage::getModel('sales/order')->loadByIncrementId($parentQuote->getReservedOrderId());
 
             ############################
-            # First check if this order matches the transaction and therefore duplicated.
+            # First check if this order matches the transaction and therefore already created
             ############################
             $preExistingTransactionReference = $preExistingOrder->getPayment()->getAdditionalInformation('bolt_reference');
             if ( $preExistingTransactionReference === $reference ) {
-                throw new Exception("The order #".$preExistingOrder->getIncrementId()." has already been processed for this quote." );
+                Mage::helper('boltpay/bugsnag')->notifyException( new Exception("The order #".$preExistingOrder->getIncrementId()." has already been processed for this quote." ), array(), 'warning' );
+                return $preExistingOrder;
             }
             ############################
 
--- a/app/code/community/Bolt/Boltpay/Helper/Api.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Api.php
@@ -267,6 +267,22 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
 
             Mage::helper('boltpay')->collectTotals($immutableQuote, true)->save();
 
+            ////////////////////////////////////////////////////////////////////////////
+            // reset increment id if needed
+            ////////////////////////////////////////////////////////////////////////////
+            /* @var Mage_Sales_Model_Order $preExistingOrder */
+            $preExistingOrder = Mage::getModel('sales/order')->loadByIncrementId($parentQuote->getReservedOrderId());
+
+            if (!$preExistingOrder->isObjectNew()) {
+                $parentQuote
+                    ->setReservedOrderId(null)
+                    ->reserveOrderId()
+                    ->save();
+
+                $immutableQuote->setReservedOrderId($parentQuote->getReservedOrderId());
+            }
+            ////////////////////////////////////////////////////////////////////////////
+
             // a call to internal Magento service for order creation
             $service = Mage::getModel('sales/service_quote', $immutableQuote);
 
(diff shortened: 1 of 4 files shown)


app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -267,6 +267,22 @@ public function createOrder($reference, $sessionQuoteId = null, $isAjaxRequest =
 
             Mage::helper('boltpay')->collectTotals($immutableQuote, true)->save();
 
+            ////////////////////////////////////////////////////////////////////////////
+            // reset increment id if needed
+            ////////////////////////////////////////////////////////////////////////////
+            /* @var Mage_Sales_Model_Order $preExistingOrder */
+            $preExistingOrder = Mage::getModel('sales/order')->loadByIncrementId($parentQuote->getReservedOrderId());
+
+            if (!$preExistingOrder->isObjectNew()) {
```
daisy1754: can we add check if this preExistingOrder is not tied to Bolt transaction reference we're looking at now?

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -267,6 +267,22 @@ public function createOrder($reference, $sessionQuoteId = null, $isAjaxRequest =
...
+                $parentQuote
+                    ->setReservedOrderId(null)
+                    ->reserveOrderId()
+                    ->save();
+
+                $immutableQuote->setReservedOrderId($parentQuote->getReservedOrderId());
+            }
+            ////////////////////////////////////////////////////////////////////////////
+
```
danac-gs: I may be wrong, but it seems to me that the following two lines do the same as this block.

$parentQuote->reserveOrderId()->save(); // reset increment id is implemented as a part of reserveOrderId()
$immutableQuote->setReservedOrderId($parentQuote->getReservedOrderId());

mccottry: @danac-gs I would say that you are right, in theory.  That is certainly how I first implemented the code after looking at the above implementation of the `reserveOrderId` method.

In practice, however, at least on Swissgear, calling this method continued to hang onto reserved order id, most likely because they have over-ridden this for their special increment IDs that all start with SWG.

Swissgear aside, because we have already incurred the expense of doing the lookup for the order id and already know that it needs to be updated, you can think of the `->setReservedOrderId(null)` as a `-force` flag saying that we are not interested in checking if the currently assigned reserved order ID exists, we trust our logic that makes our plugin work and want a new one no matter what.

app/code/community/Bolt/Boltpay/Helper/Api.php
mccottry: @daisy1754 

This check is indirectly performed here:
https://github.com/BoltApp/bolt-magento1/blob/a9fe2521362b8dc3df896d5cf1063ec73d253053/app/code/community/Bolt/Boltpay/Helper/Api.php#L192-L195

And I am adding the additional check to the branch.

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -267,6 +267,31 @@ public function createOrder($reference, $sessionQuoteId = null, $isAjaxRequest =
...
+            ////////////////////////////////////////////////////////////////////////////
+            /* @var Mage_Sales_Model_Order $preExistingOrder */
+            $preExistingOrder = Mage::getModel('sales/order')->loadByIncrementId($parentQuote->getReservedOrderId());
+
+            ############################
+            # First check if this order matches the transaction and therefore duplicated.
+            ############################
+            $preExistingTransactionReference = $preExistingOrder->getPayment()->getAdditionalInformation('bolt_reference');
+            if ( $preExistingTransactionReference === $reference ) {
+                throw new Exception("The order #".$preExistingOrder->getIncrementId()." has already been processed for this quote." );
```
daisy1754: Is this error? how about we bugsnag and return preExistingOrder?

daisy1754:
actually I'm gonna merge this to unblock other merchants

PR #122, opened 2018-07-17
Reset in use increment id complete solution

mccottry:
Logic for removing display id from Bolt order creation and adding it to the first successful hook call per the spec defined at https://docs.bolt.com/v1/docs/hook-integration-1 under **Hook Responses**


app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -635,7 +636,6 @@ public function buildCart($quote, $items, $multipage)
         $productMediaConfig = Mage::getModel('catalog/product_media_config');
         $cartSubmissionData = array(
             'order_reference' => $quote->getId(),
-            'display_id'      => $quote->getReservedOrderId(),
```
daisy1754: Let's keep this. 99% of case it'll be accurate. If not, it can be updated via hook.

mccottry: I love this comment if it means the hook for updating this is on its way!!!!

PR #129, opened 2018-07-19
Admin order updates

mccottry:
There are a couple of issues addressed here with the admin.
The first major issue was with errors caused by multiple loads of connect.js.  It was ok before, but now it is not.  The fix was simply to load connect.js once and reload the Bolt Button as many times as needed.  *This may also provide a solution for switching between multi-step and one-page keys.*ly

Secondly, preloading the address fields has been fixed.


YevhenShBolt:
I agree with Daniel.

PR #131, opened 2018-07-22
Parent id as order reference to enable bolt side duplicate order detection

mccottry:
https://app.asana.com/0/544708310157130/752340529245956

Using the parent ID as the order_reference in Bolt order creation allows for the Bolt backend to provide for integrity constraints, enforcing a single order creation per session.

This addresses the problem of double billing the customer for the same order (i.e. one order created in Magento, while two transactions created at Bolt.)

This has already been implemented in Magento 2

### DIFF
--- a/app/code/community/Bolt/Boltpay/Helper/Api.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Api.php
@@ -161,9 +161,8 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
             }
 
             $transaction = $transaction ?: $this->fetchTransaction($reference);
-            $transactionStatus = $transaction->status;
 
-            $immutableQuoteId = $transaction->order->cart->order_reference;
+            $immutableQuoteId = $this->getImmutableQuoteIdFromTransaction($transaction);
 
             /* @var Mage_Sales_Model_Quote $immutableQuote */
             $immutableQuote = Mage::getModel('sales/quote')->loadByIdWithoutStore($immutableQuoteId);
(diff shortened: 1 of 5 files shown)


app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -1103,4 +1111,51 @@ public function getOrderByQuoteId($quoteId) {
...
+     * @param object $transaction  The Bolt transaction as a php object
+     *
+     * @return string  The immutable quote id
+     */
+    public function getImmutableQuoteIdFromTransaction( $transaction ) {
+        if (strpos($transaction->order->cart->display_id, '|')) {
+            return explode("|", $transaction->order->cart->display_id)[1];
+        } else {
+            /////////////////////////////////////////////////////////////////
+            // here we deal with older versus newer version
```
daisy1754: Newer version is handled by above `return explode("|", $transaction->order->cart->display_id)[1];`, isn't it?

mccottry: @daisy1754 correct.

mccottry:
@johnmccombs1 Actually, the concept and design are very simple and we want to keep it that way.

Perhaps we can have a meeting to discuss this and I will make some diagrams for you to have a better understanding.

Most importantly, all elements of this are essential and address some huge issues that were previously left unsolved, and discussing and understanding these issues will also bring clarity to what these changes are and why they exist.

Also, I believe Magento 2 and WooCommerce follow a very similar design pattern, so, understanding this will give understanding to all those code bases as well.

app/code/community/Bolt/Boltpay/Helper/Api.php
mccottry: @daisy1754 When you push this code into production on pre-existing merchants, there will be several orders created that use the old format.  If there is no support for the legacy format, all webhooks for these orders would fail.  A good example would be payments that are in review that will update themselves or issuing a refund or canceling an order from the merchant dashboard.   This was pointed out to me by @danac-gs who also expressed a need for this addition and is the author of this excellent idea.

So, I will change the wording in this comment to 
`here we deal with the legacy hook format for backward compatibility`, which definitely expresses this more clearly.

I also already give a brief explanation of this in the methods comments:

https://github.com/BoltApp/bolt-magento1/blob/2cf827d0485710707830c478511468cf6df268ce/app/code/community/Bolt/Boltpay/Helper/Api.php#L1115-L1124

PR #133, opened 2018-07-24
Admin order webhook and other payment method support

mccottry:
This branch fixes two errors in the admin order creation.

1.) It acknowledges other payment method types and differs to Magento to process these other methods
2.) It uses the immutable quote/parent quote construct in order for webhooks to succeed on admin created Bolt orders

additionally, we assure that the admin is notified to calculate shipping in all scenarios.

### DIFF
--- a/app/code/community/Bolt/Boltpay/controllers/Adminhtml/Sales/Order/CreateController.php
+++ b/app/code/community/Bolt/Boltpay/controllers/Adminhtml/Sales/Order/CreateController.php
@@ -78,8 +78,43 @@ class Bolt_Boltpay_Adminhtml_Sales_Order_CreateController extends Mage_Adminhtml
     public function saveAction()
     {
         $boltReference = $this->getRequest()->getPost('bolt_reference');
+
+        /////////////////////////////////////////////////////////////////////////////
+        // If there is no bolt reference, then it indicates this is another payment
+        // method.  In this case, we differ to Magento to handle this after assuring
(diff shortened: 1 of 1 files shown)


app/code/community/Bolt/Boltpay/controllers/Adminhtml/Sales/Order/CreateController.php
```diff
@@ -77,9 +77,46 @@ public function loadBlockAction()
...
+        // We must assure that Magento knows to recalculate the shipping
+        $_POST['collect_shipping_rates'] = 1;
+
+        /////////////////////////////////////////////////////////////////////////////
+        // If there is no bolt reference, then it indicates this is another payment
+        // method.  In this case, we differ to Magento to handle this
+        /////////////////////////////////////////////////////////////////////////////
         $boltReference = $this->getRequest()->getPost('bolt_reference');
+
+        if (!$boltReference) {
```
dumega: Why not bring this condition to the beginning of the function? Then if this is another payment method we just return immediately.

daisy1754: +1

mccottry: @dumega @daisy1754 
Because the things before this are relevant to all shipping method.

PR #134, opened 2018-07-25
Admin order webhook and other payment method support refactor refine

mccottry:
- Refactored and expanded data normalization code into an easily overridable method.
- Made it even more clear the intentional nature of the normalization code being executed prior to deferring control to Magento for all non-Bolt orders
- Explicitly wrote out a call to data normalization code after the immutable quote is stored to the admin session for clarity
- Added more comments to make this clear within the code, inline

### DIFF
--- a/app/code/community/Bolt/Boltpay/controllers/Adminhtml/Sales/Order/CreateController.php
+++ b/app/code/community/Bolt/Boltpay/controllers/Adminhtml/Sales/Order/CreateController.php
@@ -80,31 +80,34 @@ class Bolt_Boltpay_Adminhtml_Sales_Order_CreateController extends Mage_Adminhtml
     {
 
         /////////////////////////////////////////////////////////////////////////////
+        // Case 1:
         // If there is no bolt reference, then it indicates this is another payment
         // method.  In this case, we differ to Magento to handle this
         /////////////////////////////////////////////////////////////////////////////
         $boltReference = $this->getRequest()->getPost('bolt_reference');
         if (!$boltReference) {
-            $this->_normalizeOrderData();
+            $this->_normalizeOrderData();  // We must re-normalize the data first
             parent::saveAction();
             return;
-        } else {
-            ///////////////////////////////////////////////////
-            /// We must use the immutable quote to create
-            /// this order for subsequent webhooks to succeed.
-            ///////////////////////////////////////////////////
-            /** @var Bolt_Boltpay_Helper_Api $boltHelper */
-            $boltHelper = Mage::helper('boltpay/api');
-            $transaction = $boltHelper->fetchTransaction($boltReference);
-
-            $immutableQuoteId = $boltHelper->getImmutableQuoteIdFromTransaction($transaction);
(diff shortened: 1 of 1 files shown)


app/code/community/Bolt/Boltpay/controllers/Adminhtml/Sales/Order/CreateController.php
```diff
@@ -72,28 +72,36 @@ public function loadBlockAction()
         parent::loadBlockAction();
     }
 
+
     /**
      * Saving quote and create order.  We add the Bolt reference to the session
      */
     public function saveAction()
     {
-        // some versions of Magento store the shipping method at the top level of the $_POST array
```
daisy1754: I don't think this removal is intentional?

app/code/community/Bolt/Boltpay/controllers/Adminhtml/Sales/Order/CreateController.php
```diff
@@ -72,28 +72,36 @@ public function loadBlockAction()
...
         // If there is no bolt reference, then it indicates this is another payment
         // method.  In this case, we differ to Magento to handle this
         /////////////////////////////////////////////////////////////////////////////
         $boltReference = $this->getRequest()->getPost('bolt_reference');
-
         if (!$boltReference) {
+            $this->_normalizeOrderData();
             parent::saveAction();
             return;
+        } else {
```
daisy1754: no need for "else" because you return in if branch above.

app/code/community/Bolt/Boltpay/controllers/Adminhtml/Sales/Order/CreateController.php
mccottry: @daisy1754  yes. it was. please review:

https://github.com/BoltApp/bolt-magento1/blob/aed618e1498d86db325b3d56199f45e2a1617088/app/code/community/Bolt/Boltpay/controllers/Adminhtml/Sales/Order/CreateController.php#L191-L234

app/code/community/Bolt/Boltpay/controllers/Adminhtml/Sales/Order/CreateController.php
mccottry: @daisy1754 Thank you.  I agree.  This was specifically put in for "easier reading" to highlight the intentional nature of the two separate cases because this was called into question in the parent branch.

Had it not been the case, I would have left it as it was in the parent branch where the understanding of data normalization for two cases was to be implicitly understood.  This is also why in the subject of this PR, I have written:
```
Made it even more clear the intentional nature of the normalization code being executed prior to deferring control to Magento for all non-Bolt orders
Explicitly wrote out a call to data normalization code after the immutable quote is stored to the admin session for clarity
```

However, since it's intentional nature, I think at this point, is very clear, it's definitely very ok to remove the else statement.  

**Most importantly, the pull request is now a request to be merged into `develop`, and not the original parent branch.**

PR #135, opened 2018-07-25
Enables individual store configuration

mccottry:
https://app.asana.com/0/436430054069245/756982959124649/f


YevhenShBolt:
It is a good idea but needs to be sure that it works bolt in different stores. I am not sure that all our config settings in helper can properly check the store.

PR #143, opened 2018-07-27
Bug fix shipping discounts

johnmccombs1:
Fixed bug in calculating shipping discounts.

### DIFF
--- a/app/code/community/Bolt/Boltpay/Helper/Api.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Api.php
@@ -931,7 +931,7 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
         $shippingAddress = $quote->getShippingAddress();
         $shippingAddress->setCollectShippingRates(true)->collectShippingRates()->save();
 
-        $originalDiscountedPrice = $quote->getSubtotalWithDiscount();
+        $originalDiscountedSubtotal = $quote->getSubtotalWithDiscount();
 
         $rates = $this->getSortedShippingRates($shippingAddress);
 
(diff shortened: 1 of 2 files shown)


daisy1754:
Is it possible to write a unit test for this? Since it's been rewritten a few times recently, it'd be good if we can have test to regression

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -964,13 +964,13 @@ public function getShippingAndTaxEstimate( $quote )
                 );
             }
 
-            $newDiscountedPrice = $quote->getSubtotalWithDiscount();
+            $adjustedShippingAmount = $this->getAdjustedShippingAmount($originalDiscountedSubtotal, $quote);
 
             $option = array(
                 "service" => $label,
                 "reference" => $rateCode,
-                "cost" => round(($quote->getShippingAddress()->getShippingAmount() + ($originalDiscountedPrice - $newDiscountedPrice)) * 100),
```
mccottry: @johnmccombs1  Can you please review this code and see what the effect of changing the following line will yield?

from:
```
 "cost" => round(($quote->getShippingAddress()->getShippingAmount() + ($originalDiscountedPrice - $newDiscountedPrice)) * 100),
```

to:
```
 "cost" => round(($quote->getShippingAddress()->getShippingAmount() + ($newDiscountedPrice - $originalDiscountedPrice)) * 100),
```

Above, I switch the order of $newDiscountedPrice and $originalDiscountedPrice, which appears to be the source of the logical error.

Perhaps, this one change can be substituted for all the other code.

mccottry:
@johnmccombs1 I think that switching the order of subtraction in line 972 of the original code and adding the unit test that @daisy1754 suggested could be sufficient for fixing this error.

app/code/community/Bolt/Boltpay/Helper/Api.php
johnmccombs1: I've tested this and you are right. Not sure how I didn't see that as a fix before. Great find! I'll remove my code and add a unit test and call it done. Thanks!

mccottry: @johnmccombs1 LGTM

PR #144, opened 2018-07-29
Correct the way of handling the duplicated webhook exception

dumega:
Now if the parent quote is currently being processed or has been processed, it would throw an exception, but in the catch block the script would setIsActive to true again, this logic is wrong. so I create a new Exception class to handle this in a separate way, and @daisy1754 do you think there is a response code to tell the bolt server not to retry again?


app/code/community/Bolt/Boltpay/controllers/ApiController.php
```diff
@@ -162,6 +162,11 @@ public function hookAction()
             );
             $this->getResponse()->setHttpResponseCode(201);
 
+        } catch (Bolt_Boltpay_DuplicatedTransitionException $boltPayDuplicatedTransitionException) {
+            //In this case, The parent quote is currently being processed or has been processed,
+            //and this duplicated transaction should be ignored
+            Mage::helper('boltpay/bugsnag')->notifyException($boltPayDuplicatedTransitionException);
+            $this->getResponse()->setHttpResponseCode(200);
```
daisy1754: I have a mixed feeling here. I'd say let's return error until we can be certain this only caused by harmless race condition

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -195,7 +195,7 @@ public function createOrder($reference, $sessionQuoteId = null, $isAjaxRequest =
             if ($parentQuote->isEmpty() ) {
                 throw new Exception("The parent quote ". $immutableQuote->getParentQuoteId() ." is unexpectedly missing.");
             } else if (!$parentQuote->getIsActive() ) {
-                throw new Exception("The parent quote ". $immutableQuote->getParentQuoteId() ." is currently being processed or has been processed.");
+                throw new Bolt_Boltpay_DuplicatedTransitionException("The parent quote ". $immutableQuote->getParentQuoteId() ." is currently being processed or has been processed.");
```
daisy1754: can we somehow check if this quote is processed with what bolt transaction reference?

app/code/community/Bolt/Boltpay/controllers/ApiController.php
dumega: Yes, that's what I want to discuss, in this situation, a previous webhook already started to create order, and it can determine whether success or failure then response properly, so this duplicated transition should be ignored, its response actually is useless, does bolt server support such a response code?

daisy1754: well so if you can determine that hook is called with Bolt transaction reference that is already processed, you can return 200 here. 
However, if for some reason Bolt has two different transaction reference for the single order_reference (quote_id), we're in trouble and in that case plugin should return error so we can get notified and investigate.

dumega: Done, I separate the ways to handle different situations

app/code/community/Bolt/Boltpay/controllers/ApiController.php
```diff
@@ -162,6 +162,25 @@ public function hookAction()
...
 
+        } catch (Bolt_Boltpay_DuplicatedTransitionException $boltPayDuplicatedTransitionException) {
+            //In this case, the parent quote is currently being processed or has been processed,
+            //and if this duplicated transaction has the same reference with the processed transaction, then it should be ignored
+            //and if not, there should report an error
+            if($boltPayDuplicatedTransitionException->getProcessedBoltReference() == $reference){
+                Mage::helper('boltpay/bugsnag')->notifyException( new Exception($e->getMessage()) );
+                $this->getResponse()->setHttpResponseCode(200);
+            }
+            else{
```
mccottry: @danac-gs, @daisy1754, @dumega --
Since,
- we are using the parent quote ID as the order reference
- and Bolt server-side has a 1-to-1 constraint on transaction reference and order reference

Then this case seems to be impossible to reach.  What am I missing here?  It is definitely good to have this case handled in the plugin code, but, if this is reached, it implies a bigger problem.  We would only expect the case of the transaction references being the same in the standard "racing condition scenario".

Second, @dumega I don't think that anything is set to `on-hold` here, which is a Magento order state.  I think the hook is simply failing and no merchant side actions are being taken, including order creation or any order state change.

Possibly, it is better to give an error message that identifies the transaction and order which are already in the Magento system which render this current transaction as erroneous and redundant.

app/code/community/Bolt/Boltpay/controllers/ApiController.php
```diff
@@ -162,6 +162,25 @@ public function hookAction()
             );
             $this->getResponse()->setHttpResponseCode(201);
 
+        } catch (Bolt_Boltpay_DuplicatedTransitionException $boltPayDuplicatedTransitionException) {
+            //In this case, the parent quote is currently being processed or has been processed,
+            //and if this duplicated transaction has the same reference with the processed transaction, then it should be ignored
+            //and if not, there should report an error
+            if($boltPayDuplicatedTransitionException->getProcessedBoltReference() == $reference){
+                Mage::helper('boltpay/bugsnag')->notifyException( new Exception($e->getMessage()) );
+                $this->getResponse()->setHttpResponseCode(200);
```
mccottry: @daisy1754  Is it guaranteed that the first hook coming at this point is going to be an `auth` hook?

@dumega   What happens if the hook that arrives is not an `auth` and instead is a `pending` or a `irreversibly_rejected` hook, over even if it is a payment hook?  If any of these are possibilities, then passing back a 200 OK response, then you are introducing several errors into the code.

I think there are several questions to be answered here before introducing a change like this.

dumega: @mccottry yes, you are right, if another type of webhook arrive when the parent quote is currently being processed or has been processed, the logic inside the catch block can not handle it correctly, the situation would become more complicated.

so @daisy1754 , plz let us know if `If auth is not guaranteed to be the first hook type`

app/code/community/Bolt/Boltpay/controllers/ApiController.php
dumega: @mccottry the `on-hold` in error message is trying to explain there is no need to send duplicated webhook too soon, not a status signal. and if you look into it, the message already include related reference info.

`"The parent quote ". $immutableQuote->getParentQuoteId() ." is currently being processed or has been processed by bolt transaction ".$parentQuoteBoltReference."."`

and

`$e->getMessage().' Then this redundant bolt transaction '.$reference.' is set to on-hold'

app/code/community/Bolt/Boltpay/controllers/ApiController.php
daisy1754: confirming auth is not guaranteed to be the first hook type

dumega: @daisy1754 @mccottry I review the code again, so how about this?

When the hook that arrives belong to the same transaction, we just check its type, and if it is same as the current hook that being process the quote, I think it is fine to response 200 (or it would be great if bolt provide another special code for this situation) which actually ignore this hook. And if it is a different type, we response an error code and bolt server will retry later. 

So in my opinion, `auth`, `pending` or `irreversibly_rejected` or any other type, no matter which type come first, at that moment the main purpose is to create the magento order and make sure only to create one unique magento order, and in the observer.php the function `setInitialOrderStatusAndDetails` would handle this and set the proper state for the related hook.

After that, whatever type of hook arrive, we have $_validStateTransitions defined to guarantee the proper sequence.

And then, if the hook arrives belong to another different transaction, well, it already happened, and for now we do not know how this happened, so we just need to log the related info for further debug and investigation.

app/code/community/Bolt/Boltpay/DuplicatedTransitionException.php
```diff
@@ -0,0 +1,36 @@
...
+ * @copyright  Copyright (c) 2018 Bolt Financial, Inc (https://www.bolt.com)
+ * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+/**
+ * Class Bolt_Boltpay_DuplicatedTransitionException
+ *
+ * This exception is thrown when a transaction is trying to process order creation while a previous transaction already started before its arrival 
+ */
+class Bolt_Boltpay_DuplicatedTransitionException extends Exception
```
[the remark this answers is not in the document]
mccottry: Please rename this class and `$boltPayDuplicatedTransitionException` to `$boltPayDuplicatedTransactionException`

app/code/community/Bolt/Boltpay/controllers/ApiController.php
mccottry: @dumega 
Maybe change the message to:
`$e->getMessage().' Therefore, this redundant bolt transaction '.$reference.' will not be processed as an order in Magento'

app/code/community/Bolt/Boltpay/controllers/ApiController.php
mccottry: >When the hook that arrives belong to the same transaction, we just check its type, and if it is same as the current hook that being process the quote, I think it is fine to response 200 (or it would be great if bolt provide another special code for this situation) which actually ignore this hook. And if it is a different type, we response an error code and bolt server will retry later.

I think that you should just return an error all of the time.  We should only return 200 OK when we know for sure that the hook type has been processed successfully.  There is no guarantee the the concurrent hook will succeed.  The 200 OK will take care of itself in this situation when the other hook succeeds.  That being said, I am not sure it is possible for two hooks of the same type on the same transaction reference to come in.  I believe Bolt will wait for some sort of response before sending a repeat hook request.

> So in my opinion, auth, pending or irreversibly_rejected or any other type, no matter which type come first, at that moment the main purpose is to create the magento order and make sure only to create one unique magento order, and in the observer.php the function setInitialOrderStatusAndDetails would handle this and set the proper state for the related hook.

@dumega @daisy1754 
The question here is how is the Bolt Server designed versus the plugins expectations.  The plugin expects that if it sends a 201, the Bolt server will know that the hook did not complete successfully and only and order was created.  Then it expects the hook to be resent.  If this is true, then the code is already written like this, and there is nothing more to do.  The hook details will be handled on the subsequent request.  The Observer.php method `setInitialOrderStatusAndDetails.php` is unconcerned with the webhook details and does not set status based on the web hook.  It is only aware of if the order is being created as a result of the webhook.

If the 201 is signaling to the Bolt Server that the hook was updated on the order, then we have a logical error in the plugin code because we did not handle the hook request, we only created an order as a result of the hook request.

I agree with everything else that you have written.

dumega: @daisy1754  @mccottry 
Refer to the document *Response to API hooks* 
```
HTTP status code: 200 OK
This is a simplest form of a success response. Bolt marks the hook as successful. Optionally it updates the order with the returned information. For now, we just take merchant order reference. We will expand it to other objects in the future.
```
Then , to return 200 is not proper, and here I would like to just return an error and wont check whether the types of the processing hook and the coming hook are same or not. I think it is duty of bolt server side, on the merchant server the plugin should always process the correct logic for current processing hook and prevent duplicate entry, and the bolt server should handle the logic to determine the next hook based on the response of processed hook.

PR #147, opened 2018-08-03
[WIP] Wrap Magento's getUrl() with custom function

spgitmonish:
The custom function checks if the store/request is secure and then adds the "_secure" key with value set to true as an additional parameter.

### DIFF
--- a/app/code/community/Bolt/Boltpay/Helper/Api.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Api.php
@@ -1174,12 +1174,10 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
      * @param   array $params
      * @return  string
      */
-    public function getSecureUrl($route = '', $params = array()){
+    public function getMagentoUrl($route = '', $params = array()){
         if ((Mage::app()->getStore()->isFrontUrlSecure()) &&
             (Mage::app()->getRequest()->isSecure())) {
-            // Connection is secure
             $params["_secure"] = true;
-            return Mage::getUrl($route, $params);
         }
         return Mage::getUrl($route, $params);
     }
--- a/app/code/community/Bolt/Boltpay/Helper/Api.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Api.php
@@ -1166,4 +1166,21 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
             ? explode("|", $transaction->order->cart->display_id)[0]
             : $transaction->order->cart->display_id;
     }
+    
+    /**
+     * Generate (if) secure url by route and parameters
+     *
+     * @param   string $route
+     * @param   array $params
+     * @return  string
+     */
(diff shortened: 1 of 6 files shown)


app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -1166,4 +1166,21 @@ public function getIncrementIdFromTransaction( $transaction ) {
...
\ No newline at end of file
+
+    /**
+     * Generate (if) secure url by route and parameters
+     *
+     * @param   string $route
+     * @param   array $params
+     * @return  string
+     */
+    public function getSecureUrl($route = '', $params = array()){
```
daisy1754: I suggest you name this simply `getUrl` or `getMagentoUrl`. It's confusing if it is named secure but can return unsecure url.

daisy1754: Also suggest we don't add `$params` arg if we're not using today. YAGNI.

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -1166,4 +1166,21 @@ public function getIncrementIdFromTransaction( $transaction ) {
...
+        if ((Mage::app()->getStore()->isFrontUrlSecure()) &&
+            (Mage::app()->getRequest()->isSecure())) {
+            // Connection is secure
```
daisy1754: redundant comment; suggest removing

app/design/adminhtml/default/default/template/boltpay/system/config/button.phtml
```diff
@@ -32,7 +32,7 @@
 
         var xmlhttp = new XMLHttpRequest();
 
-        xmlhttp.open("POST", "<?=$this->getUrl('boltpay/configuration/check');?>", true);
+        xmlhttp.open("POST", "<?php echo Mage::helper('boltpay/api')->getSecureUrl('boltpay/configuration/check');?>", true);
```
daisy1754: I'm actually not very familar with php. what is the difference between `<?=` and `<? echo`?

spgitmonish: From what I read '<?=' implies '<?php echo>'

app/code/community/Bolt/Boltpay/Helper/Api.php
spgitmonish: Will fix the naming.

app/design/adminhtml/default/default/template/boltpay/system/config/button.phtml
daisy1754: that's what I thought. If that's the case I'd say let's stick to  '<?=' so this diff won't contain any change other than replacing method (also '<?=' seems to be a majority across this plugin code base)

daisy1754:
@spgitmonish Woops please don't merge unapproved PR. I just updated repository config so going forward merge button is disabled if PR is not approved..

PR #151, opened 2018-08-07
Clarification comments on similarly named functions

mccottry:
Comments added to make extra sure that developers do not make the mistake of assuming that equivalent functionality is guaranteed by similar function names in this code.

### DIFF
--- a/app/code/community/Bolt/Boltpay/controllers/OrderController.php
+++ b/app/code/community/Bolt/Boltpay/controllers/OrderController.php
@@ -168,8 +168,8 @@ class Bolt_Boltpay_OrderController extends Mage_Core_Controller_Front_Action
                 }
             }
 
-            $checkout->applyShippingMethod($this->getRequest()->getPost('shipping_method', false));
-            $quote->getShippingAddress()->setShippingMethod($this->getRequest()->getPost('shipping_method', false));
+            $checkout->applyShippingMethod($this->getRequest()->getPost('shipping_method', false));  # FireCheckout Logic for adding shipping method
+            $quote->getShippingAddress()->setShippingMethod($this->getRequest()->getPost('shipping_method', false));  # Bolt logic for adding shipping method
 
             $checkout->registerCustomerIfRequested();
 


mccottry:
@YevhenShBolt No problem. 

One version of Firecheckout does this in the method applyShipping:
```
   /**
     * Update shipping method information
     * Removes previously selected method if none is available,
     * set available if only one is available,
     * set previously selected payment,
     * set default from config if possible
     *
     * @param string $methodCode Default method code
     * @return TM_FireCheckout_Model_Type_Standard
     */
    public function applyShippingMethod($methodCode = null)
    {
        if (false === $methodCode) {
            return $this->getQuote()->getShippingAddress()->setShippingMethod(false);
        }
        $rates = Mage::getModel('sales/quote_address_rate')->getCollection()
            ->setAddressFilter($this->getQuote()->getShippingAddress()->getId())
            ->toArray();

        // unset error shipping methods. Like ups_error etc.
        $hideIfFree = Mage::getStoreConfigFlag('firecheckout/general/hide_shipping_if_free');
        foreach ($rates['items'] as $k => $rate) {
            if ('freeshipping' === $rate['method']
                && empty($rate['error_message'])
                && $hideIfFree) {

                $this->getQuote()->getShippingAddress()->setShippingMethod($rate['code']);
                return $this;
            }
            if ((empty($rate['method']) && $rate['code'] !== 'productmatrix_')
                || 'customshippingrate' == $rate['method']) {

                unset($rates['items'][$k]);
                continue;
            }

            if (!$this->_canUseShippingMethod($rate['method'])) {
                unset($rates['items'][$k]);
            }
        }
        reset($rates['items']);

        if ((!$count = count($rates['items']))) {
            $this->getQuote()->getShippingAddress()->setShippingMethod(false);
        } elseif (1 === $count && !Mage::helper('firecheckout')->isOnecolumnMode()) {
            // do not apply method for onecolumn mode, because shipping methods are hidden on initial page load
            $rate = current($rates['items']);
            $this->getQuote()->getShippingAddress()->setShippingMethod($rate['code']);
        } else {
            $found = false;
            if (!$methodCode) {
                $methodCode = $this->getQuote()->getShippingAddress()->getShippingMethod();
            }
            if ($methodCode) {
                foreach ($rates['items'] as $rate) {
                    if (is_array($methodCode) || $methodCode === $rate['code']) {
                        $this->getQuote()->getShippingAddress()->setShippingMethod($methodCode);
                        $found = true;
                        break;
                    }
                }
            }
            if (!$found || !$methodCode) {
                $methodCodes = array(
                    Mage::getStoreConfig('firecheckout/general/shipping_method_code'),
                    Mage::getStoreConfig('firecheckout/general/shipping_method')
                );
                foreach ($methodCodes as $methodCode) {
                    if (!$methodCode) {
                        continue;
                    }
                    foreach ($rates['items'] as $rate) {
                        if ($methodCode === $rate['code']) {
                            $this->getQuote()->getShippingAddress()->setShippingMethod($methodCode);
                            $found = true;
                            break 2;
                        }
                    }
                }
            }
//            if (!$found) {
//                foreach ($rates['items'] as $rate) {
//                    $this->getQuote()->getShippingAddress()->setShippingMethod($rate['code']);
//                    $found = true;
//                    break;
//                }
//            }
        }
        return $this;
    }
```

where we are only concerned with doing this:
```
$quote->getShippingAddress()->setShippingMethod($this->getRequest()->getPost('shipping_method', false));
```

Stay calm. man.  ;) The code is already proven to work in production and applyShippingMethod is proven to not set the shippingMethod in the context of the use case.  This is a very simple change that probably does not deserve this much attention. ;) ... unless you feel differently, of course.

YevhenShBolt:
Thanks for the code @mccottry 

But i want to understand with Firecheckout the following (i understand that it works fine ;-) but I am tormenting by curiosity):

- i see in `applyShippingMethod ` the code `$this->getQuote()` and `$this->getQuote()->getShippingAddress()->setShippingMethod(false)`. So this pushes me to the idea that we do the same in both methods.

It seems to me that: 

`$checkout->applyShippingMethod($this->getRequest()->getPost('shipping_method', false));` (in firecheckout) and `$quote->getShippingAddress()->setShippingMethod($this->getRequest()->getPost('shipping_method', false));` (in bolt) update the same quote.

**I am trying to understand is it so?** 

So if yes maybe better to separate it like 

> if (firecheckout_module_enabled) {
  // use firecheckout method
} else {
  // use bolt method
}

PR #152, opened 2018-08-07
Wip discount api

johnmccombs1:


app/code/community/Bolt/Boltpay/Helper/Coupon.php
```diff
@@ -0,0 +1,662 @@
...
+        try {
+            $this->requestObject = json_decode(file_get_contents('php://input'));
+        } catch (\Exception $e) {
+            $this->setErrorResponseAndThrowException(
+                self::ERR_SERVICE,
+                'Unknown error getting API request.',
+                422
+            );
+
+            throw $e;
```
daisy1754: since you call `setErrorResponseAndThrowException` there is no need to re-throw here?

app/code/community/Bolt/Boltpay/Helper/Coupon.php
```diff
@@ -0,0 +1,662 @@
...
+    }
+
+    /**
+     * Sets the response error and the http status code and then throws an exception.
+     * @param int $errCode
+     * @param string $message
+     * @param int $httpStatusCode
+     * @throws Exception
+     */
+    protected function setErrorResponseAndThrowException($errCode, $message, $httpStatusCode)
```
daisy1754: I think this method should have optional 4th argument Exception and chain it to `throw new \Exception($message)` below

app/code/community/Bolt/Boltpay/Helper/Coupon.php
```diff
@@ -0,0 +1,662 @@
...
+     */
+    protected function validateAfterApplyingCoupon()
+    {
+        $immutableQuote = $this->getImmutableQuote();
+        if ($immutableQuote->getCouponCode() != $this->getCouponCode()) {
+            $this->setErrorResponseAndThrowException(
+                self::ERR_SERVICE,
+                sprintf("Invalid coupon code response for coupon [%s]", $this->getCouponCode()),
+                422,
+                $immutableQuote
```
daisy1754: I think this 4th argument is unused

app/code/community/Bolt/Boltpay/Helper/Coupon.php
```diff
@@ -0,0 +1,662 @@
...
+    protected function setErrorResponseAndThrowException($errCode, $message, $httpStatusCode)
+    {
+        $this->responseError = [
+            'code' => $errCode,
+            'message' => $message,
+        ];
+
+        $this->httpCode = $httpStatusCode;
+        $this->responseCart = $this->getCartTotals();
+        throw new \Exception($message);
```
daisy1754: This should throw custom Exception so we can distinguish known exception (bad input) vs unknown exception

app/code/community/Bolt/Boltpay/Helper/Coupon.php
```diff
@@ -0,0 +1,662 @@
...
+            'discounts' => $cart['discounts'],
+        ];
+    }
+
+    /**
+     * Sets the success response (discount code, amount, description, and type). Also sets the HTTP status to 200.
+     *
+     * @return array
+     */
+    protected function setSuccessResponse()
```
daisy1754: Please make sure to also include common response header - https://github.com/BoltApp/bolt-magento1/blob/effeb4e1b91c5195438f0b4ddf77f9281e51a8ee/app/code/community/Bolt/Boltpay/Helper/Api.php#L1063

app/code/community/Bolt/Boltpay/Helper/Coupon.php
```diff
@@ -0,0 +1,662 @@
...
+    public function applyCoupon()
+    {
+        try {
+            $this->setupVariables();
+            $this->validateCoupon();
+            $this->applyCouponToQuotes();
+            $this->validateAfterApplyingCoupon();
+
+            $this->setSuccessResponse();
+        } catch (\Exception $e) {
```
daisy1754: let's define custom Exception type for known (expected) Exception. If there is unexpected Exception we should bugsnag.

tests/unit/testsuite/Bolt/Boltpay/Helper/CouponTest.php
```diff
@@ -0,0 +1,572 @@
...
+    /**
+     * Unit test for validating empty coupon code from request if it is passed
+     */
+    public function testPassValidateEmptyCoupon()
+    {
+        $check = true;
+        try {
+            $this->setupEnvironment(array('discount_code' => 'BOLT_CODE'));
+            $this->currentMock->validateEmptyCoupon();
+        } catch (\Exception $e) {
```
mccottry: Since you are expecting a very specific exception in the case of failures, `Bolt_Boltpay_BadInputException`, it is better to only catch that type of exception in your unit test.  In this case, by catching a generic Exception, you may inadvertently be allowing for subtle bugs to slip through the cracks.

Also consider if maybe `$passed` , `$hasPassed`, `$validated`, or `$isValid` is a better name for this variable.

I will only write this here, but this applies to all of the test that you set your `$check` in this way.

app/code/community/Bolt/Boltpay/Helper/Coupon.php
```diff
@@ -0,0 +1,684 @@
...
+
+        return $incrementId;
+    }
+
+    /**
+     * Gets immutable quote id from the request.
+     *
+     * @return int
+     */
+    protected function getImmutableQuoteId()
```
mccottry: After further consideration, you should use: `Bolt_Boltpay_Helper_Api::getImmutableQuoteIdFromTransaction` and not introduce a new method unless you feel the need to wrap the call to this to simply pass it the parameter `$this->requestObject`.

app/code/community/Bolt/Boltpay/Helper/Coupon.php
```diff
@@ -0,0 +1,684 @@
...
+    {
+        return @$this->requestObject->discount_code;
+    }
+
+    /**
+     * Gets increment id from the request.
+     *
+     * @return string
+     */
+    protected function getIncrementId()
```
mccottry: After further consideration, you should use: `Bolt_Boltpay_Helper_Api::getIncrementIdFromTransaction` and not introduce a new method unless you feel the need to wrap the call to this to simply pass it the parameter `$this->requestObject`.

app/code/community/Bolt/Boltpay/Helper/Coupon.php
```diff
@@ -0,0 +1,684 @@
...
+ * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+/**
+ * Class Bolt_Boltpay_Helper_Coupon
+ *
+ * Base Magento Bolt Coupon Helper class
+ *
+ */
+class Bolt_Boltpay_Helper_Coupon extends Mage_Core_Helper_Abstract
```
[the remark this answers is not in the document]
johnmccombs1: Yeah, I totally agree with this. Helpers play a very loosely defined role and really should be avoided unless it's a strict utility class due to the fact that they are a singleton. I think this is worth looking at when we start rewriting all of our Helper classes, which we should be starting soon. And I agree that this should be changed into a model at that point for most methods.

PR #153, opened 2018-08-07
Cache javascript cart code for same session, same order details

mccottry:
Context:
It is highly unnecessary to call Bolt to create the same order over and over again for the same session.  Furthermore, by doing this, we introduce the possibility of multiple orders being authorized by Bolt for the same session.

Proposal:

For the latter, we have already activated Bolt server integrity checks by sending the parent quote id.  By caching and reusing the Bolt JS including order token that is created, we are improving efficiency by reducing web traffic, and reducing Bolt and Magento side database space.  This also serves as the merchant-side check to make sure that the same order is presented to Bolt in a singular way with a singular order token.

I believe we have already done this in Magento 2.


mccottry:
Still adding unit test.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -266,16 +271,19 @@ public function getCartDataJs($checkoutType = self::CHECKOUT_TYPE_MULTI_PAGE)
...
-     * @return string   The calculated key for this quote's cart.  Format is {quote id}_{md5 hash of cart content}
+     * @param string                       $checkoutType  'multi-page' | 'one-page' | 'admin'
+     *
+     * @return string   The calculated key for this quote's cart.
+     *                  Format is {quote id}_{md5 hash of cart content}
      */
     protected function calculateCartCacheKey( $quote, $checkoutType ) {
-        $boltHelper = Mage::helper('boltpay/api');
-        $boltCartArray = $boltHelper->buildOrder($quote, $checkoutType === 'multi-page');
+        $boltCartArray = is_array($quote) ?: Mage::getModel('boltpay/boltOrder')->buildOrder($quote, $checkoutType === 'multi-page');
```
danac-gs: If passed $quote variable is an array (the bolt cart repesentation) shoud the $boltCartArray be that array? This way it is a bool(true).
```
$boltCartArray = is_array($quote) ? $quote : Mage::getModel('boltpay/boltOrder')->buildOrder($quote, $checkoutType === 'multi-page');
```

mccottry:
Redoing with Promisy as the base code.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: @danac-gs Yep.  Good catch!

PR #163, opened 2018-08-18
Applies shipping address from transaction to order to account for new…

mccottry:
… Bolt server initiated prefetch shipping and tax estimation calls with pseudo addresses

### DIFF
--- a/app/code/community/Bolt/Boltpay/Helper/Api.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Api.php
@@ -810,6 +810,7 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
                         'tax_amount'       => (int) round($shippingAddress->getShippingTaxAmount() * 100),
                         'service'          => $shippingAddress->getShippingDescription(),
                         'carrier'          => $shippingAddress->getShippingMethod(),
+                        'reference'        => $shippingAddress->getShippingMethod(),
                         'cost'             => (int) round($totals['shipping']->getValue() * 100),
                     ));
                     $calculatedTotal += round($totals['shipping']->getValue() * 100);
--- a/app/code/community/Bolt/Boltpay/Helper/Api.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Api.php
@@ -226,16 +226,21 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
                ->setCustomerIsGuest( (($parentQuote->getCustomerId()) ? false : true) )
                ->save();
 
+            //////////////////////////////////////////////////////////////////////////////////
+            ///  Apply shipping address and shipping method data to quote directly from
+            ///  the Bolt transaction.
+            //////////////////////////////////////////////////////////////////////////////////
+            $shippingMethodCode = null;
+            if ($transaction->order->cart->shipments) {
+                $this->applyShippingAddressToQuote($immutableQuote, $transaction->order->cart->shipments[0]->shipping_address);
+                $shippingMethodCode = $transaction->order->cart->shipments[0]->reference;
+            }
+
             $immutableQuote->getShippingAddress()->setShouldIgnoreValidation(true)->save();
             $immutableQuote->getBillingAddress()->setShouldIgnoreValidation(true)->save();
 
-            /********************************************************************
-             * Setting up shipping method by option reference
-             * the one set during checkout
-             ********************************************************************/
-            $referenceShipmentMethod = ($transaction->order->cart->shipments[0]->reference) ?: false;
-            if ($referenceShipmentMethod) {
-                $immutableQuote->getShippingAddress()->setShippingMethod($referenceShipmentMethod)->save();
+            if ($shippingMethodCode) {
+                $immutableQuote->getShippingAddress()->setShippingMethod($shippingMethodCode)->save();
             } else {
                 // Legacy transaction does not have shipments reference - fallback to $service field
                 $service = $transaction->order->cart->shipments[0]->service;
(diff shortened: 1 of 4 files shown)


app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -406,26 +344,26 @@ public function getPrefetchCacheIdentifier($quote, $addressData)
...
     /**
-     * @return mixed
+     * Returns a whether P.O. box addresses are allowed for this store
+     *
+     * @return bool     true if P.O. boxes are allowed.  Otherwise, false.
      */
     protected function isPOBoxAllowed()
     {
-        return Mage::getStoreConfig('payment/boltpay/allow_po_box');
+        return (bool) Mage::getStoreConfig('payment/boltpay/allow_po_box');
```
daisy1754: I believe you can use `getStoreConfigFlag` instead?

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -1182,4 +1193,83 @@ public function getMagentoUrl($route = '', $params = array()){
...
+
+                $address->addData($addressData);
+                $address->save();
+
+                $customer->addAddress($address)
+                    ->setDefaultShippingg($address->getId())
+                    ->save();
+            }
+        }
+        $quote->removeAllAddresses();
```
daisy1754: do you remember why we remove address here? if so can you add comment explaining why?

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -1182,4 +1193,83 @@ public function getMagentoUrl($route = '', $params = array()){
...
+    /**
+     *  Updates the shipping address data and, if necessary, the billing address data to the Magento
+     *  quote
+     *
+     * @param Mage_Sales_Model_Quote    $quote             The quote to which the address will be applied
+     * @param array                     $shippingAddress   The Bolt formatted address data
+     *
+     * @return  array   The shipping address applied in Magento compatible format
+     */
+    public function applyShippingAddressToQuote( $quote, $shippingAddress ) {
```
daisy1754: `Api.php` is already too large. I suggest we create another helper class (eg., `QuoteHelper` to manipulate quote).

daisy1754: Also suggest removing "Shipping" from method name as it would also change billing address

dumega:
@mccottry , 

When create magento order, the quote loaded by **$this->getImmutableQuoteIdFromTransaction($transaction);** is the same one with the quote when handle the last shipping&tax request, and in the **indexAction** of ShippingController, we already update the address from bolt request, then actually I do not think there is need to do that again. 

Then does our plugin really have such a logic issue?

Correct me if wrong.

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
mccottry: Thank you!

app/code/community/Bolt/Boltpay/Helper/Api.php
mccottry: @daisy1754 One reason is because we need this address to be clean and reflect only what the customer enters from the BoltCheckout modal.  Some address values can been set by other plugins and typically by the standard Magento shopping cart `Estimate Shipping and Tax` module.

mccottry:
@dumega 
> When create magento order, the quote loaded by $this->getImmutableQuoteIdFromTransaction($transaction); is the same one with the quote when handle the last shipping&tax request, and in the indexAction of ShippingController, we already update the address from bolt request, then actually I do not think there is need to do that again.

Under normal circumstances, you are exactly right and the previous way would be a more efficient way to record the Bolt stored address.  However, the problem is that the Bolt server-side logic has changed and now it guesses the user's approximate address, sends the ShippingController a psuedo-address and this is what is saved to the quote.  If the significant data (usually zipcode but could include region, country, locality, etc.) of the psuedo-address matches what the customer enters as the real address, currently, the Bolt server side does not update the ShippingController with the real address and uses the shipping and tax estimates already retrieved.

If the Bolt server side were to make a second "un-listened" call with the real address, then this code would not be necessary. I have posted this as a recommendation in **eng-integrations**, however, you may view this as a future requirement in all plugins and probably increases the robustness of the system overall, especially as Bolt tries to achieve its long-term objective of being the order engine for several e-commerce platforms.

app/code/community/Bolt/Boltpay/Helper/Api.php
mccottry: Definitely.

Long-term, I believe that we are going to want to group this particular method under the "Shipping and Tax" umbrella.  Therefore, I will first propose this as a `ShippingAndTaxHelper`.

There are several methods in the ShippingController and the Api.php that I believe are destined to go here, but, that refactor will belong to a different task.

mccottry: @daisy1754 
> Also suggest removing "Shipping" from method name as it would also change billing address

I thought a good bit about this when naming the method and originally named it as you suggest.  However, what we are doing is always applying an address to the quotes shipping address.  The billing address is only applied shipping address data if it is missing this.  I have updated the methods doc to clarify this.  If you still think it is best to remove "Shipping" from the method name, then of course there is no problem with that.

_Note:  This is old logic that I don't necessarily agree with, (i.e. replacing parts of the billing address piecemeal.  I sense that it could eventually lead to an error of an incorrect billing address.  However, changing that logic would belong to a separate pull request_

dumega:
@mccottry I see.  Then @daisy1754, the change @mccottry metioned, I think our integration team never know such a significant update in the logic on bolt servder side, and I am worried about if there is any other similar change on bolt server side, and if we do not adjust the plugin to adapt such changes, I am afraid there may be other potential issues.

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -799,6 +812,7 @@ function ($item) use ($quote, &$calculatedTotal, $boltHelper) {
                         'tax_amount'       => (int) round($shippingAddress->getShippingTaxAmount() * 100),
                         'service'          => $shippingAddress->getShippingDescription(),
                         'carrier'          => $shippingAddress->getShippingMethod(),
+                        'reference'        => $shippingAddress->getShippingMethod(),
```
dumega: well, if you add reference here, then why ignore https://github.com/BoltApp/bolt-magento1/blob/9996320e72e614cbcc4e9fa3c8dad41d0d4f5429/app/code/community/Bolt/Boltpay/Helper/Api.php#L842

mccottry: @dumega That order creation code is created from a radically different context, which is the admin.  In fact, the reference is critical for the admin to work and is already applied by that time in the code.

PR #168, opened 2018-08-28
Use DOMContentLoaded instead of onload to initialize the Bolt checkout button

YevhenShBolt:
https://app.asana.com/0/436430054069245/796327971644142/f

### DIFF
--- a/app/design/frontend/base/default/template/boltpay/replace.phtml
+++ b/app/design/frontend/base/default/template/boltpay/replace.phtml
@@ -158,8 +158,8 @@ if(!$this->isAllowedReplaceScriptOnCurrentPage() || !$this->canUseBolt() ) retur
 
     };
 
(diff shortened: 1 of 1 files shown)


app/design/frontend/base/default/template/boltpay/replace.phtml
```diff
@@ -158,8 +158,8 @@ if(!$this->isAllowedReplaceScriptOnCurrentPage() || !$this->canUseBolt() ) retur
 
     };
 
-    if (window.addEventListener) {
-        window.addEventListener("load", onloadCallback, false);
+    if (document.addEventListener) {
+        document.addEventListener("DOMContentLoaded", onloadCallback);
```
daisy1754: let's rename `onloadCallback` to `onDomReady` or something like that

PR #171, opened 2018-09-06
Dispatch order save after events

mccottry:
Magento issues a standard event `checkout_submit_all_after` after order creation that we currently do not dispatch.  

We should dispatch this event so that plugins that rely on it can function properly

### DIFF
--- a/app/code/community/Bolt/Boltpay/Helper/Api.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Api.php
@@ -390,7 +390,7 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
 
         Mage::dispatchEvent(
             'checkout_submit_all_after',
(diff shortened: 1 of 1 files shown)


daisy1754:
Great find. It's interesting we still find these stuff after this plugin has been processing lots of orders across sites... This is LGTM but I'd let other folks approve.

mccottry:
@daisy1754 

> It's interesting we still find these stuff after this plugin has been processing lots of orders across sites... 

I was thinking the same, especially, since the same plugin that highlighted this omission is also used by Indique.  The do a lot of editing of the core and plugin code directly, so, this may be why theirs functions.

mccottry:
As a side note, there is another event that is also triggered in standard onepage checkout that we do not call:
checkout_type_onepage_save_order_after

But, since it is very specific to onepage checkout and we are saving orders from many different context, I am reluctant to rush this into the main codebase and will do more investigation prior.

https://app.asana.com/0/436430054069245/810491881604516

PR #177, opened 2018-09-11
Fix for setting customer session based on quote when api hook is call…

johnmccombs1:
…ed. This will resolve issues where 3rd party code depends on customer session to process orders and payments.


app/code/community/Bolt/Boltpay/controllers/ApiController.php
```diff
@@ -73,6 +73,10 @@ public function hookAction()
             $transaction = $boltHelper->fetchTransaction($reference);
             $quoteId = $boltHelper->getImmutableQuoteIdFromTransaction($transaction);
 
+            //region set session for customer
```
daisy1754: what is this "//region" thing for?

johnmccombs1: @daisy1754 I've removed it.

PR #180, opened 2018-09-12
Add generic logger trait and helper

dumega:


app/code/community/Bolt/Boltpay/Helper/LoggerTrait.php
```diff
@@ -0,0 +1,147 @@
+<?php
+
+/**
+ * Describes log levels.
+ */
+class Bolt_Boltpay_Helper_LogLevel
+{
+    const EMERGENCY = 'emergency';
```
daisy1754: I suggest we only do error, warning and inof for now

mccottry: @dumega I agree with Kazuki in that I don't 'emergency', 'alert', and 'notice' as commonly used level in things like log4J and its derivatives.  

@daisy1754 If we do cover all of the common logger levels, perhaps in the concrete classes, the implementation of the other levels can be mapped to the three you listed above, or in cases like 'trace' ignored by some loggers like bugsnag.

daisy1754:
@mccottry actually I was thinking datadog as a concrete implementation

app/code/community/Bolt/Boltpay/Helper/LoggerTrait.php
dumega: @daisy1754 I update it, and we can easily extend the log level in the future

dumega:
@mccottry yes, I just thought we may use different logger for different workaround, and traits is good for that

PR #184, opened 2018-09-14
Feature refactor helper api

johnmccombs1:
Moved several functions out of the helper api into their own classes as an initial attempt at cleaning up the helper api.


YevhenShBolt:
Maybe better move it to model, not to the helper. I guess we previously discussed it on a weekly call with Leon also. Magento does not advice to use helpers a lot. But the most common practice is when we don't know where to put some code, and we put it in a helper. (Usually, we are just grouping some independent functions inside a "helper" class.)

Using helpers is not really in the OOP spirit (it breaks the "The SingleResponsibilityPrinciple"). 

Nice article about it - https://inchoo.net/magento/misuse-of-helper-classes-in-magento/

johnmccombs1:
> Maybe better move it to model, not to the helper. I guess we previously discussed it on a weekly call with Leon also. Magento does not advice to use helpers a lot. But the most common practice is when we don't know where to put some code, and we put it in a helper. (Usually, we are just grouping some independent functions inside a "helper" class.)
> 
> Using helpers is not really in the OOP spirit (it breaks the "The SingleResponsibilityPrinciple").
> 
> Nice article about it - https://inchoo.net/magento/misuse-of-helper-classes-in-magento/

This is a good point Yevhen. I typically like to include only classes that truly could be considered static "helpers" in helpers. I kept them in here for now because I didn't have time to test the impact of a non-singleton class change, but I think you're right that I should go ahead and move anything that makes more sense as a model into the models directory. Thanks for pushing this.

daisy1754:
If you can start this with the latest develop branch and maybe do this by one class at a time, it'll be helpful. I try to respond review quickly to avoid merge conflict

johnmccombs1:
I'm going to create a new branch and pull request and slowly push up new changes for this each day so everyone will have time to review, and I won't have to worry as much about conflicts.

PR #186, opened 2018-09-17
Added initial order lookup api code to be called from Bolt for order …

johnmccombs1:
…verification

### DIFF
--- a/app/code/community/Bolt/Boltpay/controllers/OrderController.php
+++ b/app/code/community/Bolt/Boltpay/controllers/OrderController.php
@@ -204,12 +204,11 @@ class Bolt_Boltpay_OrderController extends Mage_Core_Controller_Front_Action
     {
         try {
             $hmacHeader = $_SERVER['HTTP_X_BOLT_HMAC_SHA256'];
-            $requestJson = file_get_contents('php://input');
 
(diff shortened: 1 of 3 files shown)


app/code/community/Bolt/Boltpay/controllers/OrderController.php
```diff
@@ -196,4 +196,55 @@ public function firecheckoutcreateAction()
...
+            $reference = $this->getRequest()->getParam('reference');
+
+            if (!$reference) {
+                Mage::throwException(Mage::helper('boltpay')->__("Transaction parameter is required"));
+            }
+
+            /** @var Mage_Sales_Model_Order_Payment $orderPayment */
+            $orderPayment = Mage::getModel('sales/order_payment')
+                ->getCollection()
+                ->addFieldToFilter('last_trans_id', $reference)
```
mccottry: @johnmccombs1 
I have two questions here.  

Firstly, what happens if a refund is has been issued on the order?  Does this code still stand up in searching for the `last_trans_id`?

Second, I don't have insight to what the request to this endpoint looks like, but, I am curious as to why you choose to use `transaction id` over `display id/increment_id/merchant order id`?

Wouldn't the latter be the more robust way to lookup an order with perhaps transaction id or order reference/quote_id as a means to validate it?

PR #188, opened 2018-09-18
[WIP] Fixed 1 cent issue.

YevhenShBolt:
Steps to reproduce (task - https://app.asana.com/0/0/795970314352553/f):
1) Create a product with price 14.99$
2) Create a discount for the whole cart with 20%
3) Shipping Flat Rate with price 5$

Previously it always adds 1 cent to shipping so we have shipping price 5.01$ in Bolt popup.

### DIFF
--- a/app/code/community/Bolt/Boltpay/Helper/Api.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Api.php
@@ -1002,7 +1002,7 @@ class Bolt_Boltpay_Helper_Api extends Bolt_Boltpay_Helper_Data
                 );
             }
 
-            $adjustedShippingAmount = $this->getAdjustedShippingAmount($originalDiscountedSubtotal, $quote);
+            $adjustedShippingAmount = $shippingAddress->getShippingAmount();
 
             $option = array(
                 "service" => $this->getShippingLabel($rate),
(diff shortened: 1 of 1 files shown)


app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -1060,7 +1062,9 @@ protected function getSortedShippingRates($address) {
      *
      * @return float    Discount modified as a result of the new shipping method
      */
-    public function getAdjustedShippingAmount($originalDiscountedSubtotal, $quote) {
+    public function getAdjustedShippingAmount($originalDiscountedSubtotal, $quote)
+    {
+        $sSubtotalWithDiscount = '';
```
daisy1754: not used?

app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -990,7 +990,7 @@ public function getShippingAndTaxEstimate( $quote )
                 );
             }
 
-            $adjustedShippingAmount = $this->getAdjustedShippingAmount($originalDiscountedSubtotal, $quote);
```
johnmccombs1: @YevhenShBolt we may need to be careful here. This code has been reviewed and tweaked several times and I'm worried about removing it. I believe that the case where it is needed is when a discount is applied to the subtotal based on a shipping method. For example, "$5 off when you choose Next Day Air".

app/code/community/Bolt/Boltpay/Helper/Api.php
YevhenShBolt: oops

YevhenShBolt:
@mccottry I think it should be integration tests, not unit tests.

The problem in 1 cent it goes from Magento1 discount module, a bit later i will post a comment to the task where it happening in code.

PR #189, opened 2018-09-18
Remove method applyShippingAddressToQuote because it is unused. Move …

johnmccombs1:
…2 methods getImmutableQuoteIdFromTransaction and getIncrementIdFromTransaction to Transaction helper


app/code/community/Bolt/Boltpay/Helper/Coupon.php
```diff
@@ -290,7 +294,7 @@ public function validateFromDateForRule()
         if ($date && date('Y-m-d', strtotime($date)) > date('Y-m-d')) {
             $desc = 'Code available from ' . Mage::helper('core')->formatDate(
                     new \DateTime($rule->getFromDate()),
-                    \IntlDateFormatter::MEDIUM
+                    2//\IntlDateFormatter::MEDIUM
```
daisy1754: not sure if this is intentional?

johnmccombs1: This was intentional. The issue is that some servers don't have the php_intl extension installed (our CI server doesn't have it installed for example and it throws an exception). This adds a new requirement for Bolt which I don't like. I don't like hard coding this, but it was a quick solution. Let me see if there is alternative way to accomplish this.

PR #190, opened 2018-09-18
Added ability to modify Bolt button color via admin

johnmccombs1:


app/design/adminhtml/default/default/template/boltpay/form.phtml
```diff
@@ -28,14 +28,15 @@
...
+$buttonColor = Mage::helper('boltpay')->getBoltPrimaryColor();
 $_code = $this->getMethodCode();
 $fieldName = Mage::helper('boltpay')->__('Process Order Through Bolt');
 ?>
 <ul id="payment_form_<?php echo $_code ?>" style="display:none">
     <li>
         <div class="input-box">
             <label for="<?php echo $_code ?>_payment_button"><?php echo $fieldName ?> <span class="required">*</span></label><br/>
-            <div class="bolt-checkout-button with-cards <?= $theme; ?>"></div>
+            <div class="bolt-checkout-button with-cards <?= $theme; ?>" style="--bolt-primary-action-color:<?=$buttonColor?>"></div>
```
daisy1754: Only include `style="--bolt-primary-action-color:` if `$buttonColor` is defined

YevhenShBolt: Please do not use the PHP short tags. 
Sometimes merchant server does not enable the short_open_tag=On

I guess it is safely for us to use the <?php echo 'sometext'; ?>

PR #194, opened 2018-09-24
Moved all shipping and tax functions to a new shipping and tax model,…

johnmccombs1:
… Moved getMagentoUrl from Api helper to Url helper


mccottry:
logical

PR #196, opened 2018-09-25
Adds admin edit functionality, plus accounts for guest customer's email address

mccottry:
Context:
Currently when clients are using Authorization only on order creation, they don't have the ability to "edit" an order (in Magento terms this means that Magento will cancel the existing order and create a new duplicate order for order creation). When they try, Bolt doesn't display as a payment method. 

https://app.asana.com/0/544708310157130/802144720924881


app/code/community/Bolt/Boltpay/Helper/Api.php
```diff
@@ -743,6 +743,23 @@ function ($item) use ($quote, &$calculatedTotal, $boltHelper) {
                 'country_code',
             );
 
+            $customerEmail = $quote->getCustomerEmail();
```
daisy1754: let's create helper method getCustomerEmail($quote)

mccottry: yep.

daisy1754:
Please resolve conflict

PR #202, opened 2018-10-01
Fixes disappearing Bolt Button in firecheckout

mccottry:
Setting `cssText` to empty string was removing `--bolt-primary-action-color` attribute which triggers the problem.

`cssText` was previously needed to set the `!important` directive, however, it seems the internal Bolt produced CSS was revised to not force this, and therefore, we can use `display` without the `!important` directive.

### DIFF
--- a/app/design/frontend/base/default/template/boltpay/replace_firecheckout.phtml
+++ b/app/design/frontend/base/default/template/boltpay/replace_firecheckout.phtml
@@ -45,7 +45,7 @@ if(!Mage::helper('boltpay')->canUseBolt($this->getQuote(), false) || Mage::getSt
                 bolt_button.setAttribute("style", "--bolt-primary-action-color:<?php echo $buttonColor?>");
                 <?php endif; ?>
 
-                bolt_button.style.cssText = "display: none !important";
+                bolt_button.style.display = "none";
 
                 place_order_button.parentNode.insertBefore(bolt_button, place_order_button);
             }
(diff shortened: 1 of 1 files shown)


app/design/frontend/base/default/template/boltpay/replace_firecheckout.phtml
```diff
@@ -64,7 +64,7 @@ if(!Mage::helper('boltpay')->canUseBolt($this->getQuote(), false) || Mage::getSt
             if (this.id == "p_method_boltpay") {
                 bolt_buttons.forEach(
                     function (bolt_button) {
-                        bolt_button.style.cssText = "";
+                        bolt_button.style.display = "";
```
daisy1754: shouldn't we specify display properly explicitly? "block"?

mccottry: @daisy1754 

It is not necessary.  The default value for a div element like the one we create is always "block".  Furthermore, we have never explicitly defined this up to this point.  I think that it is better to leave this as undefined and let defaults or other any other means of stylings which the merchant has defined to set this value.

If we were to set the inline style, like `element.style` does, we are overriding all stylesheet defined, and all `<style>` element defined stylings.  We don't want this if we are trying to play nice with variable unknown templates.

However, what reasoning would guide us to explicitly defining the inline styling of this element as  `block` as the better solution?

daisy1754: nvm I was not sure about implication of empty string. I was afraid that it might be hiding div but today I learned it just reset to default display value - https://stackoverflow.com/questions/7420109/what-does-style-display-actually-do

PR #209, opened 2018-10-03
Preserve and default address data

mccottry:
We are currently saving orders to Magento without some billing info.  Currently, we only assure that all the shipping info is in place.  

We must make sure that billing info is complete as well so that client plugins and third party software behave normally.


app/code/community/Bolt/Boltpay/Model/Quote.php
```diff
@@ -175,24 +175,33 @@ function ($item) use ($quote, &$calculatedTotal, $boltHelper) {
...
+            /*////////////////////////////////////////////////////////////////////////////
+               -------------------------------------
+                Billing / shipping address fields that are required
+                when the address data is sent to Bolt.
+                ------------------------------------
                 'first_name',
                 'last_name',
                 'street_address1',
                 'locality',
                 'region',
```
daisy1754: I don't think region is required depending on country

app/code/community/Bolt/Boltpay/Model/Quote.php
```diff
@@ -175,24 +175,33 @@ function ($item) use ($quote, &$calculatedTotal, $boltHelper) {
...
         } else {
 
-            // Billing / shipping address fields that are required when the address data is sent to Bolt.
-            $requiredAddressFields = array(
+            $billingAddress  = $quote->getBillingAddress();
+            $shippingAddress = $quote->getShippingAddress();
+
+            /*////////////////////////////////////////////////////////////////////////////
+               -------------------------------------
+                Billing / shipping address fields that are required
```
daisy1754: what is the purpose of this comment?

app/code/community/Bolt/Boltpay/Model/Quote.php
mccottry: @daisy1754 The doc https://docs.bolt.com/reference#create-order needs to be updated to express this.  It simply list region as a required field.

I certainly have seen that for some countries, like the Bahamas, it is not required by the Bolt full checkout modal.  Shall I update the doc?

app/code/community/Bolt/Boltpay/Model/Quote.php
mccottry: As you can see from the previous version, this was once an array and the commit was called `removed unnecessary array and left as comment`

https://github.com/BoltApp/bolt-magento1/pull/209/commits/4b22bd9f8d1d1a3b1e7d6e6b3f7a46fb67cc8ff8

My original thoughts was to remove it entirely, however, it does provide some useful information to a developer, so, I believe that it adds some sort of long-term benefit and does not do much harm.  Certainly, it can be removed.

app/code/community/Bolt/Boltpay/Model/Quote.php
```diff
@@ -246,13 +246,13 @@ function ($item) use ($quote, &$calculatedTotal, $boltHelper) {
...
+                    'region'          => $shippingRegion,
+                    'postal_code'     => $shippingAddress->getPostcode() ?: '-',
                     'country_code'    => $shippingAddress->getCountry(),
                     'phone'           => $shippingAddress->getTelephone(),
-                    'email'           => $shippingAddress->getEmail() ?: $customerEmail,
+                    'email'           => $shippingAddress->getEmail() ?: ($customerEmail ?: $billingAddress->getEmail()),
                     'phone_number'    => $shippingAddress->getTelephone(),
-                    'email_address'   => $shippingAddress->getEmail() ?: $customerEmail,
+                    'email_address'   => $shippingAddress->getEmail() ?: ($customerEmail ?: $billingAddress->getEmail()),
                 );
```
danac-gs: The doubling of fields, phone / phone_number and email / email_address, is a relic of the past when at some point the names were changing back and forth on the Bolt server side. The line of communication was not so well established at the time so this was my solution to it. I think phone_number and email_address are redundant now, at least they are not present in Magento2 plugin create order request.

mccottry: @danac-gs I'm aware of this.  Both those fields are undocumented and with confirmation from @daisy1754, I will remove those fields.

daisy1754: can confirm these are no longer used

PR #216, opened 2018-10-05
Preserves converted immutable quotes in code cleanup

mccottry:


mccottry:
WIP will be removed after sanity check test run.

PR #222, opened 2018-10-10
Moved the logic to set quote_id to session after setting customer to session

johnmccombs1:
I found a bug when testing on Golf land. Parent quote is being removed when estimating the shipping with login customer.

When setting the customer to session, it will get a current customer quote which is parent quote and then merges the quote that being on the session and then remove the parent quote.

### DIFF
--- a/app/code/community/Bolt/Boltpay/controllers/ShippingController.php
+++ b/app/code/community/Bolt/Boltpay/controllers/ShippingController.php
@@ -68,12 +68,6 @@ class Bolt_Boltpay_ShippingController extends Mage_Core_Controller_Front_Action
             /* @var Mage_Sales_Model_Quote $quote */
             $quote = Mage::getModel('sales/quote')->loadByIdWithoutStore($quoteId);
 
-            /***********************/
-            // Set session quote to real customer quote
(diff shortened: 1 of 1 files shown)


app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -103,6 +97,12 @@ public function indexAction()
             } else {
                 Mage::helper('boltpay')->setCustomerSessionById($quote->getCustomerId());
```
mccottry: Good insight.  The fix here is setting the customer id to the session prior to setting the quote id to the session.

Instead of moving the block of code for setting the quote id to the session, please move this line of code with regards to setting the customer id to the session to above the original position of session quote id being set to the session (i.e. before line 71).

Both the customer id and the quote id need to be set here as they are used in both portions of the succeeding `if\else` block.  Otherwise, you introduce the possibility of errors when this code is reached:

```
                     $response = $shippingAndTaxModel->getShippingAndTaxEstimate($quote);
                     $cacheBoltHeader = 'MISS';	                     $cacheBoltHeader = 'MISS';
```

PR #219, opened 2018-10-09
Fixes the no longer supported empty body response by handling invalid…

mccottry:
… hook transitions

PR #232, opened 2018-10-17
Set correct response body

johnmccombs1:


app/code/community/Bolt/Boltpay/controllers/ApiController.php
```diff
@@ -138,7 +138,7 @@ public function hookAction()
                         )
                     )
                 );
-                $this->getResponse()->setHttpResponseCode(200);
+                $this->getResponse()->setHttpResponseCode(200)->setBody(json_encode(array('status' => 'success')));
```
YevhenShBolt: Maybe better using the already exists function `sendResponse`, how do you think?

mccottry:
Please see #219

PR #234, opened 2018-10-18
Set backoffice order authcapture to true

johnmccombs1:


app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -234,12 +234,13 @@ protected function isAdminAndUseJsInAdmin($checkoutType)
...
-     * @param $orderCreationResponse
+     * @param      $orderCreationResponse
+     * @param null $checkoutType
      * @return array
      */
-    public function buildCartData($orderCreationResponse)
+    public function buildCartData($orderCreationResponse, $checkoutType = null)
     {
-        $authCapture = Mage::getStoreConfigFlag('payment/boltpay/auto_capture');
+        $authCapture = $checkoutType === 'admin' ? true : Mage::getStoreConfigFlag('payment/boltpay/auto_capture');
```
YevhenShBolt: I guess better to use Mage_Core_Model_Store::ADMIN_CODE instead of 'admin'.

daisy1754: this checkoutType is our own constant so technically it's not related to Mage_Core_Model_Store::ADMIN_CODE

PR #238, opened 2018-10-25
Allow shipping method to be selected multiple times during admin orde…

johnmccombs1:
…r creation


app/code/community/Bolt/Boltpay/Helper/Data.php
```diff
@@ -332,7 +351,10 @@ public function getItemImageUrl($item)
         /** @var Mage_Catalog_Model_Product $_product */
         $_product = $item->getProduct();
 
-        $image = $imageHelper->init($_product, 'thumbnail', $_product->getThumbnail());
+        $image = '';
+        try{
+            $image = $imageHelper->init($_product, 'thumbnail', $_product->getThumbnail());
+        }catch (Exception $e){}
```
YevhenShBolt: I guess we should add in 
` }catch (Exception $e){ $item = '';} `

it seems that after exception we will have a wrong value in a variable.

PR #249, opened 2018-11-13
Adds custom title/suppresses close icon display

mccottry:


mccottry:
Example of change:

image

PR #251, opened 2018-11-14
Automatically copies billing address to shipping address when billing…

mccottry:
… as shipping is selected

### DIFF
--- a/js/boltpay/admin_order.js
+++ b/js/boltpay/admin_order.js
@@ -36,7 +36,11 @@ AdminOrder.prototype.prepareParams =
                 var billingAddressData = this.serializeData('order-billing_address');
                 if (billingAddressData) {
                     billingAddressData.each(function(value) {
-                        document.querySelector("[name='"+value[0].replace('billing_', 'shipping_')+"']").value = value[1];
+                        try {
+                            document.querySelector("[name='"+value[0].replace('billing_', 'shipping_')+"']").value = value[1];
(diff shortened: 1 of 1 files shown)


js/boltpay/admin_order.js
```diff
@@ -24,6 +25,24 @@ AdminOrder.prototype.prepareParams =
...
+        ////////////////////////////////////////////////////////////////////
+        var billingAddressContainer = $('order-billing_address');
+        if (billingAddressContainer) {
+
+            var isBillingAddressShipping = document.getElementById('order-shipping_as_billing');
+            if (isBillingAddressShipping && isBillingAddressShipping.checked) {
+                var billingAddressData = this.serializeData('order-billing_address');
+                if (billingAddressData) {
+                    billingAddressData.each(function(value) {
+                        document.querySelector("[name='"+value[0].replace('billing_', 'shipping_')+"']").value = value[1];
```
daisy1754: suggest checking `document.querySelector("[name='"+value[0].replace('billing_', 'shipping_')+"']")` is non-null before setting value

LiHsuanChen:
Other than the potential null pointer, everything looks good to me.

js/boltpay/admin_order.js
mccottry: @daisy1754  Updated code to handle this and log it so that anomalies can easily be detected and support extended to these cases.

PR #255, opened 2018-11-23
Can't place order on Magento 1.7 due to newly generated address

johnmccombs1:


daisy1754:
@mccottry @danac-gs @johnmccombs1 does anyone know why the deleted code was there? if not I'm fine removing (code doesn't make sense to me) 
If there is reason I hope we can add some comments explaining why

mccottry:
> @mccottry @danac-gs @johnmccombs1 does anyone know why the deleted code was there? if not I'm fine removing (code doesn't make sense to me)
> If there is reason I hope we can add some comments explaining why

@daisy1754 I think it is a bad idea to remove this code or in general to remove any production code when there is no clear idea of it's use.

I didn't add this code, but I have commented in the past that the intent of it that I perceive and understand is the case where someone uses the Magento mechanism for estimating shipping and tax.  When used, a partial address would be added to the quote address and from their, I can imagine potential problems occurring on several levels.  So, here, it looks like this address data is temporarily removed, and then afterwards is reapplied.

1.7 seems to present an exceptional case, and therefore, this code omission could be okay in that context.

johnmccombs1:
> > @mccottry @danac-gs @johnmccombs1 does anyone know why the deleted code was there? if not I'm fine removing (code doesn't make sense to me)
> > If there is reason I hope we can add some comments explaining why
> 
> @daisy1754 My opinion is it is risky business to remove this code or in general to remove any production code when there is no clear idea of it's use.
> 
> I didn't add this code, but I have commented in the past that the intent of it that I perceive and understand is the case where someone uses the Magento mechanism for estimating shipping and tax. When used, a partial address would be added to the quote address and from their, I can imagine potential problems occurring on several levels. So, here, it looks like this address data is temporarily removed, and then afterwards is reapplied.
> 
> Certainly, your suggestion of adding commenting is in order here.
> 
> 1.7 seems to present an exceptional case, and therefore, this code omission could be okay in that context.

Just to confirm, @mccottry are you suggesting that I add a special case for this code that it's bypassed only for M1.7? I'm fine with this solution as I agree with you that we don't know why this is in there and it's possible that it's necessary even if we can determine why at the moment.

mccottry:
> Just to confirm, @mccottry are you suggesting that I add a special case for this code that it's bypassed only for M1.7? I'm fine with this solution as I agree with you that we don't know why this is in there and it's possible that it's necessary even if we can determine why at the moment.

Yes.  I am in agreement with @YevhenShBolt 's initial proposal of adding a special case for M1.7.  It is the safest way to make the change which you are implementing.  I also am in agreement with @daisy1754 that some comment should be added that emphasizes that the code to remove the address exist to account for address additions in cases such as the Magento mechanism for estimating shipping and tax that could potential interfere with customer entered shipping and tax data via the Bolt Modal.

PR #257, opened 2018-11-27
Adds support for address 3 and 4 fields

mccottry:
https://app.asana.com/0/544708310157130/955169155915862

Some merchants may implement and support street address 3 and 4.  Magento supports this and Bolt can support this.

Additionally, some may have active validation on the values of street address line 3 and street address line 4, checking if it is a string which is likely done through the event `customer_address_validation_after`.  

Here, as a JIC (just-in-case), we provide an empty string value for these fields as our default.


mccottry:
Note: 
Something seems to be wrong with the Selenium integration test that is not related to this change.  Seems to be a connection being refused.  JAVA stack trace says:

```
Could not retrieve console log for failed test { Error: An unknown server-side error occurred while processing the command.
    at log("browser") - index.js:263:37
    at elements("//*[contains(text(), 'Thank you for your purchase!')]") - isExisting.js:46:17
    at isExisting("//*[contains(text(), 'Thank you for your purchase!')]") - waitForExist.js:67:22
  details: undefined,
  message: 'java.net.ConnectException: Connection refused (Connection refused)',
  type: 'RuntimeError',
  seleniumStack: 
   { type: 'UnknownError',
     message: 'An unknown server-side error occurred while processing the command.',
     orgStatusMessage: 'java.net.ConnectException: Connection refused (Connection refused)' } }
ERROR: java.net.ConnectException: Connection refused (Connection refused)
chrome
    at new RuntimeError (/home/travis/build/BoltApp/bolt-magento1/integration-tests/node_modules/webdriverio/build/lib/utils/ErrorHandler.js:143:12)
```

mccottry:
Testing problems resolved and passing.

app/code/community/Bolt/Boltpay/Model/ShippingAndTax.php
```diff
@@ -43,11 +43,16 @@ public function applyShippingAddressToQuote( $quote, $shippingAddress ) {
             throw new Exception(Mage::helper('boltpay')->__("Address must contain postal_code and country_code."));
         }
 
+        $shippingStreet = (@$shippingAddress->street_address1 ?: '') . "\n"
```
daisy1754: can you do something like:

```
implode('\n', array_filter(array(
  @$shippingAddress->street_address1,
  @$shippingAddress->street_address2,
  @$shippingAddress->street_address3,
  @$shippingAddress->street_address4))
```

Your test results in `'Sample Street 10' . "\n" . 'Apt 123'` instead of `'Sample Street 10' . "\n" . 'Apt 123\n\n'`

daisy1754: ok fair point. how about `trim` then? magento does trim.
https://github.com/mgtf/magento-1.9.2.0/blob/b5c1fc794d477f7d59b43c8740da0c9597e047a8/app/code/core/Mage/Customer/Model/Address/Abstract.php#L162

will it cause issue if newline code at the end is missing?

It seems weird to me that we add extra new lines code but if we need these for make magento happy I'm ok

mccottry: @daisy1754 trim will work.

daisy1754: can we trim in plugin, not in test?

I want to understand if we need trailing newlines for `Sample Street 10\nApt 123\n\n`. If not needed, I'd prefer removing these. It may cause weird UI (extra spaces) somewhere.

mccottry: > can we trim in plugin, not in test?
> 
> I want to understand if we need trailing newlines for `Sample Street 10\nApt 123\n\n`. If not needed, I'd prefer removing these. It may cause weird UI (extra spaces) somewhere.

Ok.  I will ignore this:

>Additionally, some may have active validation on the values of street address line 3 and street address line 4, checking if it is a string which is likely done through the event customer_address_validation_after.
>Here, as a JIC (just-in-case), we provide an empty string value for these fields as our default.

Instead, if an error happens in validation, we can consider it as a need for merchant specific custom code.

tests/unit/testsuite/Bolt/Boltpay/Model/ShippingAndTaxTest.php
```diff
@@ -73,7 +73,8 @@ public function testApplyShippingAddressToQuote() {
             'region' => 'California'
         );
         $result = $this->currentMock->applyShippingAddressToQuote($quote, $shippingAddress);
-
+        array_walk( $result, function(&$value, $key){ $value = trim($value); } );
```
daisy1754: remove this

tests/unit/testsuite/Bolt/Boltpay/Model/ShippingAndTaxTest.php
```diff
@@ -103,12 +104,13 @@ public function testApplyShippingAddressToQuote_withsomefieldmissing() {
...
```
daisy1754: remove this

tests/unit/testsuite/Bolt/Boltpay/Model/ShippingAndTaxTest.php
mccottry: @daisy1754, perhaps it maybe good to add this to the main plugin code.  Trailing newlines and spaces have no place in this context and in fact can cause problems with Magento's algorithms

daisy1754:
let's merge

PR #263, opened 2018-12-06
Remove jQuery dependency

knightaaron838:
magento1.8 does not contain jQuery out-of-box


knightaaron838:
Yeah, we can do it only for 1.8

And I already include noconflict js file for different version

I think, it is better than loading from external CDN

mccottry:
> Couple of questions:
> 
> * Can we do this only or 1.8
> * What happens if store already has jQuery with different version?
> * Is it better to include jQuery in our plugin vs loading from external CDN?

@daisy1754 It should be pointed out that up to this point, a lot of care has been taking not to force include jQuery in the plugin and to use native javascript or prototype.js/scriptaculous equivalents instead to avoid some of the issues that you mentioned above concerning conflicts.

The one exception, which I believe is an erroneous exception was the Bolt error message popup to replace `alert`

However, if we will be introducing jQuery to the plugin, then there is also an opportunity to update the existing javascript to reflect this.

knightaaron838:
It's good idea.
I will change the jquery codes to javascript code

app/design/frontend/base/default/template/boltpay/popup.phtml
```diff
@@ -16,37 +16,40 @@
...
-            this.popupElm.find('.popup-content').html(message);
+            this.popupElm.getElementsByClassName('popup-content')[0].innerHTML = message;
             return this;
         },
         setMessage: function(message) {
             this.addMessage(message);
         },
         setTitle: function(title) {
-            this.popupElm.find('.popup-title').html(title);
+            this.popupElm.getElementsByClassName('.popup-title').innerHTML = title;
```
daisy1754: `getElementsByClassName ` returns NodeList. you need index

app/design/frontend/base/default/template/boltpay/popup.phtml
```diff
@@ -16,37 +16,40 @@
...
+                popup_close[i].addEventListener( 'click', function() {
+                    self.close();
+                });
+            }
         },
         close: function() {
-            this.popupElm.removeClass('active');
-            this.popupElm.fadeOut();
+            this.popupElm.classList.remove('active');
+            this.popupElm.style.display = "none";
```
daisy1754: this changes behaviour. can you fadeout. there should be a css

app/design/frontend/base/default/layout/boltpay.xml
```diff
@@ -19,7 +19,7 @@
 <layout version="0.1.0">
 
   <default>
-    <reference name="head">
+    <reference name="head">
```
daisy1754: unrelated change

app/design/frontend/base/default/layout/boltpay.xml
```diff
@@ -20,7 +20,7 @@
 
   <default>
     <reference name="head">
-      <block type="boltpay/checkout_boltpay" name="track_script" as="track_script" template="boltpay/track.phtml" />
+	    <block type="boltpay/checkout_boltpay" name="track_script" as="track_script" template="boltpay/track.phtml" />
```
daisy1754: unrelated change

app/design/frontend/base/default/template/boltpay/popup.phtml
```diff
@@ -16,37 +16,53 @@
...
+                    element.style.display = 'none';
+                }
+                element.style.opacity = op;
+                element.style.filter = 'alpha(opacity=' + op * 100 + ")";
+                op -= op * 0.1;
+            }, 50);
         },
         addMessage: function(message) {
-            this.popupElm.find('.popup-content').html(message);
+            this.popupElm.getElementsByClassName('popup-content')[0].innerHTML = message;
```
mccottry: You might want to be careful here and setTitle referencing the 0 element of this array.  Probably, the element which you are searching for should always exist, but, who knows.

PR #265, opened 2018-12-13
Promisifies the checkout process across one-page, multi-page, and admin

mccottry:
Context:
We have moved to the Promise as a standard means of returning a Bolt order token and autocapture directive.
One advantage of this approach is that it allows for more simple control of efficient on demand Bolt order and Magento quote creation.

Proposal:
Extend the Promise approach to Payment Only, Multi Step, and Backend/Admin/Console Bolt order creation. We have already implemented the Promise for Firecheckout.  

https://app.asana.com/0/544708310157130/942487273563864


mccottry:
Pending update of testing benchmarks and a proper run through production readiness.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -134,6 +107,62 @@ function() {
...
+                                function() {
+                                     if ($validPreconditionsCheck) {
+                                        new Ajax.Request('$checkoutTokenUrl', {
+                                            method:'post',
+                                            parameters: $postData,
+                                            onSuccess: function(response) {
+                                                if(response.responseJSON.error) {                                                        
+                                                    reject(response.responseJSON.error_messages);
+                                                    
+                                                    // BoltCheckout is currently not doing anything reasonable to alert the user of a problem, so we will do something as a backup
```
daisy1754: please file a asana task with details.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -134,6 +107,62 @@ function() {
...
+            $validPreconditionsCheck = 'true';
+            $postData = "'true'";
+        }
+
+        return <<<PROMISE
+                    new Promise( 
+                        function (resolve, reject) {
+                            var checkoutAjaxId = setInterval(
+                                function() {
+                                     if ($validPreconditionsCheck) {
```
daisy1754: hmm, this seems hacky.. can we instead listen for form state change and call `BoltCheckout.configure` whenever needed? I assume we'll need that logic anyway to deal with change in form data?

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -197,18 +226,8 @@ public function getCartDataJs($checkoutType = self::CHECKOUT_TYPE_MULTI_PAGE)
...
-                // Only shipping, tax and discounts can change, and only if the shipping, tax
-                // and discount calculations change on the Magento server
-                ////////////////////////////////////////////////////////////////////////////////
-                /** @var Mage_Sales_Model_Quote $immutableQuote */
-                $immutableQuote = $boltHelper->cloneQuote($sessionQuote, $checkoutType);
-                ////////////////////////////////////////////////////////////////////////////////
-
-                $orderCreationResponse = $this->getBoltOrderToken($immutableQuote, $checkoutType);
+
+                $orderCreationResponse = $this->getBoltOrderToken(null, $checkoutType);
```
daisy1754: - why not directly call `getBoltOrderTokenPromise`?
- variable should be named something like promise

app/code/community/Bolt/Boltpay/controllers/OrderControllerTrait.php
```diff
@@ -0,0 +1,60 @@
...
+trait Bolt_Boltpay_OrderControllerTrait {
+
+    /**
+     * Creates the Bolt order and returns the Bolt.process javascript.
+     */
+    public function createAction() {
+
+        try {
+            if (!$this->getRequest()->isAjax()) {
+                Mage::throwException(Mage::helper('boltpay')->__("Bolt_Boltpay_Adminhtml_Sales_Order_CreateController::createAction called with a non AJAX call"));
```
daisy1754: is `Bolt_Boltpay_Adminhtml_Sales_Order_CreateController` relevant here?

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -134,6 +107,62 @@ function() {
...
+            $validPreconditionsCheck = 'isFireCheckoutFormValid';
+            $postData = 'checkout.getFormData ? checkout.getFormData() : Form.serialize(checkout.form, true)';
+        } else if ( $checkoutType === self::CHECKOUT_TYPE_ADMIN ) {
+            $checkoutTokenUrl = Mage::helper('boltpay/url')->getMagentoUrl("adminhtml/sales_order_create/create/checkoutType/$checkoutType", true);
+            $validPreconditionsCheck = ( $checkoutType === self::CHECKOUT_TYPE_ADMIN ) ? 'isAdminFormValid' : 'true';
+            $postData = "'true'";
+        } else {
+            $checkoutTokenUrl = Mage::helper('boltpay/url')->getMagentoUrl("boltpay/order/create/checkoutType/$checkoutType");
+            $validPreconditionsCheck = 'true';
+            $postData = "'true'";
```
daisy1754: what does true mean here?

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -134,6 +107,62 @@ function() {
...
+     * @return string javascript Promise string used for initializing BoltCheckout
+     */
+    protected function getBoltOrderTokenPromise($checkoutType) {
+
+        if ( $checkoutType === self::CHECKOUT_TYPE_FIRECHECKOUT ) {
+            $checkoutTokenUrl = Mage::helper('boltpay/url')->getMagentoUrl('boltpay/order/firecheckoutcreate');
+            $validPreconditionsCheck = 'isFireCheckoutFormValid';
+            $postData = 'checkout.getFormData ? checkout.getFormData() : Form.serialize(checkout.form, true)';
+        } else if ( $checkoutType === self::CHECKOUT_TYPE_ADMIN ) {
+            $checkoutTokenUrl = Mage::helper('boltpay/url')->getMagentoUrl("adminhtml/sales_order_create/create/checkoutType/$checkoutType", true);
```
matt-thomason: As far as I can tell, the second param for `getMagentoUrl` needs to be an array. Do you mean to have some sort of `['someParam' => true]` here instead?

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -134,6 +107,62 @@ function() {
...
+            $validPreconditionsCheck = ( $checkoutType === self::CHECKOUT_TYPE_ADMIN ) ? 'isAdminFormValid' : 'true';
```
matt-thomason: Does `true` here mean we do not need any precondition checks?

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -423,7 +424,7 @@ public function buildOnCheckCallback($checkoutType, $quote)
                         } "). "
         
                         bolt_hidden.classList.add('required-entry');
-                        return is_valid;
+                        return (isAdminFormValid = is_valid);
```
matt-thomason: What is this assignment used for? Do you mean to do the assignment here? It works fine since it's returning `is_valid`, but this is a little misleading and hard to read.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
This is doing exactly what your are saying.  It is listen for a state change.  When the state is valid it moves on.  There are different checkout context, and so this is supplied by PHP on page render depending on the context.

I highly disagree with the classification as "hacky", and the assumption that there will always be a form is incorrect.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: @matt-thomason Good catch.  I missed the second parameter and should have passed an empty array.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: Yes.  Also, the conditional ternary operator is unnecessary.  Updating.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: @daisy1754 wrote:
hmm, this seems hacky.. can we instead listen for form state change and call BoltCheckout.configure whenever needed? I assume we'll need that logic anyway to deal with change in form data?

mccottry: @daisy1754 ok.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: I'll remove it to simplify.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: done.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: Good point.  I probably was to caught up in the idea of preserving legacy interfaces.  I will update to a direct call and remove the `null` condition.

For the variable name, due to that it can be something other than a Promise, perhaps using the BoltCheckout name for this, `cartData` is the most appropriate.

app/code/community/Bolt/Boltpay/controllers/OrderControllerTrait.php
mccottry: Good catch.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: On second thought, the change that you ask for is more involved and I believe it should be handled in a separate pull request.  Particularly, that PR would involve changing the long-standing how and when to call `BoltCheckout.configure`.  This PR is already involved enough.
  
Also, I am not in agreement that this approach is hacky and would need to understand the end goal a bit better with a discussion.  The other approach will involve more code in more areas, utilizing the same strategy and I estimate will appear at least the same level of "hackiness" and probably more.  So, that is further reason for that to be dealt in it's own devoted pull request and would make this PR too far involved in another unrelated topic.

daisy1754: As far as I can tell, current code execute the function every 300ms and check the condition `isReadyToCreateBoltOrder`. Can we instead explicitly call `BoltCheckout.configure` when you are ready to create Bolt order?

mccottry: done.

mccottry: @daisy1754 done.

mccottry: SetInterval removed

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -328,70 +331,66 @@ public function buildBoltCheckoutJavascript($checkoutType, $quote, $hintData, $c
...
-                  success: $onSuccessCallback,
-
-                  close: function() {
-                     $onCloseCallback
-                  }
-                }
-        );");
+            var configureIntervalId = setInterval(
+                function() {
+                     if (isReadyToCreateBoltOrder) {
```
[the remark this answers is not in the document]
mccottry: resolved

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -423,13 +422,14 @@ public function buildOnCheckCallback($checkoutType, $quote)
...
                         bolt_hidden.classList.add('required-entry');
+                        isReadyToCreateBoltOrder = is_valid;
                         return is_valid;
                     }
                     ";
             case self::CHECKOUT_TYPE_FIRECHECKOUT:
                 return
                     "
-                    return (isFireCheckoutFormValid = checkout.validate());
+                    return (isReadyToCreateBoltOrder = checkout.validate());
```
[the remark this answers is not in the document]
mccottry: it's gone.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -197,20 +212,10 @@ public function getCartDataJs($checkoutType = self::CHECKOUT_TYPE_MULTI_PAGE)
...
-                $immutableQuote = $boltHelper->cloneQuote($sessionQuote, $checkoutType);
-                ////////////////////////////////////////////////////////////////////////////////
-
-                $orderCreationResponse = $this->getBoltOrderToken($immutableQuote, $checkoutType);
-
-                if (@!$orderCreationResponse->error) {
+
+                $cartData = $this->getBoltOrderTokenPromise($checkoutType);
+
+                if (@!$cartData->error) {
```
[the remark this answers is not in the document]
mccottry: true, no pun intended.  will update.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -134,6 +103,52 @@ function() {
...
+                            new Ajax.Request('$checkoutTokenUrl', {
+                                method:'post',
+                                parameters: $parameters,
+                                onSuccess: function(response) {
+                                    if(response.responseJSON.error) {                                                        
+                                        reject(response.responseJSON.error_messages);
+                                        
+                                        // BoltCheckout is currently not doing anything reasonable to alert the user of a problem, so we will do something as a backup
+                                        alert(response.responseJSON.error_messages);
+                                        location.reload();
```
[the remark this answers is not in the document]
mccottry: As implemented, since the event that we are relying upon is form submission, there should be no problem here in the average use case as the retry is triggered by user interaction with a form.  Alert and reload was an old approach to the Bolt modal not supplying sufficient mechanisms for displaying a message to the user and closing the Bolt modal.

I will implement these and do this in the normal, elegant way.  When the Bolt modal code is updated, then we can use those mechanisms.

app/design/adminhtml/default/default/template/boltpay/form.phtml
```diff
@@ -47,6 +47,9 @@ $fieldName = Mage::helper('boltpay')->__('Process Order Through Bolt');
 </ul>
 
 <script type="text/javascript">
+
+    var isReadyToCreateBoltOrder = false;
```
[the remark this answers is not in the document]
mccottry: Not needed.  I thought I removed all of these.  This is an unused variable and will be removed.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -423,13 +439,14 @@ public function buildOnCheckCallback($checkoutType, $quote)
...
-                        return is_valid;
+                        if (!is_valid) return false;
                     }
                     ";
             case self::CHECKOUT_TYPE_FIRECHECKOUT:
                 return
                     "
-                    return (isFireCheckoutFormValid = checkout.validate());
+                    var is_valid = checkout.validate();
+                    if (!is_valid) return false;
```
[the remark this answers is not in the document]
mccottry: obviously.  As we see, before, I did an assignment operation and just returned the value of that.  This was for readability in self describing code.  That approach was questioned as unclear or confusing by Matt, so we opted for this style and for the sake of consistency with the block of code directly above that uses the same variable names and identifies a design pattern.  However, I will update it to your style preference which is as I would also naturally write that line of code.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -390,8 +374,40 @@ public function buildBoltCheckoutJavascript($checkoutType, $quote, $hintData, $c
...
+            var json_cart = $jsonCart;
+            var json_hints = \$hints_transform($jsonHints);
+            var order_completed = false;
+            var check_count = 0;
+            var configure_bolt = function() {
+                $boltConfigureCall
+                return true;
+            };
+            
+            BoltCheckout.configure(
```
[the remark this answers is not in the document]
mccottry: @daisy1754 I have full confidence that it works as I tested both.  However, I did implement this in an alternative manner that I passed over to avoid conversations about "hacky" approaches.  *I will use the alternative approach, but would advice that the Bolt code not change so that configure could be called from within check as it can be now.  This is very powerful as is.*

Monitoring a blur event is the most expensive event to be monitoring and would invoke several blocking modals decreasing UX in both firecheckout and the admin at minimum.  In worse case, we also incur several http ajax calls and possible create several useless quotes.  This approach should be avoided when possible and it is proven possible to avoid.  If we want to check the state without this extra overhead and not rely on the natural event, **Bolt button click**, with the current design, we should use the initial approach polling approach.

Still, the most reasonable present and/or future alternatives for event driven efficient Bolt button configuration are:

1.) Bolt provides an click event that is performed prior to opening the modal that provides the opportunity to run or re-run configure.  Preferably, configure is not run until click.  In other words, separate the idea of Bolt configuration from that of Bolt Button render.  They are not the same and should not be treated in the same way.  Or if we want to use different terms to think about this more clearly, we can say that we first `configure` the Bolt button and once it is configured, we `render` the button.  After the rendered button is clicked, we `check` the button to if we can `load` the bolt form.  Loading options set by configure can be change at this time prior to the bolt form loading.

2.) We accept the fact that the `check` function allows for configure to be recalled and we make sure to respect this and ensure that it remains supported in the future.  This is probably the best approach that only requires the addition of some other unit and integration test for the Bolt modal.  It also makes sense that in the Bolt `check` method, be perform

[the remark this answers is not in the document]
daisy1754: Both your suggestion and current implementation initialize token only after user clicks the button. Given "instant load" is a huge value prop for us (https://mailchi.mp/319511d27567/new-at-bolt-instant-load-for-bolt-checkout-and-address-validation-2767153?utm_source=Testers&utm_campaign=284d8426c8-EMAIL_CAMPAIGN_2018_12_27_07_26_COPY_02&utm_medium=email&utm_term=0_fb951463cd-284d8426c8-517057421 ) I believe we should call `configure` whenever form is ready. Please note configure call does not invoke any blocking logic or UI.

Whenever form is updated, we should check if form is ready. Then we can call configure, which will involve Ajax call. I'd assume for most of cases you usually don't edit form much after it's complete but I might miss something here.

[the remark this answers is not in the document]
mccottry: I now understand the priority is for UX and instant load is highly valued.  I certainly will correct standard checkout and payment only checkout with your specification of calling BoltCheckout.configure immediately.

Regarding Firecheckout and admin checkout, we are not concerned with `BoltCheckout.configure`'s blocking logic, but rather the blocking logic of the UI incurred by calling the native `validate` methods of FireCheckout and of the admin form.  In both cases, the UI is blocked and we are trying to avoid making additional calls to this method.  However, I can create a solution for this without having to make additional calls and honoring the goal of "instant load".

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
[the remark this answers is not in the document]
mccottry: Since Promise is no longer associated with modal open, `reload` is being removed.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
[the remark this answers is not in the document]
daisy1754: hmm I don't see need for calling `validate` before calling BoltCheckout.configure. You can call configure when order is ready from Bolt's standpoint.

How about following:
 - we resister listener to detect change in fields that Bolt cares about (eg., email, name, or street address)
 - when listener is called, we check if all required fields from Bolt perspective is ready
 - if so we call BoltCheckout.configure with promise (this will mean Ajax call)
 - when user actually click checkout button, within check callback, we call framework's validate method

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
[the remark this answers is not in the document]
matt-thomason: My concern was for a line of code that read `return (isAdminFormValid = is_valid);`, not for assignment returns in general. If these variables are not being used elsewhere, it's much easier to read `return is_valid` throughout. Unless I'm missing something, `return is_valid` is equivalent to `return XXXX = is_valid` unless `XXXX` is passed by reference.

I don't mean to upend a coding convention that you have all been using though, so the previous style is fine. It's better than `if (!is_valid) return false;`.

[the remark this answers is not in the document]
mccottry: @matt-thomason 
1.) `isAdminFormValid` was used at a later time outside of the function.  It has global scope, so, yes, it is a reference to something global and has purpose

2.) `return is_valid` is not equivalent to `if (!is_valid) return false;`.  We only want to return in the case of false.  If true, we keep processing.  `if (condition_is_false) return false;` is the way to go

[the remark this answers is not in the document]
matt-thomason: Well, I don't know how I goofed on that. You're right 👍

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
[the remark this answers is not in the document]
mccottry: The need for not creating 1000s of extra quotes in the context of Product Page loads coupled with the minimal difference in time for Bolt to appear outweigh any minute gains in instant loan in the Product Page and Firecheckout context.  Instant load is not a priority in admin order creation.  Instant load will be respected in the shopping cart and the one-page-checkout scenario by calling BoltCheckout.configure immediately.

PR #267, opened 2018-12-16
Ensure a valid billing address is present

mccottry:
https://app.asana.com/0/544708310157130/942487273563864

This removes hack piece-mill logic of taking parts of the shipping address so that if any critical part of the billing address is missing, a true valid address is applied.

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/BoltOrder.php
+++ b/app/code/community/Bolt/Boltpay/Model/BoltOrder.php
@@ -26,8 +26,10 @@ class Bolt_Boltpay_Model_BoltOrder extends Mage_Core_Model_Abstract
     const ITEM_TYPE_PHYSICAL = 'physical';
     const ITEM_TYPE_DIGITAL  = 'digital';
 
-    // Store discount types, internal and 3rd party.
-    // Can appear as keys in Quote::getTotals result array.
+    /**
+     * @var array  Store discount types, internal and 3rd party.
+     *             Can appear as keys in Quote::getTotals result array.
+     */
     protected $discountTypes = array(
         'discount',
         'giftcardcredit',
@@ -39,6 +41,13 @@ class Bolt_Boltpay_Model_BoltOrder extends Mage_Core_Model_Abstract
         'amgiftcard', // https://amasty.com/magento-gift-card.html
         'amstcred', // https://amasty.com/magento-store-credit.html
         'awraf',    //https://ecommerce.aheadworks.com/magento-extensions/refer-a-friend.html#magento1
+    );
+
+    /**
+     * @var array  list of country codes for which Bolt requires a region
+     */
+    protected $countriesRequiringRegion = array(
+        'US', 'CA',
     );
 
     /**
@@ -175,14 +184,11 @@ class Bolt_Boltpay_Model_BoltOrder extends Mage_Core_Model_Abstract
         } else {
 
             $billingAddress  = $quote->getBillingAddress();
+            $this->correctBillingAddress($quote);
+
             $shippingAddress = $quote->getShippingAddress();
 
             $customerEmail = $this->getCustomerEmail($quote);
-
-            $billingRegion = $billingAddress->getRegion();
-            if (empty($shippingRegion) && !in_array($billingAddress->getCountry(), array('US', 'CA'))) {
-                $billingRegion = $billingAddress->getCity();
-            }
 
             ///////////////////////////////////////////
             // Include billing address info if defined.
@@ -196,7 +202,7 @@ class Bolt_Boltpay_Model_BoltOrder extends Mage_Core_Model_Abstract
                     'first_name'      => $billingAddress->getFirstname(),
                     'last_name'       => $billingAddress->getLastname(),
                     'locality'        => $billingAddress->getCity(),
-                    'region'          => $billingRegion,
+                    'region'          => $billingAddress->getRegion(),
                     'postal_code'     => $billingAddress->getPostcode() ?: '-',
                     'country_code'    => $billingAddress->getCountry(),
                     'phone'           => $billingAddress->getTelephone(),
@@ -343,6 +349,55 @@ class Bolt_Boltpay_Model_BoltOrder extends Mage_Core_Model_Abstract
 
         // otherwise, we have no better thing to do than let the Bolt server do the checking
         return $magentoDerivedCartData;
+    }
+
+    /**
+     * Checks if billing address of a quote has all the expected fields.
+     * If not, the address is updated by using the provided shipping address.
+     *
+     * @param Mage_Sales_Model_Quote $quote Quote that is expected to contain both
+     *                                      a billing and shipping address.
+     *
+     * @return bool  true if a correction is made, otherwise false
+     */
+    public function correctBillingAddress($quote)
+    {
+        $billingAddress = $quote->getBillingAddress();
+        $wasCorrected = false;
+
+        /////////////////////////////////////////////
+        /// Changes persisted to Magento and Bolt
+        /////////////////////////////////////////////
+        if (
+            !$billingAddress->getStreet() ||
+            !$billingAddress->getCity() ||
+            !$billingAddress->getCountry()
+        )
+        {
+            $billingAddress->setData($quote->getShippingAddress()->getData());
+            $billingAddress
+                ->setAddressType(Mage_Sales_Model_Quote_Address::TYPE_BILLING)
+                ->save();
+            $quote->save();
+            $wasCorrected = true;
+        }
+        /////////////////////////////////////////////
+
+        /////////////////////////////////////////////
+        /// Changes persisted only to Bolt
+        /////////////////////////////////////////////
+        if (!$billingAddress->getRegion() && !in_array($billingAddress->getCountry(), $this->countriesRequiringRegion)) {
+            $billingAddress->setRegion($billingAddress->getCity());
+            $wasCorrected = true;
+        }
+
+        if (!$billingAddress->getPostcode()) {
+            $billingAddress->setPostcode('-');
+            $wasCorrected = true;
+        }
+        /////////////////////////////////////////////
+
+        return $wasCorrected;
     }
 
     /**
--- a/app/code/community/Bolt/Boltpay/Model/ShippingAndTax.php
+++ b/app/code/community/Bolt/Boltpay/Model/ShippingAndTax.php
@@ -85,24 +85,7 @@ class Bolt_Boltpay_Model_ShippingAndTax extends Mage_Core_Model_Abstract
         }
 
         $quote->getShippingAddress()->addData($addressData)->save();
-
-        $billingAddress = $quote->getBillingAddress();
-
-        $quote->getBillingAddress()->addData(
-            array(
-                'email' => $billingAddress->getEmail() ?: ($shippingAddress->email ?: $shippingAddress->email_address),
-                'firstname' => $billingAddress->getFirstname() ?: $shippingAddress->first_name,
-                'lastname' => $billingAddress->getLastname() ?: $shippingAddress->last_name,
-                'street' => implode("\n", $billingAddress->getStreet()) ?: $shippingAddress->street_address1 . ($shippingAddress->street_address2 ? "\n" . $shippingAddress->street_address2 : ''),
-                'company' => $billingAddress->getCompany() ?: $shippingAddress->company,
-                'city' => $billingAddress->getCity() ?: $shippingAddress->locality,
-                'region' => $billingAddress->getRegion() ?: $region,
-                'region_id' => $billingAddress->getRegionId() ?: $regionId,
-                'postcode' => $billingAddress->getPostcode() ?: $shippingAddress->postal_code,
-                'country_id' => $billingAddress->getCountryId() ?: $shippingAddress->country_code,
-                'telephone' => $billingAddress->getTelephone() ?: ($shippingAddress->phone ?: $shippingAddress->phone_number)
-            )
-        )->save();
+        Mage::getModel('boltpay/boltOrder')->correctBillingAddress($quote);
 
         return $addressData;
     }


app/code/community/Bolt/Boltpay/Model/BoltOrder.php
```diff
@@ -41,6 +43,13 @@ class Bolt_Boltpay_Model_BoltOrder extends Mage_Core_Model_Abstract
         'awraf',    //https://ecommerce.aheadworks.com/magento-extensions/refer-a-friend.html#magento1
     );
 
+    /**
+     * @var array  list of country codes for which Bolt requires a region
```
daisy1754: IIRC, it was magento that was complaining for missing region

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
```diff
@@ -345,6 +351,80 @@ protected function getCorrectedTotal($projectedTotal, $magentoDerivedCartData)
...
     }
 
+    /**
+     * Checks if billing address of a quote has all the expected fields.
+     * If not, the address is updated by using the provided shipping address.
+     *
+     * @param Mage_Sales_Model_Quote $quote Quote that is expected to contain both
+     *                                      a billing and shipping address.
+     *
+     * @return bool  true if a correction is made, otherwise false
```
daisy1754: how about "true if billing address was updated"?

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
```diff
@@ -345,6 +351,80 @@ protected function getCorrectedTotal($projectedTotal, $magentoDerivedCartData)
...
+     */
+    public function correctBillingAddress($quote)
+    {
+        $billingAddress = $quote->getBillingAddress();
+        $shippingAddress = $quote->getShippingAddress();
+        $wasCorrected = false;
+
+        /////////////////////////////////////////////
+        /// Changes persisted to Magento and Bolt
```
daisy1754: why some change is saved to magento while other is not?

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
```diff
@@ -345,6 +351,80 @@ protected function getCorrectedTotal($projectedTotal, $magentoDerivedCartData)
...
+        /////////////////////////////////////////////
+        if (
+            !$billingAddress->getStreet() ||
+            !$billingAddress->getCity() ||
+            !$billingAddress->getCountry()
+        )
+        {
+            $oldPrefix = $billingAddress->getPrefix();
```
daisy1754: rename this and below to $billingFoo

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
```diff
@@ -345,6 +351,80 @@ protected function getCorrectedTotal($projectedTotal, $magentoDerivedCartData)
...
+            $oldFirstName = $billingAddress->getFirstname();
+            $oldMiddleName = $billingAddress->getMiddlename();
+            $oldLastName = $billingAddress->getLastname();
+            $oldSuffix = $billingAddress->getSuffix();
+
+            $billingAddress->setData($shippingAddress->getData());
```
daisy1754: why replace all fields, not only street, city, country/?

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
```diff
@@ -345,6 +351,80 @@ protected function getCorrectedTotal($projectedTotal, $magentoDerivedCartData)
...
+            $wasCorrected = true;
+        }
+
+        if (!$billingAddress->getName()) {
+            $billingAddress
+                ->setPrefix($shippingAddress->getPrefix())
+                ->setFirstname($shippingAddress->getFirstname())
+                ->setMiddlename($shippingAddress->getMiddlename())
+                ->setLastname($shippingAddress->getLastname())
+                ->setSuffix($shippingAddress->getSuffix());
```
daisy1754: not save?

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
```diff
@@ -345,6 +351,80 @@ protected function getCorrectedTotal($projectedTotal, $magentoDerivedCartData)
...
+    /**
+     * Checks if billing address of a quote has all the expected fields.
+     * If not, the address is updated by using the provided shipping address.
+     *
+     * @param Mage_Sales_Model_Quote $quote Quote that is expected to contain both
+     *                                      a billing and shipping address.
+     *
+     * @return bool  true if a correction is made, otherwise false
+     */
+    public function correctBillingAddress($quote)
```
daisy1754: this method needs unit test

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
matt-thomason: I agree. If there's a specific reason it should be onlined why in a comment.

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
```diff
@@ -175,15 +184,12 @@ function ($item) use ($quote, &$calculatedTotal, $boltHelper) {
         } else {
 
             $billingAddress  = $quote->getBillingAddress();
+            $this->correctBillingAddress($quote);
```
matt-thomason: Shouldn't we correct the billing address before getting it from `quote`?

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
mccottry: @daisy1754 Let me clarify the context/motivation for this PR -

Context:
While Bolt nor Magento technically require a Billing Address for an order, certain ERP systems do.  A request came in from Indique to include billing information in all orders.

However, to address 

> IIRC, it was magento that was complaining for missing region

As I remember it, the problem with region was that Bolt required a region on billing while Magento did not for countries like Bermuda where there are no regions.  This was resolved by using the city as the region to send to Bolt.  While this is related to this PR, it is not the same.

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
mccottry: seems the same info to me.  I'll change as you have requested.

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
mccottry: ok.  I'll add more comments.

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
mccottry: ok

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
mccottry: Actually, on second thought, `update` to me implies a change to the database while corrected does not necessarily give this connotation.  Corrected seems like a better choice for this reason.

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
mccottry: @daisy1754 
I'm not sure what you are asking here as in which variable you are asking be named billingFoo. 
 Also, $billingFoo is a very ambiguous variable name, and very confusing.  If you are talking about using the metasyntactic variable convention FooBar, I will point out in zero places in the entire Magento code place that uses this psuedo variable name and it seems to be a strange convention to be adding here.

I believe that I must be confused with the request here.

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
mccottry: yes.  This was a mistake that I made after changing the parameter from the billing address object to the quote object.  I forgot to change the order.

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
mccottry: saved.

mccottry:
@daisy1754 @matt-thomason Thank you so very much for this thorough review.  I am always striving to cover more edge cases and to be less blinded by my confidence and these reviews are really helpful in ever improving my coding approach.

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
daisy1754: Hmm I don't think we validate region. I just tried calling our API and was able to create order w/o region:

```
curl -X POST \
  https://api-sandbox.bolt.com/v1/merchant/orders \
  -H 'Content-Type: application/json' \
  -H 'Origin: http://bolt.com' \
  -H 'X-Api-Key: c117fec29f5c1365dacbc62390b5b6f98c0698a812d7015849319f9f7b863064' \
  -H 'X-Nonce: 123456789123' \
  -H 'cache-control: no-cache' \
  -d '{
  "cart": {
    "total_amount": 110,
    "discounts": [{
      "description": "testdiscount",
      "amount": 250
    }],
    "currency": "USD",
    "items": [{
      "name": "Pizza Slice",
      "reference": "",
      "description": "Test description",
      "sku": "1234567890",
      "total_amount": 360,
      "unit_price": 60,
      "currency": "USD",
      "quantity": 6,
      "options": null,
      "tax_amount": 0
    }],
    "billing_address": {
      "street_address1": "Test",
      "street_address2": "",
      "locality": "San Francisco",
      "country": "United States",
      "country_code": "US",
      "first_name": "Kazuki",
      "last_name": "Nishiura",
      "company": null,
      "phone": "1111111111",
      "email": "kazuki@bolt.com"
    }
  }
}'
```

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
```diff
@@ -345,6 +351,87 @@ protected function getCorrectedTotal($projectedTotal, $magentoDerivedCartData)
...
+        $quote = $billingAddress->getQuote();
+
+        $wasCorrected = false;
+
+        /////////////////////////////////////////////
+        /// Changes persisted to Magento and Bolt
+        /////////////////////////////////////////////
+        if ($fallbackAddress) {
+            if (
+                !trim($billingAddress->getStreetFull()) ||
```
daisy1754: why these can be empty in the first place?

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
```diff
@@ -345,6 +351,87 @@ protected function getCorrectedTotal($projectedTotal, $magentoDerivedCartData)
...
+            !$billingAddress->getRegion() &&
+            !in_array($billingAddress->getCountry(), $this->countriesRequiringRegion)
+        )
+        {
+            $billingAddress->setRegion($billingAddress->getCity());
+            $wasCorrected = true;
+        }
+
+        if (!$billingAddress->getPostcode()) {
+            $billingAddress->setPostcode('-');
```
daisy1754: is postal code not required by magento?

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
mccottry: @daisy1754 
In the past and in the current documentation, region was required:
https://docs.bolt.com/v1/reference#create-order

Apparently, it is not now, but it seems like this is a mistake because whereas there is only one San Francisco in the US which is in California, there are several Newtown[s] in the US in several states.

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
mccottry: @daisy1754 That's the million dollar question and I can answer it partially.  Initially, when any part of the billing was missing, the old code removed it entirely.  This is one way in which this use to happen.  We have fixed that problem.

A second way is when there is a store which is selling virtual products and we do not collect address information because of this.

A third way is if there are custom address entry forms or routines which for whatever reason, does not enforce billing address data and only is concerned with shipping data.  In Europe, for instance, it is common that billing address data is not associated with the credit card.

And finally, there maybe situations that currently I can't explain.  Indique is one of them.  It is quite possible that it was because of the first reason and Indique not updating their plugin, but because of the second two reasons listed, I don't see any harm in checking for this here.

What makes all this possible is that in the code, we turn off address validation for Bolt orders on order creation:

https://github.com/BoltApp/bolt-magento1/blob/develop/app/code/community/Bolt/Boltpay/Model/Order.php#L124-L125

This has been a long standing practice and I am not challenging that decision in this pull request, but it is a question to ask:  should we be conditionally validating these addresses?

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
mccottry: @daisy1754 
This is a good question.  This varies from store to store depending on configurations.  See these admin pages below:

**Configurable postal code requirement **
image

**Configurable region requirement**
image

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
daisy1754: hmm, sounds like for 2 and 3, billing address is intended to be blank. are we certain billing address is still missing some field unintentionally with latest plugin?  I'd like to avoid code (which we need to maintain) if it's not needed.

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
```diff
@@ -345,6 +351,87 @@ protected function getCorrectedTotal($projectedTotal, $magentoDerivedCartData)
...
+        }
+        if ($wasCorrected) $quote->save();
+        /////////////////////////////////////////////
+
+        /////////////////////////////////////////////
+        /// Changes persisted only to Bolt
+        /// These are workarounds that deal with the
+        /// Bolt requirements for fields that are not
+        /// required by Magento and should not be
+        /// persisted in Magento
```
daisy1754: Mutating address object but not saving to magento seems very weird to me. If some code in code path calls $address->save later, these change will be written to database.

Given we only want to change data for $cartSubmissionData, let's set fallback in where we construct $cartSubmissionData ?

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
daisy1754: If you prefer to keep these, let's set bugsnag.

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
daisy1754: let's remove comment 'list of country codes for which Bolt requires a region' since it's not true. I agree Bolt should probably add validation on region but as of right now this comment is not true.

mccottry: done.

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
mccottry: @daisy1754 
Can we set it in bugsnag and schedule the code for removal if it is not triggered in bugsnag after a single quarter, let's say end of March?

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
mccottry: @daisy1754 makes sense.

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
daisy1754: sounds good. let's create asana task to track code removal

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
```diff
@@ -345,6 +363,89 @@ protected function getCorrectedTotal($projectedTotal, $magentoDerivedCartData)
...
+        $bugsnag = Mage::helper('boltpay/bugsnag');
+
+        $quote = $billingAddress->getQuote();
+
+        $wasCorrected = false;
+
+        /////////////////////////////////////////////
+        /// Changes persisted to Magento and Bolt
+        /////////////////////////////////////////////
+        if ($fallbackAddress) {
```
daisy1754: nit: can you early return here?

```
if (!$fallbackAddress) {
  return false;
}
```

mccottry: done

PR #268, opened 2018-12-21
Add pre-commit hook for checking PHP Lint and MEQP1 style checking

matt-thomason:
Here is my first pass at githooks to check linting and style. After pulling this branch, all you need to do is a `composer install` and you're set up. You may need to enter magento credentials the first time you `composer install` (to fetch the `MEQP1` standard).

When you commit, the following happen in order:

1. A basic lint is performed using `php -l`. If this fails, your commit is aborted.
2. A style check against `MEQP1` is performed using `phpcs`. This checks for both errors and warnings.
3. You are given an (optional) prompt to automatically fix warnings and errors.
4. `phpcs` is run again, only checking errors. If no errors are present, your commit goes through - if errors still exist, your commit is aborted.

Please find some screenshots attached with examples:

**PHP Lint fails, abort immediately**
screen shot 2018-12-21 at 9 19 53 am

**PHPCS finds and fixes warnings, then commits**
screen shot 2018-12-21 at 9 18 16 am

**PHPCS finds and fixes warnings, but cannot fix an error, so aborts**
screen shot 2018-12-21 at 9 17 20 am

Let me know if you have any questions or concerns!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .DS_Store
 .idea
+/vendor
--- a/tools/setup.sh
+++ b/tools/setup.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+
+cp tools/pre-commit .git/hooks/pre-commit
+chmod +x .git/hooks/pre-commit
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,14 @@
     "ext-PDO": "*"
   },
   "require-dev": {
+    "squizlabs/php_codesniffer": "3.*",
+    "magento/marketplace-eqp": "2.*",
     "phpunit/phpunit": "^4.8"
+  },
+  "scripts": {
+    "post-install-cmd": [
+      "bash tools/setup.sh"
+    ]
   },
   "repositories": [
     {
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 .DS_Store
-.idea
\ No newline at end of file
+.idea
+/vendor
```
matt-thomason: I'm adding the vendor file to the `.gitignore`. We should also decide what we want to do with our `composer.lock` files across all our projects. In most projects I've worked on, companies usually commit their `composer.lock` files so that all developers are on the same versions across all development computers. If we don't want to do that (so that we're always pulling something when we deploy, for example), then we should make the decision to add `composer.lock` to the `.gitignore`. What are your thoughts @daisy1754 ?

tools/setup.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+cp tools/pre-commit .git/hooks/pre-commit
```
matt-thomason: Warning! If anyone has an existing `pre-commit` hook, it will be overridden.

matt-thomason:
I will make similar PRs to M2 and eventually Woo if this method jives with everyone. I'll only make this PR first for simplicity.

.gitignore
daisy1754: I agree. I think we should add composer.lock to git

PR #290, opened 2019-02-03
Shipping applied event

mccottry:
https://app.asana.com/0/544708310157130/1107922957237577

Adds an event for when shipping has been applied to a quote just before it is converted to an order


app/code/community/Bolt/Boltpay/Model/Observer.php
```diff
@@ -110,12 +110,27 @@ public function verifyOrderContents($observer)
...
+    * @param $observer  observer whose event object holds the 'quote' and the
+    *                   'shippingMethodCode'
+    *
+    */
+    public function doAfterShippingMethodApplied($observer)
+    {
+        /* @var Mage_Sales_Model_Quote $quote */
+        $quote = $observer->getEvent()->getQuote();
+        /* @var string $shippingMethodCode */
+        $shippingMethodCode = $observer->getEvent()->getShippingMethodCode();
```
johnmccombs1: Is this just to give us a method to override later? Just curious since the local variable is being set but nothing is done with it.

mccottry: @johnmccombs1  Correct, I intended to add in the comments above that this is a stub class with the purpose of strongly advertising to use this method in the future and illustrating how to use it.  Where as it might be obvious for us, there will come a time when future developers will find this information to be quite useful.  The comment is being added.

@daisy1754 
> can you explain the use case? curious if we can instead dispatch something like onBeforeSaveOrder

I DM'd you the specific use case that motivated this implementation.  `onBeforeSaveOrder` already exist and does not cover this scenario.

PR #294, opened 2019-02-05
Update composer vendor name to boltpay

daisy1754:


matt-thomason:
Merging this now - in case @daisy1754 is busy today 🎆

PR #295, opened 2019-02-06
Add generic event filter construct

mccottry:
https://app.asana.com/0/544708310157130/1108368077993428

A generic filter event is need for filtering strings, arrays, objects, etc. such as javascript, product descriptions, prices, address data, etc.


mccottry:
> @mccottry can you give me a little more information about the use case for this. The code looks good but I'm not exactly sure why we need it.

@johnmccombs1 Please read the linked Asana task.  You will also get a pull request shortly with an immediate concrete use case.

PR #297, opened 2019-02-07
Use discount as positive since Bolt uses positive

mccottry:
Sum of discounts sent to Bolt was not accounted for as a positive number.  This revises Bolt totals calculations treating discounts as a positive number like it is done on the Bolt server.


chaseadams509:
Please let me know once this gets push to dev branch and on sandbox so I can run integration tests.

PR #299, opened 2019-02-09
[WIP] Collects shipping totals on order creation

mccottry:
https://app.asana.com/0/544708310157130/1108910545116701

Always collects shipping totals on order creation to assure shipping methods are there to pass validation.

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Order.php
+++ b/app/code/community/Bolt/Boltpay/Model/Order.php
@@ -123,11 +123,18 @@ class Bolt_Boltpay_Model_Order extends Mage_Core_Model_Abstract
                 $shippingMethodCode = $transaction->order->cart->shipments[0]->reference;
             }
 
-            $immutableQuote->getShippingAddress()->setShouldIgnoreValidation(true)->save();
             $immutableQuote->getBillingAddress()->setShouldIgnoreValidation(true)->save();
 
+            /** @var Mage_Sales_Model_Quote_Address $shippingAddress */
+            $shippingAddress = $immutableQuote
+                ->getShippingAddress()
+                ->setCollectShippingRates(true)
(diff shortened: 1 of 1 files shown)


app/code/community/Bolt/Boltpay/Model/Order.php
```diff
@@ -123,11 +123,18 @@ public function createOrder($reference, $sessionQuoteId = null, $isAjaxRequest =
...
             }
 
-            $immutableQuote->getShippingAddress()->setShouldIgnoreValidation(true)->save();
             $immutableQuote->getBillingAddress()->setShouldIgnoreValidation(true)->save();
 
+            /** @var Mage_Sales_Model_Quote_Address $shippingAddress */
+            $shippingAddress = $immutableQuote
+                ->getShippingAddress()
+                ->setCollectShippingRates(true)
+                ->collectShippingRates()
```
johnmccombs1: I seem to recall that there was a reason that we didn't want to do this but I can't remember now. Something about shipping method rates not being the same sometimes as they were at the time that the customer selected the rate, or sometimes the 3rd party fails to return rates at this step. This shouldn't happen of course, but it does leave us open to an additional potential discrepancy. Do we know why sometimes the shipping rates go missing? I'm good with this solution but there might be some fallout.

mccottry: @johnmccombs1 I thought about this an hour or two while implementing the code and there are several factors to consider.  Still, I expected and I am glad for your response as there are several things in the release implementation and in this solution that leave me feeling uneasy.  For this reason, I am going to change this to a WIP and reopen the pull request after rewriting and discussing the problems with you and the M1 team offline.

mccottry:
Handle via PR #304

PR #303, opened 2019-02-11
Ensure customer is logged in when saving the order

johnmccombs1:
Asana task: https://app.asana.com/0/564264490825835/1108525009378275

### DIFF
--- a/app/code/community/Bolt/Boltpay/controllers/OrderController.php
+++ b/app/code/community/Bolt/Boltpay/controllers/OrderController.php
@@ -63,23 +63,10 @@ class Bolt_Boltpay_OrderController extends Mage_Core_Controller_Front_Action
             $transactionHelper = Mage::helper('boltpay/transaction');
             /** @var  Bolt_Boltpay_Model_Order $orderModel */
             $orderModel = Mage::getModel('boltpay/order');
-            $immutableQuoteId = $transactionHelper->getImmutableQuoteIdFromTransaction($transaction);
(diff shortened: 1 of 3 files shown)


app/code/community/Bolt/Boltpay/controllers/OrderController.php
```diff
@@ -63,10 +63,23 @@ public function saveAction()
             $transactionHelper = Mage::helper('boltpay/transaction');
             /** @var  Bolt_Boltpay_Model_Order $orderModel */
             $orderModel = Mage::getModel('boltpay/order');
-            $order = $orderModel->getOrderByQuoteId($transactionHelper->getImmutableQuoteIdFromTransaction($transaction));
+            $immutableQuoteId = $transactionHelper->getImmutableQuoteIdFromTransaction($transaction);
+            $order = $orderModel->getOrderByQuoteId($immutableQuoteId);
 
             if ($order->isObjectNew()) {
-                $orderModel->createOrder($reference, $checkoutSession->getQuoteId(), true, $transaction);
+                $sessionQuoteId = $checkoutSession->getQuoteId();
```
mccottry: @johnmccombs 
This is a great insight.  I was thinking about another error that I am seeing that comes via orphaned transactions where the cause could be that the customer is not logged in.  For this reason, it will be better if you move this code to `Bolt_Boltpay_Model_Order::createOrder`

Can you make this change as I would like to use this code A.S.A.P?  I'm hoping this morning if possible.

PR #310, opened 2019-02-19
Allows merchant to specify quote cloning/Bolt order creation on Bolt button click

mccottry:
https://app.asana.com/0/941895179897714/1110284468674223/f

The config is an `Extra config` boolean value called `cloneOnClick`.  

Example usage to enable cloning on Bolt button click:

In the Magento Admin `System -> Configuration -> Payment Methods -> Bolt Pay -> Extra Config`
```
{
"boltPrimaryColor":"#fff000",
"hintsTransform":
"
     function(hints){
          hints.prefill = 
               {\"email\":\"leon@bolt.com\"};
          return hints;
     }
",
"cloneOnClick": true
}
```

### DIFF
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -212,6 +212,7 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
         $onCheckCallback = $boltHelper->buildOnCheckCallback($checkoutType, $quote);
 
         $hintsTransformFunction = $boltHelper->getExtraConfig('hintsTransform');
+        $shouldCloneImmediately = !$boltHelper->getExtraConfig( 'cloneOnClick' );
 
         $boltConfigureCall =
         "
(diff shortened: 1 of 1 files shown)


mccottry:
> It seems as though this is the kind of thing that is better suited for being an Admin panel option, but this is sufficient for now for pushing out a fix for L&G

@matt-thomason This is an admin panel option via the extra config section.  If you are saying that it should be a more highly visible configuration option, I disagree.  To me, this falls under the category of an option that is concerned with low-level code implementation details that would be confusing to most merchants and  that we really don't want to be broadcasting.  You can view this as an "advanced configuration" that users should change only if they know what they are doing and understand the implications of that action.  General configurations should be simple and self-explanatory to merchants and does not contradict general Bolt goals like branding or advertised features.  To the average merchant, the explanation of this feature is neither simple, nor is is 100% in tune with https://bit.ly/2SKFrfk

For these reasons, it was placed as extra config and should be set by integration engineers when they deem it appropriate.

That being said, even without this feature enable, Bolt Everywhere Bolt order creation deferred to after button click via Promise and cart order token caching would solve aforementioned merchant's problem sufficiently.  This just takes things one step more further because this merchant specifically asked about the low-level details.

matt-thomason:
> If you are saying that it should be a more highly visible configuration option, I disagree. To me, this falls under the category of an option that is concerned with low-level code implementation details that would be confusing to most merchants and that we really don't want to be broadcasting.

I think we both agree on this - apologies if I was unclear. I meant that this would be a good option in the admin.bolt.com panel, hiding it from merchants completely.

mccottry:
> > If you are saying that it should be a more highly visible configuration option, I disagree. To me, this falls under the category of an option that is concerned with low-level code implementation details that would be confusing to most merchants and that we really don't want to be broadcasting.
> 
> @mccottry I think we both agree on this - apologies if I was unclear. I meant that this would be a good option in the admin.bolt.com panel, hiding it from merchants completely.

That's an interesting proposition.

PR #318, opened 2019-02-22
Fix extra config issue when the value is empty string

dumega:


app/code/community/Bolt/Boltpay/Model/Admin/ExtraConfig.php
```diff
@@ -61,6 +61,13 @@ class Bolt_Boltpay_Model_Admin_ExtraConfig extends Mage_Core_Model_Config_Data
      */
     public function save()
     {
+        // when json_decode an empty string, PHP7 will trigger an Syntax error, so directly return without json_decode for empty string value
```
matt-thomason: We will still have the issue in function `getExtraConfig` lower down. Can we move this and the `json_decode` portion into `normalizeJSON`, rename it to `normalizeExtraConfig` or something, and make so we have a simple `$optionsJson = $this->normalizeExtraConfig($this->getValue); // Returns empty array if passed empty string`?

dumega: @matt-thomason 

1. If the value is an empty string, it means the merchant want to remove any extra config setting, and based on current logic,  when $optionsJson = $this->normalizeExtraConfig($this->getValue);
whatever it get an empty array or string, the config value would be reset to old value, also it is hard to detect whether it is a json error or real value in such a case.

2. On the other hand, in the function `getExtraConfig`, if the Mage::getStoreConfig('payment/boltpay/extra_options') return an empty string, `$allExtraConfigs` would get an empty array, then $rawValue would be assigned an empty string too. So json_decode an empty string wont cause any exception or issue in this function.

matt-thomason: Sorry you are right - I missed this nuance.

dumega:
@mccottry if so, once the merchant input anything on this field and saved into configuration, there is no chance to update it to empty. https://github.com/BoltApp/bolt-magento1/blob/fe47e9df975e46d4d353c837e5798765461b1b7d/app/code/community/Bolt/Boltpay/Model/Admin/ExtraConfig.php#L87

mccottry:
> @mccottry if so, once the merchant input anything on this field and saved into configuration, there is no chance to update it to empty.
> 
> [bolt-magento1/app/code/community/Bolt/Boltpay/Model/Admin/ExtraConfig.php](https://github.com/BoltApp/bolt-magento1/blob/fe47e9df975e46d4d353c837e5798765461b1b7d/app/code/community/Bolt/Boltpay/Model/Admin/ExtraConfig.php#L87)
> 
> Line 87 in [fe47e9d](/BoltApp/bolt-magento1/commit/fe47e9df975e46d4d353c837e5798765461b1b7d)
> 
>          if (!$areAllOptionsValid) {

@dumega The current logic will absolutely allows for empty strings to be entered.  That is why php 5.6 was not throwing an error for anyone else.

A light touch solution can be found here:
https://github.com/BoltApp/bolt-magento1/pull/319/files

dumega:
@mccottry  yes, you are right, let me close this PR.

PR #323, opened 2019-02-24
Magento 1 v2.0

mccottry:
This code enables Pre-Auth order creation in the M1 context

List of changes include:
- create_order webhook endpoint implemented
- new bolt error code handling for pre-auth order creation
- custom bolt preauth statuses added
- **several unit test updated for pre-auth**
- payment failure handling implemented
- Hides pre-auth orders from admin order list
- clear cart cache on pre-auth order cancel
- does not validate apple pay addresses during shipping and tax calc
- **more updating and fixing unit test**
- grand totals check prior to final save
- safe guards against other plugins changing from pre-auth statuses
- moving front session events only to front end
- Adds pending_bolt and canceled_bolt order cleanup logic to every 15 minutes
- fix for firecheckout not interfering with click events + simultaneous order c…  …
- accounts concurrent pre-auth order request
- fixes to allow for Bolt to enable upon new store installation

_commits to various changes are listed below and can be reviewed separately at the commit level_


mccottry:
> can you please add PR description with list of changes? Also if you can split this into multiple PRs that would be great (eg., maybe `OrderCreationException.php` can be its own PR?)
@daisy1754 
Splitting this into multiple PRs is not possible or productive at this phase with the demands of getting this live.  It would serve no benefit.  Most important is that unit test and integration test pass.

mccottry:
> also can you add unit test?

@daisy1754 

No problem with adding some unit test.  The most beneficial test are not unit test, but are the integration test to test error codes. 

@danielatbolt, @daisy1754 , @matt-thomason 
Good unit test take a lot of extra side effort and are much less valuable the the same code being in release on Energy Conscious and GrowersHouse.

We know the need for unit test exists and we know that we are going to have a 2.0.1, which is more likely to be the official public release.  I move that we separate the creation unit test from this PR.  Unit test can be done separately which does not continue to block merging this back to develop which in turning is blocking DataDog and several other initiatives.

mccottry:
> @mccottry I'm not against putting "beta" code in a small set of merchant as long as you test it extensively on merchant's environment (which I believe you did for energy concisous), but I don't want us to merge code without proper review. I don't know how soon we want to push the version update, but 2000LoC makes quick review almost impossible.
> 
> At quick glance, I see following changes that I think could be pulled into separate PR
> 
> * adding error class
> * adding API trait
> * cron to cleanup order resources
> * update to CouponHelper to make test stable
> * new events (eg., bolt_boltpay_validate_coupons_before)
> 
> WDYT? If you afraid breaking things accidentally, you can keep this PR as is and cherry-pick portions into separate branches one-by-one.

Absolutely @daisy1754 ! We must give this a proper review.  And thank you very much for making me aware to look even deeper and itemize all the things which had to change to make this a cohesive whole.  I'll certainly extend the list of sub-topics. 

Unfortunately, this type of change was never going to be a quick review and it shouldn't be reviewed quickly. The 2000 plus lines of code change to make pre-auth happen definitely is more than a little bit.  But it is only a little bit more than the 2000 plus lines of code in the approved pull request found here for the same WooCommerce project:

https://github.com/BoltApp/bolt-woocommerce/pull/374/files

I think that set a superb precedence for PRs for the pre-auth project.  So, my hope is that we can get this reviewed in the same way, as a whole, and then work in the future on how to break down large task for quick and easy reviews.  We can also have as many round table discussions as necessary to work through the code, which I estimate will be much more efficient and much more rewarding than massive effort of trying to cherry pick this code that has evolved over such a long period of time.  Can we go ahead schedule a meeting for this?

daisy1754:
To be fair 900 LoC of the WooCommerce PR is around testing. I still think this PR contains lots of different topics:

 - adding error class
 - adding API trait
 - cron to cleanup order resources
 - update to CouponHelper to make test stable
 - new events (eg., bolt_boltpay_validate_coupons_before)

For instance addition of new events deserves separate PR and separate discussions IMO

PR #325, opened 2019-02-26
Allow invoices to be created for $0.00 orders

johnmccombs1:
Asansa task: https://app.asana.com/0/564264490825835/1111448577711715


johnmccombs1:
> Looks fine, but the asana task link doesn't work for me. Can you update it so that we can reference it if we need to in the future?

I fixed it. It was private. Thanks!

PR #328, opened 2019-03-07
Voiding after partial capture

johnmccombs1:
Asana task: https://app.asana.com/0/564264490825835/1106472851816631

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Payment.php
+++ b/app/code/community/Bolt/Boltpay/Model/Payment.php
@@ -445,14 +445,7 @@ class Bolt_Boltpay_Model_Payment extends Mage_Payment_Model_Method_Abstract
                     $order->setState(Mage_Sales_Model_Order::STATE_PAYMENT_REVIEW, true, $message);
                     $order->save();
                 } elseif ($newTransactionStatus == self::TRANSACTION_CANCELLED) {
-                    $order = $payment->getOrder();
-                    $payment->setParentTransactionId($reference);
-                    $payment->setTransactionId(sprintf("%s-void", $reference));
-                    $payment->addTransaction(Mage_Sales_Model_Order_Payment_Transaction::TYPE_VOID, null, true);
-                    $message = $this->boltHelper()->__('BOLT notification: Transaction authorization has been voided');
-                    $order->setState(Mage_Sales_Model_Order::STATE_CANCELED, true, $message);
-                    $payment->save();
-                    $order->save();
+                      $this->handleVoidTransactionUpdate($payment);
                 } elseif ($newTransactionStatus == self::TRANSACTION_REJECTED_IRREVERSIBLE) {
                     $order = $payment->getOrder();
                     $payment->setParentTransactionId($reference);
@@ -971,5 +964,56 @@ class Bolt_Boltpay_Model_Payment extends Mage_Payment_Model_Method_Abstract
     {
         return $newTransactionStatus == self::TRANSACTION_COMPLETED ||
               ($newTransactionStatus == self::TRANSACTION_AUTHORIZED && $prevTransactionStatus == self::TRANSACTION_AUTHORIZED);
+    }
+
+    /**
+     *  Handles two void cases sent by Bolt.
+     *  1.) Complete void an order cancelling all funds
+     *  2.) Partial void where partial capture has occurred but authorization has expired
+     *
+     * @param Mage_Payment_Model_Info $payment
+     */
+    public function handleVoidTransactionUpdate(Mage_Payment_Model_Info $payment){
+
+        $authTransaction = $payment->getAuthorizationTransaction();
+        $order = $payment->getOrder();
+        $amount =  $order->getBaseGrandTotal() - $order->getBaseTotalPaid() ;
+
+        if ($authTransaction->canVoidAuthorizationCompletely()) {
+            // True void
+            $reference = $payment->getAdditionalInformation('bolt_reference');
+            $payment->setParentTransactionId($reference);
+            $payment->setTransactionId(sprintf("%s-void", $reference));
+            $payment->setShouldCloseParentTransaction(true);
+            $transaction = $payment->addTransaction(Mage_Sales_Model_Order_Payment_Transaction::TYPE_VOID, null, true);
+            $message = $this->getVoidMessage($payment, $transaction, $amount);
+            $order->setState(Mage_Sales_Model_Order::STATE_CANCELED, true, $message);
+        } else if (!$authTransaction->getIsClosed()) {
+            // Open authorization has expired and partial capture has taken place.
+            // We do not change the order state.  We only need to close the authorization.
+            $authTransaction->closeAuthorization();
+            $message = $this->getVoidMessage($payment, $authTransaction, $amount);
+            $order->addStatusHistoryComment($message);
+        }
+
+        $order->save();
+    }
+
+    /**
+     * @param Mage_Payment_Model_Info $payment
+     * @param $transaction
+     * @param null $amount
+     * @return string
+     */
+    protected function getVoidMessage(Mage_Payment_Model_Info $payment, $transaction, $amount = null)
+    {
+        $order = $payment->getOrder();
+        $message = $this->boltHelper()->__('BOLT notification: Transaction authorization has been voided.');
+        if (isset($amount)) {
+            $message .= ' ' . $this->boltHelper()->__('Amount: %s.', $order->getBaseCurrency()->formatTxt($amount, array()));
+        }
+        $message .= ' ' . $this->boltHelper()->__('Transaction ID: "%s".', $transaction->getHtmlTxnId());
+
+        return $message;
     }
 }


app/code/community/Bolt/Boltpay/Model/Payment.php
```diff
@@ -975,4 +969,63 @@ protected function isCaptureRequest($newTransactionStatus, $prevTransactionStatu
         return $newTransactionStatus == self::TRANSACTION_COMPLETED ||
               ($newTransactionStatus == self::TRANSACTION_AUTHORIZED && $prevTransactionStatus == self::TRANSACTION_AUTHORIZED);
     }
+
```
mccottry: I have some serious reservations about this series of methods.  The primary case is very simple for cancelled orders and has been robustly handled from what I see from v0.0.1 some nearly two years ago written by Akshaya.  Adding this level of complexity makes very little sense.

Essentially, we are trying to force an exceptional type of webhook,  a "partial cancel", into a very discretely defined and firmly established logic for a separate case, "just cancel the order - no money collected".  This approach is bound to add new errors.  Can we try to handle this as an exceptional case like here?  

https://github.com/BoltApp/bolt-magento1/blob/develop/app/code/community/Bolt/Boltpay/controllers/ApiController.php#L192-L227

If we handle this as a Bolt_Boltpay_InvalidTransitionException taking into account the previous order state of completed and the new state of cancelled, then this PR can become very simple.  In that case, we would just return a 200 OK and optionally add an updated transaction to report the partial cancel.  It is much better to add a `$isCancelForCompletedOrder` condition, and if we feel it looks cleaner, make this as a single helper method called `canAssumeHookIsHandled` .

At the same time we preserve tried an true logic for the normal use case leaving what has been unchanged for nearly two years as unchanged, avoiding regression errors.  

I will point out some problems that I see in this PR logic below.

app/code/community/Bolt/Boltpay/Model/Payment.php
```diff
@@ -975,4 +969,63 @@ protected function isCaptureRequest($newTransactionStatus, $prevTransactionStatu
...
+        $payment->setTransactionId(sprintf("%s-void", $reference));
+        $payment->setShouldCloseParentTransaction(true);
+        $transaction = $payment->addTransaction(Mage_Sales_Model_Order_Payment_Transaction::TYPE_VOID, null, true);
+
+        $amount = $this->getVoidAmount($payment);
+        $message = $this->getVoidMessage($payment, $amount, $transaction);
+        $order = $payment->getOrder();
+
+        // update order state and add comments
+        $order->setState(isset($amount) ? Mage_Sales_Model_Order::STATE_CANCELED : Mage_Sales_Model_Order::STATE_PROCESSING, true, $message);
```
mccottry: The biggest problem with this logic is in the use case for which the code is introduced to solve - The order being `Completed`  and then `Cancel` webhook received.  In both cases, whether there was a partial capture/partial void or not, changing the order state is an error.

In the case of partial void, the logic says to switch the order to `Processing`  A completed order from the merchant perspective has already been shipped and requires no other processing.  If we switch an already completed order to processing, we are inviting the problem of the merchant reshipping an already shipped order.  This is not good.

In the case that the order is marked completed and the a cancel hook comes in and changes the status to canceled, then we are running the risk of ERP's and inventory getting out of sync.   

In both cases, ignoring the hook and just sending a 200 OK is preferred.  Remember, the only problem we are trying to fix is noise at the Bolt end.  Merchant-side everything is already resolved.

app/code/community/Bolt/Boltpay/Model/Payment.php
```diff
@@ -75,7 +75,7 @@ class Bolt_Boltpay_Model_Payment extends Mage_Payment_Model_Method_Abstract
 
     protected $_validStateTransitions = array(
         self::TRANSACTION_AUTHORIZED => array(self::TRANSACTION_AUTHORIZED, self::TRANSACTION_COMPLETED, self::TRANSACTION_CANCELLED, self::TRANSACTION_REJECTED_REVERSIBLE, self::TRANSACTION_REJECTED_IRREVERSIBLE, self::TRANSACTION_PENDING),
-        self::TRANSACTION_COMPLETED => array(self::TRANSACTION_REFUND, self::TRANSACTION_NO_NEW_STATE, self::TRANSACTION_COMPLETED),
+        self::TRANSACTION_COMPLETED => array(self::TRANSACTION_REFUND, self::TRANSACTION_NO_NEW_STATE, self::TRANSACTION_COMPLETED, self::TRANSACTION_CANCELLED),
```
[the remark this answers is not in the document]
danac-gs: @mccottry I think these are the transitions between Bolt transaction states, not the order ones, and are labeled correctly. I do not agree how hooks are implemented, here and in M2 and other plugins. There is too much unnecessary complexity added and there was a timeline field based solution already implemented in Drupal (still running) and M2 (discarded). I agree with Leon that the order state must not change. If the (Magento) authorization transaction for the order is in the open state we should probably just close it, add a history message to the order and return OK.

app/code/community/Bolt/Boltpay/Model/Payment.php
danac-gs: I agree. Fixing Bolt failing hooks noise problem should have no side effects on the Magento order.

johnmccombs1: So, I just realized that there are actually 3 issues being addressed in this PR and the primary one isn't actually addressed. Here are the issues:

1) If customer clicks Void button on Magento, we need to ensure that skip_hook_notification is set to true
2) If customer creates a partial capture and then voids the remaining transaction from Bolt, we currently cancel the order which isn't correct. This use case wasn't implemented two years ago because at the time we didn't have the ability to do a partial capture and it wasn't necessary. For example, If I have a $100 order and I do a $10 partial capture, and then a week later when the auth expires, Bolt sends a void request, it shouldn't cancel the order as I have processed part of the order. If you look here https://www.screencast.com/t/NEXcbljcZvaS you'll see that $amount is only set if it is a complete void (void amount == grand total) in which case we should cancel the order, but this doesn't handle a partial capture scenario. If it's a partial capture we want to put the order back into a processing state and let Magento handle it. For example, if the order should go to complete, then Magento will switch the status to complete on $order->save() which you can see here https://www.screencast.com/t/A9GTivB4l. So I still think this code is valid for partial capture.

3) Case described in the task which is that in some cases Bolt sends a partial void webhook and for some reason Magento is responding with an empty response body. I just realized that I didn't solve this problem, and I agree that if the Magento order is in a completed status and Bolt sends a void hook we should just respond with a 200 status. Thanks for calling this out because it made me realize that I hadn't actually fixed the original intent of the task.

I'm going to create 3 PRs for the cases above, but @mccottry and @danac-gs can you confirm that you agree with my assessment about not cancelling partially voided orders?

mccottry: @johnmccombs1   I like your approach.  Can modify it to be something like this:

```
    /**
     *  Handles two void cases sent by Bolt.
     *  1.) Complete void an order cancelling all funds
     *  2.) Partial void where partial capture has occurred but authorization has expired
     *
     * @param Mage_Payment_Model_Info $payment
     */
    public function handleVoidTransactionUpdate(Mage_Payment_Model_Info $payment){

        $authTransaction = $payment->getAuthorizationTransaction();
        $order = $payment->getOrder();
        $amount =  $order->getBaseGrandTotal() - $order->getBaseTotalPaid() ;

        if ($authTransaction->canVoidAuthorizationCompletely()) {
            // True void
            $reference = $payment->getAdditionalInformation('bolt_reference');
            $payment->setParentTransactionId($reference);
            $payment->setTransactionId(sprintf("%s-void", $reference));
            $payment->setShouldCloseParentTransaction(true);
            $transaction = $payment->addTransaction(Mage_Sales_Model_Order_Payment_Transaction::TYPE_VOID, null, true);
            $message = $this->getVoidMessage($payment, $transaction, $amount);
            $order->setState(Mage_Sales_Model_Order::STATE_CANCELED, true, $message);
        } else if (!$authTransaction->getIsClosed()) {
            // Open authorization has expired and partial capture has taken place.
            // We do not change the order state.  We only need to close the authorization.
            $authTransaction->closeAuthorization();
            $message = $this->getVoidMessage($payment, $authTransaction, $amount);
            $order->addStatusHistoryComment($message);
        }

        $order->save();
    }

    /**
     * @param Mage_Payment_Model_Info $payment
     * @param $transaction
     * @param null $amount
     * @return string
     */
    protected function getVoidMessage(Mage_Payment_Model_Info $payment, $transaction, $amount)
    {
        $order = $payment->getOrder();
        $message = Mage::helper('boltpay')->__('BOLT notification: Transaction authorization has been voided. Amount: %s.', $order->getBaseCurrency()->formatTxt($amount, array()));
        $message .= ' ' . Mage::helper('boltpay')->__('Transaction ID: "%s".', $transaction->getHtmlTxnId());
        return $message;
    }
```

This contains all the ideas that you outlined above.  The difference in the code other than reducing the lines is that it avoids changing the order state to anything but cancelled as a result of a void hook.  While it is true that for a partial capture, the order should be changed to Processing, this is not a partial capture, but a partial void, or more accurately stated, a closing of the authorization transaction.

johnmccombs1: > @johnmccombs1 the only part I am concerned of is this
> 
> ```
> // update order state and add comments
> $order->setState(isset($amount) ? Mage_Sales_Model_Order::STATE_CANCELED : Mage_Sales_Model_Order::STATE_PROCESSING, true, $message);
> ```
> 
> You set the order state to either CANCELED or PROCESSING.
> What happens if the order was COMPLETE?
> Will Magento always switch to the correct status on $order->save() ?
> Perhaps you should set it to CANCELED if it is a full void, as you do already, and leave the state unchanged otherwise.

@danac-gs I did this because we have diverged from what Magento default does with this. I believe we did so because at the time partial capture wasn't allowed so voiding payment also likely means that we should cancel the order. But with partial capture, it changes things. The reason for changing this to PROCESSING in cases of partial capture is because this is what default Magento does for all voids. Then on save it fixes the order to be in the correct state. This is how default Magento handles a void https://www.screencast.com/t/NEXcbljcZvaS. Notice that it always changes the status to PROCESSING. The reason for this is that on $order->save(), Magento then figures out what state to put it in and if it should be in a complete state it will change the state back to complete which you can see here https://www.screencast.com/t/A9GTivB4l.

Having said all of that though, on Monday I'm going to dig deeper into what @mccottry suggested and see if that is a better alternative.

app/code/community/Bolt/Boltpay/Model/Payment.php
```diff
@@ -445,14 +445,7 @@ public function handleTransactionUpdate(
                     $order->setState(Mage_Sales_Model_Order::STATE_PAYMENT_REVIEW, true, $message);
                     $order->save();
                 } elseif ($newTransactionStatus == self::TRANSACTION_CANCELLED) {
```
danac-gs: The code looks good.
However, there might be an issue here, not 100% sure but it needs to be checked.
Are you sure the status is CANCELLED when authorization times-out.
If you do a partial capture (transaction status is AUTHORIZED) and then void from the merchant dashboard
the new status is not CANCELLED but COMPLETED. I assume the same status is set on authorization time out.
So this may be another case:
1. previous transaction status: AUTHORIZED
2. new transaction status: COMPLETED
3. NO NEW CAPTURE

PR #329, opened 2019-03-07
Simplifies creation and use of helpers with better logical division

mccottry:
https://app.asana.com/0/544708310157130/1113209144650896

PR #331, opened 2019-03-12
Change method from helper to boltHelper to avoid Block object method name collision

mccottry:
https://app.asana.com/0/544708310157130/1113209144650896
https://app.asana.com/0/544708310157130/1113748340562763
https://app.asana.com/0/544708310157130/1113616110414457

Differs from PR #329 in that it: 
 
- renames helper method to boltHelper to avoid collision
- updates shipping unitTest with new helper name
- removes infinite loop to useless method applyBoltTraceId


daisy1754:
can you add description explaining difference between https://github.com/BoltApp/bolt-magento1/pull/329 and this PR?

mccottry:
> can you add description explaining difference between #329 and this PR?
@daisy1754 
For the moment, hold off on this PR.  I will have to do some manual Production Readiness as I was relying on being able to us Cirlce CI for that, but we are a few steps away from being able to use Circle CI in that way.

mccottry:
> can you add description explaining difference between #329 and this PR?

@daisy1754 
The difference is actually in the description and resolves a minor issue while at the same time adding better semantics.

However, the true major issue experienced in integration testing was something I identified in PR #329 and marked as  a TODO that @matt-thomason requested an investigation into:

https://github.com/BoltApp/bolt-magento1/blob/c0ea65089ab17eff3fd568d3c562c0eebabac643/app/code/community/Bolt/Boltpay/Helper/ApiTrait.php#L218-L238

https://github.com/BoltApp/bolt-magento1/pull/329#discussion-diff-263961883R221

As suspected, this code does in fact cause an infinite loop while it is effectively doing nothing.  Removing it fixes the issue checkout.

PR #336, opened 2019-03-26
Allow updating order state manually

johnmccombs1:
Asana task: https://app.asana.com/0/564264490825835/1115713860206740

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Payment.php
+++ b/app/code/community/Bolt/Boltpay/Model/Payment.php
@@ -671,6 +671,11 @@ class Bolt_Boltpay_Model_Payment extends Mage_Payment_Model_Method_Abstract
      */
     protected function createInvoice($order, $captureAmount) {
         if (isset($captureAmount)) {
+            $boltMaxCaptureAmountAfterRefunds = $this->getBoltMaxCaptureAmountAfterRefunds($order);
+            if($captureAmount > $boltMaxCaptureAmountAfterRefunds){
+                $captureAmount = $boltMaxCaptureAmountAfterRefunds;
+            }
+
             $this->validateCaptureAmount($order, $captureAmount);
 
             if($order->getGrandTotal() > $captureAmount) {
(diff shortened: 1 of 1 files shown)


app/code/community/Bolt/Boltpay/Model/Payment.php
```diff
@@ -191,6 +191,10 @@ public function fetchTransactionInfo(Mage_Payment_Model_Info $payment, $transact
                 Mage::getSingleton('adminhtml/session')->addNotice($message);
             }
 
+            if ($this->isAnAllowedUpdateFromAdminPanel($payment, $transactionStatus)) {
+                $this->updateOrderStateToProcessing($payment);
```
daisy1754: I think we need some comment here explaining why

mccottry: I can understand a desire to allow the Admin to override the status and state of any order, however, I don't agree that this is best done via the Magento "Get Payment Update" button that is suppose to be returning the most up-to-date status ads defined by the selected Payment Gateway, (Bolt is our case.)  

But, @johnmccombs1, I think that you are trying to say that the state and status of processing is a necessary and transient pre-condition to move an order from an (unholded/get payment update) state, to be moved to which ever state that Bolt sends back.  If this is the case, then I approve of these changes.  I just would like to know answer to these questions?

How does an order get into the state where the "Get Payment Update" is presented?   I had always believed that it is after the order was moved from being on-hold.  When this un-hold happens, the internal Magento behavior is to set the state and status to that which is was before the order was placed on hold.  Why is this previous state not sufficient for moving the most up to date Bolt status?

In essence, I am still trying to understand why you feel creating what seems as an extra transition to the Processing state is necessary?

Finally, do you believe switching the order to the state of Processing will trigger some events and have unexpected side effects for some merchants whose ERP maybe waiting for such a state?

johnmccombs1: @mccottry while walking through the reason for this code, it occurred to me that there might be a better solution altogether. The problem that I'm trying to solve here is this:

Currently when there is a mismatch the order transitions as follows:
1. Order goes into Payment Review state
2. Mismatch occurs
3. Order goes on hold
4. Merchant unhold's the order to fix the problem
5. Order goes back to payment review state (previous state)
6. Merchant clicks Get Update Payment button
7. Mismatch still occurs because even if the payment matches, the code looks at the original transaction amount (not the amount after fixing with a credit memo) and throws an exception so order state stays in Payment Review

In order to fix these I've been manually updating the order state/status to processing once it's been resolved. So the code here was doing the same thing:

1. If capture is completed already and the order is in payment review, go ahead and move it to processing.

However, after walking through this and realizing that what I'm doing to fix these isn't necessarily the best way to resolve the problem in code, I'm going to dig in a little deeper and see if I can just update the code to see if there really still is a mismatch and not throw an exception if there isn't. I'm hoping there is a bug in the code that's throwing the exception in the first place so I don't have to fix it here. Then the order will go into processing on its own without this fix.

PR #337, opened 2019-03-26
Ignore the partial refund fixing mismatch

johnmccombs1:
Asana task: https://app.asana.com/0/564264490825835/1115713860206738

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Payment.php
+++ b/app/code/community/Bolt/Boltpay/Model/Payment.php
@@ -534,7 +534,7 @@ class Bolt_Boltpay_Model_Payment extends Mage_Payment_Model_Method_Abstract
                         $creditmemo->setPaymentRefundDisallowed(false);
                         $creditmemo->register()->save();
                     }
-                } else { // partial refund
+                } else if (!$this->isPartialRefundFixingMismatch($order, $transactionAmount)) { // partial refund
                     $isPartialRefund   = true;
                     //actually for order with bolt payment, there is only one invoice can refund
                     foreach ($invoiceIds as $k => $invoiceId) {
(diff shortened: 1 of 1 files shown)


johnmccombs1:
> @johnmccombs1, @daisy1754
> 
> I will have to admit that I do not understand how we are doing a partial refund when it is only allowed when the refund amount is equal to the amount of the total paid on the order. To me, it seems like this is the condition for a full refund. This could be that I just don't understand the problem and the change. Certainly, total paid could represent a partial payment/capture on the order, but, I still think this is a minority case for partial refunds. As I see this, if a partial refund in made on Bolt, the Magento order would not be updated unless it was the full refund????? This seems to place the Magento order data to be out of sync with the Bolt order data.
> 
> Outside of my lack of understanding of this, unit test and integration test are failing.
> 
> At minimum, we may wish to add some integration and unit test to confirm that these are producing production ready results.
> 
> @johnmccombs1
> Please take time to consider whether we are achieving what we wish to achieve here. I am too uncertain to approve or request change to this PR, but normally would request a change. For now, can you just explain how this change makes sense in the context of a partial refund?

@mccottry let me explain the use case and why it matters here. Currently, there is an issue where sometimes a discount is applied successfully on Magento but not on Bolt which causes a mismatch (there are other cases where this could occur, but this is a specific one that this code was developed for). For example, let's say that a discount applies to Magento and the price is $100, but the discount doesn't get applied on the Bolt side so the price is $120. If the merchant needs to refund $20, currently it will refund it both on Magento as well as on Bolt. So the new totals will be Bolt: $100 and Magento: $80. The purpose of this code is to check to see if the Bolt total matches Magento and if not, it doesn't apply the refund. So in the example above let's say that they again refund $20. Then Bolt drops to $100 and Magento skips the refund. If they refund again for $20, the refund will be applied to Magento and both totals will drop to $80. Let me know if this doesn't make sense or if you find any other issues.

matt-thomason:
@johnmccombs1 Can you write a bit more about how this check solves the greater issue outlined in the Asana task?

Also, should there be an else case reporting that we went through this control statement didn't do anything?

johnmccombs1:
> @johnmccombs1 Can you write a bit more about how this check solves the greater issue outlined in the Asana task?
> 
> Also, should there be an else case reporting that we went through this control statement didn't do anything?

Sure @matt-thomason Here is an example to explain the issue that the code here solves https://github.com/BoltApp/bolt-magento1/pull/337. Let's say that there is a product that costs $120. Customer adds it to their cart. Before the customer completes checkout though, the Magento admin changes the product price to $100. When they place an order there will be a mismatch. Our typical practice is to tell merchants to create a refund on whichever system the price is wrong. In this case it's wrong on the Bolt side so we would ask them to refund on Bolt. Here is how the current process would work.

**Starting totals**
Bolt total:$120 
Magento total $100

**Current Process**
1. Issue the refund of $20 on Bolt => Total refunded amount = $20
2. Refund hook is sent to Magento to create credit memo (current hook amount = $20)
3. Magento issues a refund for $20
4. The Magento total is now $80 and Bolt is now $100. Mismatch still occurs
5. Merchant issues an additional refund of $10 due to customer complaining about being billed too much.
6. Refund hook is sent to Magento to create credit memo for $10
7. Magento issues a refund for $10
8. The Magento total is now $70 and Bolt is now $90. Mismatch still occurs and customer doesn't see the correct amount via email or in their My Account section.

**New Process**
1. Issue the refund of $20 on Bolt => Total refunded amount = $20
2. Refund hook is sent to Magento to create credit memo (current hook amount = $20)
3. Check to see if $120 - $20 (total Bolt refunds)  + $20 (hook refund amount) != $100 - 0 (Magento refunds)
4. Ignore the hook since the totals don't match and it's assumed that this means that the refund is to fix the Bolt total
6. Issue the refund of $10 on Bolt => Total refund amount = $30
7. Refund hook is sent to Magento to create credit memo of $10
8. Logic check: $120 - $30 (total Bolt refunds) + $10 (hook refund amount) = $100 - 0 (Magento refunds)
9. Create partial credit memo in Magento for $10
10. The Magento total is now $90 and Bolt is now $90. Mismatch doesn't occur and customer sees the correct amount via email and in their My Account section.

This also solves the problem of a full refund. Right now a full refund will fail in the case above since if they try to refund $20 to make them match and then refund $100, Magento will actually only have $80 to refund. With the new code it will succeed since the initial $20 refund will only occur on the Bolt side and the remaining totals will match. Then they can refund the remaining amount of $100 without issues. Let me know if this doesn't make sense.

johnmccombs1:
@matt-thomason I forgot to answer your second question. If you look here https://www.screencast.com/t/hGH9m4iqe, you can see that the refund doesn't occur in the order comments because it was skipped, but the second one does. Is that clear enough, or should I add some additional logging?

matt-thomason:
> @matt-thomason I forgot to answer your second question. If you look here https://www.screencast.com/t/hGH9m4iqe, you can see that the refund doesn't occur in the order comments because it was skipped, but the second one does. Is that clear enough, or should I add some additional logging?

Nope, that's clear enough I think.

PR #339, opened 2019-03-29
Fix codesniffer warning around strpos

daisy1754:


mccottry:
oh. codesniffer is a little to smart for its own good. 👓

PR #341, opened 2019-03-29
Allowed custom route for Bolt button.

YevhenShBolt:
Allowed custom route for Bolt button.
This change allows showing Bolt button if the checkout/cart (or other default pages which we use with Block/Checkout/Bolt.php) page was rewritten.

### DIFF
--- a/app/code/community/Bolt/Boltpay/Helper/ConfigTrait.php
+++ b/app/code/community/Bolt/Boltpay/Helper/ConfigTrait.php
@@ -127,7 +127,8 @@ trait Bolt_Boltpay_Helper_ConfigTrait {
     {
         $checkoutType = Bolt_Boltpay_Block_Checkout_Boltpay::CHECKOUT_TYPE_MULTI_PAGE;
         $config = $this->getPaymentBoltpayConfig('allowed_button_by_custom_routes', $checkoutType);
-        $result = explode(',', $config);
+        // removes all NULL, FALSE and Empty Strings but leaves 0 (zero) values
+        $result = array_filter(explode(',', $config));
 
         return (empty($result)) ? [] : $result;
     }
--- a/app/code/community/Bolt/Boltpay/Helper/ConfigTrait.php
+++ b/app/code/community/Bolt/Boltpay/Helper/ConfigTrait.php
@@ -128,7 +128,8 @@ trait Bolt_Boltpay_Helper_ConfigTrait {
         $checkoutType = Bolt_Boltpay_Block_Checkout_Boltpay::CHECKOUT_TYPE_MULTI_PAGE;
         $config = $this->getPaymentBoltpayConfig('allowed_button_by_custom_routes', $checkoutType);
         // removes all NULL, FALSE and Empty Strings but leaves 0 (zero) values
-        $result = array_filter(explode(',', $config));
+        $configData = explode(',', $config);
+        $result = array_values(array_filter(array_map('trim', $configData), 'strlen'));
 
         return (empty($result)) ? [] : $result;
     }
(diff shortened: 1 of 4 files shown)


app/code/community/Bolt/Boltpay/Helper/ConfigTrait.php
```diff
@@ -120,6 +120,18 @@ public function getPaymentBoltpayConfig($configPath, $checkoutType)
...
+    /**
+     * @return array
+     */
+    public function getAllowedButtonByCustomRoutes()
+    {
+        $checkoutType = Bolt_Boltpay_Block_Checkout_Boltpay::CHECKOUT_TYPE_MULTI_PAGE;
+        $config = $this->getPaymentBoltpayConfig('allowed_button_by_custom_routes', $checkoutType);
+        $result = explode(',', $config);
+
+        return (empty($result)) ? [] : $result;
```
danac-gs: $result of the explode is never empty
In case of the empty $config string it is
array (
  0 => '',
)
which does not brake things in this particular case but is incorrect.
you can first filter out empty values and then check if it is empty
array_filter(explode(',', $config));

YevhenShBolt: thanks, updated.

app/code/community/Bolt/Boltpay/Helper/ConfigTrait.php
```diff
@@ -120,6 +120,19 @@ public function getPaymentBoltpayConfig($configPath, $checkoutType)
         return ($checkoutType === Bolt_Boltpay_Block_Checkout_Boltpay::CHECKOUT_TYPE_ADMIN) && !Mage::getStoreConfig('payment/boltpay/use_javascript_in_admin') ? '' : $configValue;
     }
 
+    /**
+     * @return array
+     */
+    public function getAllowedButtonByCustomRoutes()
```
daisy1754: unit test please - test for empty, data w/o comma, and data /w command would be good.

app/code/community/Bolt/Boltpay/etc/system.xml
```diff
@@ -177,6 +177,16 @@
               <show_in_store>1</show_in_store>
             </selector_styles>
 
+            <allowed_button_by_custom_routes>
+                <label>Allowed Button By Custom Route(s)</label>
```
mccottry: @YevhenShBolt, Relabeling this to say: "Show Bolt Button on Pages By Custom Route(s)" is more clear to me what this setting does.

Can we update it?

YevhenShBolt: @mccottry i will update. Thanks.

app/code/community/Bolt/Boltpay/Helper/ConfigTrait.php
YevhenShBolt: added

app/code/community/Bolt/Boltpay/etc/system.xml
YevhenShBolt: @mccottry updated

daisy1754: I don't think it's updated?

app/code/community/Bolt/Boltpay/Helper/ConfigTrait.php
daisy1754: I don't see any test. Did you forget to include it in PR?

YevhenShBolt:
@daisy1754 thanks. I missed it.
I updated label and added tests.

PR #348, opened 2019-04-11
Keep the cart after placing an order from the product page

johnmccombs1:
Asana task: https://app.asana.com/0/564264490825835/1117712752858652

This PR adds custom order confirmation page

### DIFF
--- a/app/code/community/Bolt/Boltpay/etc/config.xml
+++ b/app/code/community/Bolt/Boltpay/etc/config.xml
@@ -89,7 +89,7 @@
         <observers>
           <bolt_boltpay_clear_shopping_cart>
             <class>Bolt_Boltpay_Model_Observer</class>
-            <method>clearShoppingCart</method>
+            <method>clearShoppingCartExceptPPCOrder</method>
             <depends>
               <ZZZZZ_Bolt />  <!--  psuedo module to place execution at the back of event queue -->
             </depends>
--- a/app/code/community/Bolt/Boltpay/Helper/Data.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Data.php
@@ -208,6 +208,9 @@ class Bolt_Boltpay_Helper_Data extends Mage_Core_Helper_Abstract
                     initBoltButtons();
                     ";
                 break;
+            case Bolt_Boltpay_Block_Checkout_Boltpay::CHECKOUT_TYPE_PRODUCT_PAGE:
+                $quoteId = Mage::getSingleton('checkout/session')->getQuoteId();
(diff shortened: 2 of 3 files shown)


app/code/community/Bolt/Boltpay/etc/config.xml
```diff
@@ -176,13 +176,6 @@
 
   <admin>
     <routers>
```
daisy1754: can you send me separate PR for this? assuming it's not related to product page checkout

app/code/community/Bolt/Boltpay/Helper/GeneralTrait.php
```diff
@@ -150,6 +150,16 @@ public function isShoppingCartPage()
...
     }
 
+    /**
+     * @return bool
+     */
+    public function isProductPageCheckout()
+    {
+        $checkoutSession = Mage::getSingleton('checkout/session');
+
+        return $checkoutSession->getData('checkout_type') == Bolt_Boltpay_Block_Checkout_Boltpay::CHECKOUT_TYPE_PRODUCT_PAGE ? true : false;
```
matt-thomason: This could be simplified as `return $checkoutSession->getData('checkout_type') == Bolt_Boltpay_Block_Checkout_Boltpay::CHECKOUT_TYPE_PRODUCT_PAGE;`

app/code/community/Bolt/Boltpay/etc/config.xml
johnmccombs1: > LGTM, can you attach screenshot of success screen?

See the screenshot here https://screencast.com/t/yQvvWj7es

app/code/community/Bolt/Boltpay/Helper/GeneralTrait.php
johnmccombs1: Thank @matt-thomason, I've updated it.

app/code/community/Bolt/Boltpay/controllers/OnepageController.php
```diff
@@ -92,4 +92,29 @@ public function saveShippingMethodAction()
             throw $e;
         }
     }
+
+    /**
+     * Order success action
+     */
+    public function successAction()
```
mccottry: Unfortunately, we are overriding the same method here and there is no clear way to distinguish if this is a product page order or a normal order.  

https://github.com/BoltApp/bolt-magento1/blob/cadf0069d45a6570253310aa39a9c40a3f717cf9/app/code/community/Bolt/Boltpay/controllers/OnepageController.php#L43-L80

Since you wish to pursue this without adding something to the DB then my suggestion is to pass a URL GET paramater for `is_bolt_pdp`.  However, by far, I think the far better idea is to extend the order model table with this `is_bolt_pdp`.  With this flag, you will not need to override the success URL at all and we will always be able to clearly tell which orders came from product page and which from some other source.  Is there a reason why you are not wanting to extend the order model?

PR #353, opened 2019-04-29
Ignore the credit hook request if credit memos exist

johnmccombs1:
Here the case that this PR fixes: Bolt order $100, Magento order $100
1. Merchant creates an offline Magento credit memo for $10
2. No Bolt refund is created
3. Merchant goes to their Bolt account and issues a refund of $10 to match the Magento refunded amount.
4. Credit hook request is sent to Magento
5. A new credit memo of $10 is created on Magento causing the Magento total to be $80 while the Bolt total is $90

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Payment.php
+++ b/app/code/community/Bolt/Boltpay/Model/Payment.php
@@ -462,6 +462,10 @@ class Bolt_Boltpay_Model_Payment extends Mage_Payment_Model_Method_Abstract
                     $order->setState(self::ORDER_DEFERRED, true, $message);
                     $order->save();
                 } elseif ($newTransactionStatus == self::TRANSACTION_REFUND) {
+                    // Ignore credit hook request if credit memos exist
+                    if ($this->ignoreCreditHookRequest($payment, $transactionAmount)) {
+                        return;
+                    }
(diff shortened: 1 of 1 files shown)


johnmccombs1:
> The logic is not good here. If the merchant chooses to do an OFFLINE credit memo, we must assume that he has done that for a reason. Maybe 10 dollars was refunded from out of pocket or the cash register. That is the whole purpose of an OFFLINE refund.
> 
> The correct logic is as you described where the totals do not match. If merchant wishes to have the two totals match, then they must do an ONLINE refund from Magento or must initiate the refund from Bolt.
> 
> If the merchant does not have the option to issue an ONLINE refund from Magento, and wishes to have this, then the correct fix is to figure why ONLINE refunds are not enabled in Magento and fix the problem there.
> 
> Otherwise, you can offer this as a customization for any merchant that is requesting this non-standard behavior.

Ugh...I hate when I get caught up in a problem without fully thinking through it. This is a good callout. We've had a few merchants with this issue recently and it requires a manual change in the database to fix it, but maybe it's more a question of training, or maybe simply that we need to add .a simple "Delete Offline Credit Memo" button so this can be easily fixed on the Magento side.

mccottry:
> maybe it's more a question of training, or maybe simply that we need to add .a simple "Delete Offline Credit Memo" button so this can be easily fixed on the Magento side.

Both sound like the right path to me

PR #354, opened 2019-04-29
Implement temporarily rejected payment API in admin

johnmccombs1:
Asana task: https://app.asana.com/0/564264490825835/803147412595231

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Payment.php
+++ b/app/code/community/Bolt/Boltpay/Model/Payment.php
@@ -25,6 +25,8 @@ class Bolt_Boltpay_Model_Payment extends Mage_Payment_Model_Method_Abstract
     const METHOD_CODE               = 'boltpay';
     const TITLE                     = "Credit & Debit Card";
     const TITLE_ADMIN               = "Credit and Debit Card (Powered by Bolt)";
+    const DECISION_APPROVE          = 'approve';
+    const DECISION_REJECT           = 'reject';
 
     // Order States
     const ORDER_DEFERRED = 'deferred';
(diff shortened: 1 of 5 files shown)


app/code/community/Bolt/Boltpay/Model/Payment.php
```diff
@@ -1071,4 +1068,95 @@ protected function isPartialRefundFixingMismatch(Mage_Sales_Model_Order $order,
...
+     *
+     * @return bool
+     */
+    public function canReviewPayment(Mage_Payment_Model_Info $payment)
+    {
+       return $payment->getAdditionalInformation('bolt_transaction_status') == self::TRANSACTION_REJECTED_REVERSIBLE;
+    }
+
+    /**
+     * Attempt to accept a payment that us under review
```
daisy1754: that *is*

daisy1754: maybe "Attempt to approve a order"? payment is already taken, decision is whether we approve it or not

app/code/community/Bolt/Boltpay/Model/Payment.php
```diff
@@ -1071,4 +1068,95 @@ protected function isPartialRefundFixingMismatch(Mage_Sales_Model_Order $order,
...
+     *
+     * @param Mage_Payment_Model_Info $payment
+     * @return bool
+     * @throws Exception
+     */
+    public function acceptPayment(Mage_Payment_Model_Info $payment)
```
daisy1754: let's call it approve instead of accept

mccottry: @daisy1754 @matt-thomason This is a superclass function that he is implementing here.  He could use a wrapper method to change the name, but for built in Magento, this is how they defined it. :(

app/code/community/Bolt/Boltpay/Model/Payment.php
```diff
@@ -778,10 +777,8 @@ public static function transactionStatusToOrderStatus( $transactionStatus ) {
                 $new_order_status = Mage_Sales_Model_Order::STATE_PROCESSING;
                 break;
             case Bolt_Boltpay_Model_Payment::TRANSACTION_PENDING:
-                $new_order_status = Mage_Sales_Model_Order::STATE_PAYMENT_REVIEW;
-                break;
             case Bolt_Boltpay_Model_Payment::TRANSACTION_REJECTED_REVERSIBLE:
-                $new_order_status = Bolt_Boltpay_Model_Payment::ORDER_DEFERRED;
```
mccottry: @johnmccombs1 While it maybe a good idea to change the order state to `Mage_Sales_Model_Order::STATE_PAYMENT_REVIEW`, it is also a good idea to keep the status as accurate as possible, so this shouldn't be changed.  It should reflect Bolt terminology.  You should change this status;s label in the database and associate its state with `Mage_Sales_Model_Order::STATE_PAYMENT_REVIEW`.

The UI, as @daisy1754 pointed out should clearly express that Bolt rejected the payment and preferably should warn that approving is not indemnified

PR #355, opened 2019-05-02
Don't bugsnag from CI

daisy1754:


mccottry:
@johnmccombs1 
Good catch.  I saw this two weeks ago and it was my mistake not to make this bug fix against develop and then merge or rebase 2.0.0 to the develop branch.

While the solutions are similar, can we please use this approach instead:
https://github.com/BoltApp/bolt-magento1/commit/9c466a463c5dd986719203ec900839d4753be236#diff-16f0ef380b9b673f4772297351eaae1eR1020

The difference is that in the case the order does not have a transaction, it can still be voided.  Definitely, if it is in the Bolt dashboard, it should by all rights have a Magento transaction, but, in the case that it doesn't and the order exist in both locations, Magento and Bolt, the order is still canceled.  Let me know what you think.

mccottry:
@daisy1754 
I'm curious as to how and why this works.  I imagine that the CI is in a different release stage than `development` and `production`, but it seems like those are the only options provided.  Is the CI bugsnag release stage configured differently somehow?

daisy1754:
@mccottry  we use stage 'test' for PHPUnit
https://github.com/BoltApp/bolt-magento1/blob/a0c9a20a48c102ed8af7fd5ceac329e7d57bf4dc/app/code/community/Bolt/Boltpay/Helper/BugsnagTrait.php#L55

PR #357, opened 2019-05-03
Ingore void hook for unauthorized order

johnmccombs1:
Asana task: https://app.asana.com/0/564264490825835/1120800506043077

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Payment.php
+++ b/app/code/community/Bolt/Boltpay/Model/Payment.php
@@ -1000,7 +1000,7 @@ class Bolt_Boltpay_Model_Payment extends Mage_Payment_Model_Method_Abstract
         $order = $payment->getOrder();
         $amount =  $order->getBaseGrandTotal() - $order->getBaseTotalPaid() ;
 
-        if ($authTransaction->canVoidAuthorizationCompletely()) {
+        if (!$authTransaction || $authTransaction->canVoidAuthorizationCompletely()) {
             // True void
             $reference = $payment->getAdditionalInformation('bolt_reference');
             $payment->setParentTransactionId($reference);


mccottry:
@johnmccombs1 
Good catch.  I saw this two weeks ago and it was my mistake not to make this bug fix against develop and then merge or rebase 2.0.0 to the develop branch.

While the solutions are similar, can we please use this approach instead:
https://github.com/BoltApp/bolt-magento1/commit/9c466a463c5dd986719203ec900839d4753be236#diff-16f0ef380b9b673f4772297351eaae1eR1020

The difference is that in the case the order does not have a transaction, it can still be voided.  Definitely, if it is in the Bolt dashboard, it should by all rights have a Magento transaction, but, in the case that it doesn't and the order exist in both locations, Magento and Bolt, the order is still canceled.  Let me know what you think.

johnmccombs1:
> @johnmccombs1
> Good catch. I saw this two weeks ago and it was my mistake not to make this bug fix against develop and then merge or rebase 2.0.0 to the develop branch.
> 
> While the solutions are similar, can we please use this approach instead:
> [9c466a4#diff-16f0ef380b9b673f4772297351eaae1eR1020](https://github.com/BoltApp/bolt-magento1/commit/9c466a463c5dd986719203ec900839d4753be236#diff-16f0ef380b9b673f4772297351eaae1eR1020)
> 
> The difference is that in the case the order does not have a transaction, it can still be voided. Definitely, if it is in the Bolt dashboard, it should by all rights have a Magento transaction, but, in the case that it doesn't and the order exist in both locations, Magento and Bolt, the order is still canceled. Let me know what you think.

Thank @mccottry for the solution. I've updated the PR and confirmed it works.

PR #359, opened 2019-05-04
mPR01 - Distinct Bolt Exceptions with new OrderCreationException class implemented

mccottry:
https://app.asana.com/0/544708310157130/1121158206133098

In version 2.0, we improve our error design by designating all Bolt generated exceptions as `Bolt_Boltpay_BoltException`'s (_BoltException_) which allows us to distinguish between Bolt expected exceptions and all other types.

A _BoltException_ will have access to all helper methods as provided the global Bolt trait that is now ubiquitous among all Bolt classes.  The global helper greatly improves IDE code hinting for development.

Finally, we introduce the `Bolt_Boltpay_OrderCreationException` (_OrderCreationException_) which closely adheres to the specification defined here:
https://docs.google.com/document/d/1L5DzKak94_QVS-4VR0u4atXk4M4SvXe_krcZDZiC9rg/edit

Deviations are noted in the comments, and the exceptions have been extended for fuller coverage.  Additionally, the new shipping change code has been implemented. 

 Error messages are i18n ready.


mccottry:
> nice. this is exactly why I wanted separate PRs. Can you please point this PR to `develop` branch?

@daisy1754 
I'm glad I can make this informative for you and for others. This is one means to do it and it takes time.

It is not appropriate to point many of the subsquent changes back to develop because it will break 1.x code.

Please read the description in the parent task.  I am going to add the links.  but it reads:

**_The process will be merging each micro code back to an intermediate branch which is a copy of the current develop branch.  Once all interdependent micro pieces are merged to intermediate branch, the collective pull request will be complete and the intermediate branch will be merged into develop, signifying the release of v2.0.0

Each pull request are pointed to the furthest unapproved micro PR down the chain.  Once a lower numbered micro PR is approved and merged, then the base of this will be lowered until it reaches the intermediate branch._**

app/code/community/Bolt/Boltpay/OrderCreationException.php
```diff
@@ -0,0 +1,214 @@
...
+
+    /**
+     * Sets the httpCode based on the error
+     *
+     * @param int    $code         Bolt error code
+     * @param string $dataTemplate specific Bolt error sub-category
+     *
+     * @return int The HTTP code that was set
+     */
+    private function setHttpCode( $code, $dataTemplate ) {
```
daisy1754: I think readability would increase if you split logic to get code. eg., you can define `getCode` method that just returns code

```
private function getHttpCode( $code, $dataTemplate ) {
switch ($code) {
            case self::E_BOLT_GENERAL_ERROR:
                if ($dataTemplate === self::E_BOLT_GENERAL_ERROR_TMPL_HMAC) {
                    return 401;
                }
            case self::E_BOLT_ORDER_ALREADY_EXISTS:
                return 409;
            }
}
```

and then instead of `$this->setHttpCode` you can do `$this->httpCode = $this->getHttpCode`

app/code/community/Bolt/Boltpay/OrderCreationException.php
```diff
@@ -0,0 +1,214 @@
...
+            case self::E_BOLT_GENERAL_ERROR:
+                if ($dataTemplate === self::E_BOLT_GENERAL_ERROR_TMPL_HMAC) {
+                    $this->httpCode = 401;
+                    break;
+                }
+            case self::E_BOLT_ORDER_ALREADY_EXISTS:
+                $this->httpCode = 409;
+                break;
+            case self::E_BOLT_CART_HAS_EXPIRED_TMPL_NOT_FOUND:
+                $this->httpCode = 404;
```
daisy1754: does magento or php have constant for these HTTP status code? If possible, I prefer we say `return StatusCodes.NotFound` instead of `return 404`

app/code/community/Bolt/Boltpay/OrderCreationException.php
```diff
@@ -0,0 +1,214 @@
...
+     * @param int    $code          The Bolt defined error code
+     * @param string $dataTemplate  specific Bolt error sub-category
+     * @param array  $dataValues    An array of values to be added to the data template
+     * @param string $message       The exception message to throw.
+     * @param Throwable|null $previous  [optional] The previously throwable used for exception chaining.
+     */
+    public function __construct($code = self::E_BOLT_GENERAL_ERROR, $dataTemplate = self::E_BOLT_GENERAL_ERROR_TMPL_GENERIC, array $dataValues = array(), $message = null, Throwable $previous = null)
+    {
+        // If code is invalid, we will use the generic message
+        if (!in_array($code, self::$validCodes)) {
```
daisy1754: Hm, this seems error prone. i can imagine someone adding constant to this class without realizing you have to update `validCodes` array also. Do we need this logic?

app/code/community/Bolt/Boltpay/OrderCreationException.php
```diff
@@ -0,0 +1,214 @@
...
+ *
+ * This exception is thrown when an error occurs while trying to create a backend order
+ */
+class Bolt_Boltpay_OrderCreationException extends Bolt_Boltpay_BoltException
+{
+
+    /**
+     * Database error or error in incoming request
+     */
+    const E_BOLT_GENERAL_ERROR               = 2001001; // verified
```
daisy1754: what does `verified` mean?

app/code/community/Bolt/Boltpay/OrderCreationException.php
```diff
@@ -0,0 +1,214 @@
...
+     * Currently not used -- shipping tax in M1 is used to fix rounding problems and reports the
+     * full cart tax.  Therefore, checking full tax is sufficient
+     */
+    const E_BOLT_CART_HAS_EXPIRED_TMPL_SHIPPING_TAX       = '{"reason": "Shipping tax amount has changed", "old_value": "%d", "new_value": "%d"}';
+
+    /**
+     * Item price changes
+     */
+    const E_BOLT_ITEM_PRICE_HAS_BEEN_UPDATED         = 2001004;
+    const E_BOLT_ITEM_PRICE_HAS_BEEN_UPDATED_TMPL    = '{"product_id": "%d", "old_price": "%d", "new_price": "%d"}';
```
daisy1754: writing JSON with template variable seems strange to me. One downside is if you write an invalid JSON here you won't notice until runtime. How about we use plain PHP object or array and convert it to json string when this class generating error message?

app/code/community/Bolt/Boltpay/OrderCreationException.php
daisy1754: or instead of `getHttpCode` maybe you can have map from error code to http status code

app/code/community/Bolt/Boltpay/OrderCreationException.php
```diff
@@ -0,0 +1,214 @@
...
+        self::E_BOLT_CART_HAS_EXPIRED,
+        self::E_BOLT_ITEM_PRICE_HAS_BEEN_UPDATED,
+        self::E_BOLT_OUT_OF_INVENTORY,
+        self::E_BOLT_DISCOUNT_CANNOT_APPLY,
+        self::E_BOLT_DISCOUNT_DOES_NOT_EXIST,
+        self::E_BOLT_SHIPPING_PRICE_HAS_BEEN_UPDATED
+    );
+
+    /**
+     * Bolt_Boltpay_InvalidTransitionException constructor.
```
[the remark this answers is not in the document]
mccottry: > can you add unit test for `Bolt_Boltpay_OrderCreationException`. Only checking one specific error would be sufficient.

@daisy1754  
Yes.  I certainly will.

Proper unit test to check error the triggering of error codes will take an large amount of time.  It is not trivial and I do not want to approach it like is a trivial thing.  Checking one error code will not be sufficient and is at the heart of v2.0.0.  Even more, I prefer integration test and verifying the Bolt modal does what it is suppose to do.

Solid additional unit test will be done in a separate PR.
https://app.asana.com/0/941895179897714/1121207201774756

app/code/community/Bolt/Boltpay/OrderCreationException.php
mccottry: it was a marker that I used when I manually verified all error codes by manipulating the magento admin.  I forgot to remove these markers

app/code/community/Bolt/Boltpay/OrderCreationException.php
mccottry: They are not a template _variables_.  They are appropriately unwavering template **constants**.  Plain old PHP objects can not be specified as constants.  Like with most programming languages, we are required to stick with scalar values for our use case.

The list JSON strings are all verified valid.  The function "__" which injects the variable portion of the templates uses `vsprintf` which assures the variables will be in the expected data type.  The design here is simple and solid.

The statement above is like saying, "if someone creates a PHP object that is encoded to JSON in an invalid format, we will not know until run time".  The solution is that should verify that the PHP object encodes to the expected format prior to PR.  Similarly, if someone adds to the constants or modifies them in the future, they should assure that they have written valid JSON before requesting a pull.

app/code/community/Bolt/Boltpay/OrderCreationException.php
mccottry: It is not error prone.  If the person forgets to add a valid code, they will get a generic error message with the reason for the error clearly stated in a generic way.  This will be very clear on any basic inspection of Datadog or Bugsnag

However, the person should not forget to register their new error code.  It is clearly commented and should be very obvious to do so to junior and mid-level developers.

app/code/community/Bolt/Boltpay/OrderCreationException.php
mccottry: There are many good approaches to doing many things.  We must makes sure that are assuring that a good, reasonable approach is used, which is not always our own. 

This is an stylistic request that I agree with.  Having the class data members visibly set in the constructor makes for a faster easier read.

A map is not a good approach.  If the business logic expands, the map will prove to be limiting.  Currently, you can see this for case `self::E_BOLT_GENERAL_ERROR` and `default`.  If the code maps to `self::E_BOLT_GENERAL_ERROR`, then only in the case that the template is for `self::E_BOLT_GENERAL_ERROR_TMPL_HMAC` should 401 be returned, otherwise the default 422.  There is construct for default values in a associative array in PHP, so the switch/case construct is optimal for this use case.

Task created here:
https://app.asana.com/0/1118494470563243/1121207201774762

app/code/community/Bolt/Boltpay/OrderCreationException.php
mccottry: Maybe.  But, if not, we will implement it.  I agree.  Here it will help the engineer understand the HTTP spec'd significance of each error code.  This is a non-imperative beautification request.

Task tracked here:
https://app.asana.com/0/1118494470563243/1121207201774763

app/code/community/Bolt/Boltpay/OrderCreationException.php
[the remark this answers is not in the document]
daisy1754: Please, please don't treat unit testing as a separate step. It must be part of your change. If you add new meaningful logic, your PR must include unit test. I've kept saying this for long time. 

For other comments I'll give other team members a chance to jump in in case people have any opinion, but for this specific one my answer is clear - unit test has to be part of this PR.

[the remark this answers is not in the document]
mccottry: Oops, Kazuki, perhaps I was unclear.  I am certainly adding a unit test for one specific error for this PR that you feel is sufficient.  That is what I meant by, _"Yes, I certainly will."_

I am also saying that _"Solid additional unit test will be done in a separate PR."_ because this is at the heart of this project and its importance can not be overstated.  UI integration test also seem very important to me.

Your answer is clear.  Apologies for me being unclear.  Thank you for the solid and stern feedback.  It is certainly reflecting in an overall high quality of this release.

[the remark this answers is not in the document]
mccottry: @daisy1754 
I now understand where the confusion lies on my end.  You are looking at this from a micro perspective, (i.e. do the methods `getJson` and `getHttpCode` produce valid and expected results when  `Bolt_Boltpay_OrderCreationException` is constructed.)

My mind, at this point after so many beta releases and seeing the code perform at that level, is accustomed to be looking and thinking at the same thing from a macro level, (i.e. when manipulating the Magento system to produce the various error states, and being called from the validation methods, are all the correct JSON errors triggered and displayed by the the `Bolt_Boltpay_OrderCreationException` class.) 

With my understanding fixed, I'm sure now I can quickly produce what you are asking for.

Thanks again.  Your reviews have often been the ones to push me to deliver at an even higher-level and I am very grateful that you take the time for this.  Now, let's rise these PR's up to the next level and get this code into official production release!

mccottry:
@daisy1754 It appears in adding unit test, I merged the other changes in the PR above in here.  It is very tricky trying to keep track of this and keep these things unnaturally apart, I will see what I can do about this.

mccottry:
The branch became mix with other files.  The new mPR is here:
https://github.com/BoltApp/bolt-magento1/pull/364

app/code/community/Bolt/Boltpay/OrderCreationException.php
daisy1754: what do you think about removing this logic and `$validCodes`?

PR #360, opened 2019-05-04
mPR02 - Bolt order creation bugfixes, improved internal docs, naming, and formatting

mccottry:
Bolt order creation is improved by
- bugfix: Uses proper id of cart item which is critical for cart item change inspection
- enhancement: Adding coupon reference to the Bolt order
- adding session/quote context information when adding discounts to the order

Style improved by
- using conventional boolean naming conventions 
- adding better spacing and indentation

Test improved by
- being consistent in using Quote's `getAllVisibleItems` method to generate bolt cart

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/BoltOrder.php
+++ b/app/code/community/Bolt/Boltpay/Model/BoltOrder.php
@@ -76,11 +76,13 @@ class Bolt_Boltpay_Model_BoltOrder extends Bolt_Boltpay_Model_Abstract
      * Generates cart submission data for sending to Bolt order cart field.
      *
      * @param Mage_Sales_Model_Quote        $quote      Magento quote instance
-     * @param bool                          $multipage  Is checkout type Multi-Page Checkout, the default is true, set to false for One Page Checkout
+     * @param bool                          $isMultipage  Is checkout type Multi-Page Checkout, the default is true, set to false for One Page Checkout
      *
      * @return array            The cart data part of the order payload to be sent as to bolt in API call as a PHP array
+     *
+     * @throws Mage_Core_Model_Store_Exception if the store cannot be determined
      */
-    public function buildCart($quote, $multipage)
+    public function buildCart($quote, $isMultipage)
     {
 
         ///////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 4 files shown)


app/code/community/Bolt/Boltpay/Model/BoltOrder.php
```diff
@@ -325,12 +327,20 @@ protected function addDiscounts( $totals, &$cartSubmissionData, $quote = null )
...
+                $data = array(
                     'amount'      => $discountAmount,
                     'description' => $totals[$discount]->getTitle(),
                     'type'        => 'fixed_amount',
                 );
+                if ($discount === 'discount' && $quote->getCouponCode()) {
+                    $coupon = Mage::getModel('salesrule/coupon')->load($quote->getCouponCode(), 'code');
+                    $rule = Mage::getModel('salesrule/rule')->load($coupon->getRuleId());
+
+                    if (strpos($totals[$discount]->getTitle(), $rule->getName()) !== false) {
```
daisy1754: can you add comment explaining why and what you do here?

mccottry: no problem.

app/code/community/Bolt/Boltpay/Model/BoltOrder.php
```diff
@@ -325,12 +327,29 @@ protected function addDiscounts( $totals, &$cartSubmissionData, $quote = null )
...
-
-                $cartSubmissionData['discounts'][] = array(
+                $data = array(
                     'amount'      => $discountAmount,
                     'description' => $totals[$discount]->getTitle(),
                     'type'        => 'fixed_amount',
                 );
+                if ($discount === 'discount' && $quote->getCouponCode()) {
+                    /////////////////////////////////////////////////////////////
+                    /// We want to get the apply the coupon code as a reference.
```
daisy1754: why we need to set `reference` field? Is this just for our own reference?

mccottry: Hello, Kazuki.  For our own reference is a great reason in and of itself on several fronts.  Yes, for that reason, certainly.  In fact, if we have the ability store the reference to the coupon, or the shipping method, or an item, it may considered quite silly or a missed opportunity to omit it.

The primary reason that we are interested in the reference is that in pre-auth, we are tracking if values have changed and attempting to exactly report which element, if any, has changed.  For this, we look each element up, including coupons.  For certain, searching for a coupon identified by `primary key "39488420"` is much more reliable for look up than a title like `"Bob's Crazy Random Weekly Discount!!!"`

PR #362, opened 2019-05-05
v2.0.0 - micro PR progress tracker

mccottry:
This PR is used to track the progress of the micro PRs

When no diffs exist in the two branches, 2.0.0 merge to develop will be complete.


mccottry:
went from 37 files to 28 after mPR.01.  Super!

PR #363, opened 2019-05-05
mPR.03 - Adds new bolt statuses for pre auth orders

mccottry:
To allow for distinct handling of Bolt pre-auth orders, we introduce two new statuses.

https://app.asana.com/0/1118494470563243/1121207201774769

- `pending_bolt`, which is in the standard Magento state of `pending_payment`, and
- `canceled_bolt`, which is in the standard Magento state of `canceled`


daisy1754:
Why can't we just use magento's built in statuses? `pending_payment` and `cancelled` seems great fit for preauth usecase

mccottry:
> Why can't we just use magento's built in statuses? `pending_payment` and `cancelled` seems great fit for preauth usecase

@daisy1754 
Magento order **STATE**, not status, is what is important for Magento business logic.
Status, in Magento, is merely an alias for the various states while adding contextual or meta-data.

A distinct, Bolt namespaced status is fundamental in our treating of Bolt orders in special way while in this vulnerable pending_payment state that many merchants will otherwise not be prepared to handle or accept.  It allows us to easily identify our orders in these orders.  Equally as important, it aids us in prevent other plugins from manipulating our orders which is real world critical.

Another major benefit that we will see in a later pull request is that this allows us to optionally filter these messages from the Magento Admin which allows us to add this version to merchants while exposing no new order states.  This will significantly reduce merchant resistance to our new change.

PR #367, opened 2019-05-07
Cleanup unused generated session quotes on PDP

johnmccombs1:
Asana task: https://app.asana.com/0/564264490825835/1120842637218639

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Cron.php
+++ b/app/code/community/Bolt/Boltpay/Model/Cron.php
@@ -25,7 +25,7 @@ class Bolt_Boltpay_Model_Cron
     use Bolt_Boltpay_BoltGlobalTrait;
 
     /**
-     * After an immutable quote has existed for 2 weeks or more, we remove it from the system.
+     * After an immutable quote / unused generated session quote on PDP has existed for 2 weeks or more, we remove it from the system.
      * At this point, only the order object is relevant for converted orders and any immutable
      * quote that was to be converted will have been handled well before this time.
      *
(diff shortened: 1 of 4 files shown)


app/code/community/Bolt/Boltpay/sql/bolt_boltpay_setup/upgrade-1.4.1-1.4.1.1.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+$installer = Mage::getResourceModel('sales/setup', 'sales_setup');
+$installer->startSetup();
+
+Mage::log('Installing Bolt 1.4.1.1 updates', null, 'bolt_install.log');
+
+if (!$installer->getConnection()->tableColumnExists($installer->getTable('sales/quote'), "is_from_pdp")) {
+    $installer->addAttribute(
+        "quote",
+        "is_from_pdp",
```
daisy1754: let's add "bolt" so it's clear this is from bolt plugin. how about `is_bolt_pdp`?

app/code/community/Bolt/Boltpay/sql/bolt_boltpay_setup/upgrade-1.4.1-1.4.1.1.php
```diff
@@ -0,0 +1,39 @@
+<?php
```
daisy1754: can you confirm if this script works when we upgrade 1.4.1 to 2.0.0 as well?

johnmccombs1: Yes

app/code/community/Bolt/Boltpay/sql/bolt_boltpay_setup/upgrade-1.4.1-1.4.1.1.php
johnmccombs1: Updated.

johnmccombs1:
> @johnmccombs1
> This is a great cron job extension. In your `upgrade-1.4.1-1.4.1.1.php`, I'm thinking that it is a good idea to add a one time cleanup code for all product page orders that exist in the system prior to the new field that you are adding. I think it is a pretty safe bet for quotes that are saved with an empty cart. It will not affect any current sessions.

@mccottry I'm going to create a one time use script to handle this case so we don't end up with code that always wipes out empty quotes for new merchants. I've created a separate task for this. Let me know if you need anything else for PR.

PR #371, opened 2019-05-08
mPR.06 - Include coupon in shipping and tax estimate 

mccottry:
Correctly add coupon in shipping calculations and events at all key shipping and tax moments to remove the need to override Bolt core methods

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/ShippingAndTax.php
+++ b/app/code/community/Bolt/Boltpay/Model/ShippingAndTax.php
@@ -61,7 +61,7 @@ class Bolt_Boltpay_Model_ShippingAndTax extends Bolt_Boltpay_Model_Abstract
             . (@$shippingAddress->street_address3 ?: '') . "\n"
             . (@$shippingAddress->street_address4 ?: '')
         );
-            
+
         $addressData = array(
             'email' => @$shippingAddress->email ?: $shippingAddress->email_address,
             'firstname' => @$shippingAddress->first_name,
@@ -138,10 +138,9 @@ class Bolt_Boltpay_Model_ShippingAndTax extends Bolt_Boltpay_Model_Abstract
         );
 
         try {
-            $this->boltHelper()->collectTotals(Mage::getModel('sales/quote')->load($quote->getId()));
             $originalCouponCode = $quote->getCouponCode();
-
             if ($parentQuote) $quote->setCouponCode($parentQuote->getCouponCode());
+            $this->boltHelper()->collectTotals(Mage::getModel('sales/quote')->load($quote->getId()), true);
 
             //we should first determine if the cart is virtual
             if($quote->isVirtual()){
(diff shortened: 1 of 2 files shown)


app/code/community/Bolt/Boltpay/Model/ShippingAndTax.php
```diff
@@ -216,10 +226,20 @@ public function getShippingAndTaxEstimate( $quote )
      * @param Mage_Sales_Model_Quote $quote    Quote which has been updated to use new shipping rate
      * @param string $shippingRateCode         Shipping rate code composed of {carrier}_{method}
      */
-    public function applyShippingRate($quote, $shippingRateCode) {
+    public function applyShippingRate($quote, $shippingRateCode, $clearTotalsCollectedFlag = true ) {
```
daisy1754: where you pass false?

app/code/community/Bolt/Boltpay/Model/ShippingAndTax.php
```diff
@@ -202,12 +203,21 @@ public function getShippingAndTaxEstimate( $quote )
                 );
 
                 $response['shipping_options'][] = $option;
+
+                Mage::dispatchEvent(
+                    'bolt_boltpay_shipping_option_added',
```
daisy1754: we already have top-level event for `bolt_boltpay_filter_shipping_and_tax_estimate`. do we need per-option event?

app/code/community/Bolt/Boltpay/Model/ShippingAndTax.php
mccottry: In pre-auth order creation, it seems this flag is set to false in the `create_order` method.  Though the code is well tested, I can not logically think of why we would ever do harm by clearing the pre-calculated totals, and it theoretically always seems good to do this.  All other calls use the default true, so, I will remove this.

app/code/community/Bolt/Boltpay/Model/ShippingAndTax.php
mccottry: In general, we are going to want to be very liberal about the inclusion of events in our system.  There have been several integrations that did not take into the long-term impact of the changes that were being made, and our goal is to provide for every opportunity possible for the integration engineer to not resort to bad design choices.

Events, as you mentioned, are a separate topic in its entirety and will hold back what is intended to be a quick and simple PR.   Instead of going file by file, which is some ways is easier and in some ways present some challenges in breaking up the code, I will go with the logical task division breaking down the code, which makes tracking progress and completion a bit more difficult, but should help us focus on one concept at a time.

Events are being removed for this PR and will be added to a future PR to speed up this process.

daisy1754: thanks. btw I'm against adding event when we don't have concrete use case. If we add event it'll be public contract we have to maintain, let's only add when only needed, also we should be careful about what events are suitable to expose

PR #374, opened 2019-05-09
mPR07 - Uses M1 standard naming and stucture convention for auxillary controller classes

mccottry:
Use M1 convention for auxillary controller classes


app/code/community/Bolt/Boltpay/controllers/Adminhtml/Sales/Order/CreateController.php
```diff
@@ -16,7 +16,6 @@
  */
 
 require_once(Mage::getModuleDir('controllers','Mage_Adminhtml').DS.'Sales'.DS.'Order'.DS.'CreateController.php');
-require_once(Mage::getModuleDir('controllers','Bolt_Boltpay').DS.'OrderControllerTrait.php');
```
matt-thomason: Does this `require_once` need to be replaced for the renamed trait?

mccottry: @matt-thomason It is not necessary.  autoload with find it with the new name.

PR #375, opened 2019-05-09
 mPR.08 - DRYs out common webhook functionality

mccottry:


app/code/community/Bolt/Boltpay/controllers/OnepageController.php
```diff
@@ -27,69 +27,15 @@ class Bolt_Boltpay_OnepageController
 {
     use Bolt_Boltpay_BoltGlobalTrait;
 
-    public function saveShippingMethodAction()
```
daisy1754: I'm confused. why this file needs to be updated when it doesn't rely on new trait?

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * @category   Bolt
+ * @package    Bolt_Boltpay
+ * @copyright  Copyright (c) 2018 Bolt Financial, Inc (https://www.bolt.com)
+ * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+/**
+ * Trait Bolt_Boltpay_Controller_Traits_ApiControllerTrait
+ *
+ * Defines generalized actions associated with API calls to and from the Bolt server
```
daisy1754: this is only for API call *from* Bolt server, no?

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * @method Mage_Core_Controller_Response_Http getResponse()
+ * @method Mage_Core_Model_Layout getLayout()
+ */
+trait Bolt_Boltpay_Controller_Traits_ApiControllerTrait {
+    use Bolt_Boltpay_BoltGlobalTrait;
+    /**
+     * @var string The body of the request made to this controller
+     */
+    protected $payload;
```
daisy1754: our request is always JSON. I wonder if we should put json_decoding in this trait as well

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * @var string The signed payload which used the stores signing secret
+     */
+    protected $signature;
```
daisy1754: I don't think this has to be instance variable. you can inline this within function

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * @var bool determines if JSON is expected return type for preDispatch optimization.
+     */
+    protected $willReturnJson = true;
```
daisy1754: Bolt API always expect JSON I believe. do we need this?

app/code/community/Bolt/Boltpay/controllers/OnepageController.php
```diff
@@ -27,69 +27,15 @@ class Bolt_Boltpay_OnepageController
...
-                }
-
-                Mage::helper('boltpay')->collectTotals($this->getOnepage()->getQuote())->save();
-                $this->getResponse()->setBody(Mage::helper('core')->jsonEncode($result));
-            }
-        } catch (Exception $e) {
-            Mage::helper('boltpay/bugsnag')->notifyException($e);
-            throw $e;
-        }
+        $this->willReturnJson = false;
```
daisy1754: this class doesn't use ApiTrait. Why these flags need to be set?

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * @var bool mandates that all request to this controller must be signed
+     */
+    protected $requestMustBeSigned = true;
```
daisy1754: same as above; this must always be true

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * POST data in response to a request
+     *
+     * @param int   $httpCode       standard HTTP response code
+     * @param array $data           a PHP array to be encoded as JSON and sent as a response body to Bolt
+     * @param bool  $doJsonEncode   instructs whether to JSON encode the data, true by default
+     *
+     * @throws Zend_Controller_Response_Exception if the error code is not within the valid range
+     */
+    protected function sendResponse($httpCode, $data = array(), $doJsonEncode = true )
```
daisy1754: I think it'll be simpler if we remove $doJsonEncode and always do json encoding

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -174,16 +160,18 @@ public function indexAction()
...
         set_time_limit(30);
         ignore_user_abort(true);
 
         /** @var Mage_Sales_Model_Quote $quote */
         $quote = Mage::getSingleton('checkout/session')->getQuote();
 
         if(!$quote->getId() || !$quote->getItemsCount()){
-            $this->getResponse()->clearAllHeaders()->setHeader('Content-type', 'application/json');
-            $this->getResponse()->setBody("{}");
+            $this->sendResponse(
```
daisy1754: not directly related to this PR but shouldn't this be error?

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -228,8 +216,11 @@ public function prefetchEstimateAction()
         }
 
         $response = Mage::helper('core')->jsonEncode(array('address_data' => $addressData));
-        $this->getResponse()->clearAllHeaders()->setHeader('Content-type', 'application/json');
-        $this->getResponse()->setBody($response);
+        $this->sendResponse(
+            200,
+            $response,
```
daisy1754: inline `array('address_data' => $addressData)` and we can remove 3rd arg

app/code/community/Bolt/Boltpay/controllers/ApiController.php
```diff
@@ -31,26 +31,7 @@ public function hookAction()
     {
```
daisy1754: maybe I don't understand how trait works; don't you have to specify that this class use ApiTrait?

app/code/community/Bolt/Boltpay/controllers/OnepageController.php
mccottry: @daisy1754 Remember, we are doing micro pull request.  There is no perfect way to break down so many interdependent changes.  Trying entire files will have too many changes, so we will focus in on ideas.  The focus here is API Trait capabilities.

This controller/file will rely on api trait capabilities in a later Pull Request.  Specifically, this is the controller that creates the success page that takes a Bolt payload and Bolt signature as parameters.  By default, however, we want it to behave like a standard controller... not necessarily returning JSON and definitely not requiring a signature.

For now, it is just important to understand that it has been enabled to behave with API controller traits.

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
mccottry: I can improve the language here.  I should use the words "Web hooks" rather than Api.  I will move this to say "Defines generalized actions associated with web hooks."

This includes from (receiving and processing), and to (response).

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
mccottry: JSON encoding is in the method sendResponse, which is part of the trait.

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
mccottry: If this was commonly derived for an HMAC header and was always set, then I would follow that approach and just pull this in the verification method.  However, since there are distinct instances where the signature is not in the HMAC header, (e.g. in the success page the signature is encoded as  a URL get parameter), it is much more elegant to have this as an instance variable accessible throughout the inheriting classes.

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
mccottry: If this was always true, I wouldn't include it.  There are various instances requiring different values, especially when thinking long-term
- Must be signed - Web Hook, Shipping and Tax,  Onepage Controller (OPC) Success Page
- Must be accessible without being signed - OPC all other endpoints, OPC non-Bolt and 1.x Bolt Success Page

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
mccottry: Yes.  There are a variety of situations, especially when thinking big picture, long-term.  Short-term, immediate, we have
- returns JSON - Web Hook, Shipping And Tax
- returns HTML - Success Page

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
mccottry: `$doJsonEncode` is an option parameter and last in the list of parameters.  It can be ignored in all cases and conveniently produce an encoded JSON response.  Removing it adds no value.  Keeping it adds the value of removing the need for silly code in the case that json exists that is already encoded.  We don't want to be encouraging this sort of code.

`$this-sendResponse( 422, json_decode({"point":"This is silly"}) );`

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
mccottry: It is related and included because `sendResponse` is an ApiControllerTrait method.

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
mccottry: Yes, I know and I considered that.  I decided against it because it would attempting to be too smart for no good reason and potentially causing problems because of it.  I came to that after taking a quick look at the internals of that Magento jsonEncode method.

```
    /**
     * Encode the mixed $valueToEncode into the JSON format
     *
     * @param mixed $valueToEncode
     * @param  boolean $cycleCheck Optional; whether or not to check for object recursion; off by default
     * @param  array $options Additional options used during encoding
     * @return string
     */
    public function jsonEncode($valueToEncode, $cycleCheck = false, $options = array())
    {
        $json = Zend_Json::encode($valueToEncode, $cycleCheck, $options);
        /* @var $inline Mage_Core_Model_Translate_Inline */
        $inline = Mage::getSingleton('core/translate_inline');
        if ($inline->isAllowed()) {
            $inline->setIsJson(true);
            $inline->processResponseBody($json);
            $inline->setIsJson(false);
        }

        return $json;
    }
```
and then `Zend_Json::encode`

```
    /**
     * Encode the mixed $valueToEncode into the JSON format
     *
     * Encodes using ext/json's json_encode() if available.
     *
     * NOTE: Object should not contain cycles; the JSON format
     * does not allow object reference.
     *
     * NOTE: Only public variables will be encoded
     *
     * NOTE: Encoding native javascript expressions are possible using Zend_Json_Expr.
     *       You can enable this by setting $options['enableJsonExprFinder'] = true
     *
     * @see Zend_Json_Expr
     *
     * @param  mixed $valueToEncode
     * @param  boolean $cycleCheck Optional; whether or not to check for object recursion; off by default
     * @param  array $options Additional options used during encoding
     * @return string JSON encoded object
     */
    public static function encode($valueToEncode, $cycleCheck = false, $options = array())
    {
        if (is_object($valueToEncode)) {
            if (method_exists($valueToEncode, 'toJson')) {
                return $valueToEncode->toJson();
            } elseif (method_exists($valueToEncode, 'toArray')) {
                return self::encode($valueToEncode->toArray(), $cycleCheck, $options);
            }
        }

        // Pre-encoding look for Zend_Json_Expr objects and replacing by tmp ids
        $javascriptExpressions = array();
        if(isset($options['enableJsonExprFinder'])
           && ($options['enableJsonExprFinder'] == true)
        ) {
            /**
             * @see Zend_Json_Encoder
             */
            #require_once "Zend/Json/Encoder.php";
            $valueToEncode = self::_recursiveJsonExprFinder($valueToEncode, $javascriptExpressions);
        }

        // Encoding
        if (function_exists('json_encode') && self::$useBuiltinEncoderDecoder !== true) {
            $encodedResult = json_encode($valueToEncode);
        } else {
            #require_once 'Zend/Json/Encoder.php';
            $encodedResult = Zend_Json_Encoder::encode($valueToEncode, $cycleCheck, $options);
        }

        //only do post-proccessing to revert back the Zend_Json_Expr if any.
        if (count($javascriptExpressions) > 0) {
            $count = count($javascriptExpressions);
            for($i = 0; $i < $count; $i++) {
                $magicKey = $javascriptExpressions[$i]['magicKey'];
                $value    = $javascriptExpressions[$i]['value'];

                $encodedResult = str_replace(
                    //instead of replacing "key:magicKey", we replace directly magicKey by value because "key" never changes.
                    '"' . $magicKey . '"',
                    $value,
                    $encodedResult
                );
            }
        }

         return $encodedResult;
    }
```

Like above, this PR is not concerned with changing long-standing logic.  It's goal is merely to use the common ApiControllerTrait interface using the existing logic.

app/code/community/Bolt/Boltpay/controllers/OnepageController.php
mccottry: @daisy1754 Remember, we are doing micro pull request. There is no perfect way to break down so many interdependent changes. Trying entire files will have too many changes, so we will focus in on ideas. The focus here is API Trait capabilities.

This controller/file will rely on api trait capabilities in a later Pull Request. Specifically, this is the controller that creates the success page that takes a Bolt payload and Bolt signature as parameters. By default, however, we want it to behave like a standard controller... not necessarily returning JSON and definitely not requiring a signature.

For now, it is just important to understand that it has been enabled to behave with API controller traits and that in the 1.x context, and success page context, we do not want to declare this as returning JSON.

app/code/community/Bolt/Boltpay/controllers/ApiController.php
mccottry: Good catch!!!
I made an error here.  Breaking these mPRs down is very demanding and I missed porting that over.

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
daisy1754: but you can write

> $this-sendResponse( 422, array("point" => "This is not silly") ); 

right? 

I don't think human should write raw JSON. Let's write object and let `json_encode` function handle JSON encoding. 
Within this PR, all instances you already use object or you can easily replace it with object

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
daisy1754: I'm confused. are you saying you don't want to change this because `Mage::helper('core')->jsonEncode` and `json_encode` behave differently? If so it seems to me it's another good reason to centralize all json encode logic inside `sendResponse` method.

mccottry:
> I think success page is only example of APIs hosted by that is meant to be called from Browser. I don't think we should share trait there.

@daisy1754 
I disagree.

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
mccottry: @daisy1754
I disagree. We must extend that class to support our API request that is outside the context of the session. That extension exhibits the exact APIController traits while it is not an API controller.  This is the exact reason traits exist.

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
mccottry: I'm saying we don't know the reason why `Mage::helper('core')->jsonEncode` is being used here.  There could be a valid reason, and there could not.  I would normally not just take a chance on things like this, but, assessing risk and likelihood, it is likely someone here just was being "smart".  There are more important things to defend than json encode and a optional parameter so I will change these.

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
mccottry: ok. no problem.

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
daisy1754: I'm confused. are you saying we have to extend success page controller so it can be read from Bolt's API request? I thought it's always called from user session

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
mccottry: There is no need for a wrapper method around JSON encode.

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
mccottry: It is true that the success URL is called from the user's session, however, it relies on session variables which are set during order creation.  Order creation happens outside of the user's session, so these variables are not set and available via direct call to the Magento standard class.  To support this, we must extend the method to support reading what would normally be session variables from encoded URL parameters.

app/code/community/Bolt/Boltpay/controllers/ApiController.php
mccottry: `use` statement added.

app/code/community/Bolt/Boltpay/controllers/OnepageController.php
mccottry: The use statement was accidently omitted in the "cherry-picking" like approach to making these pull request.  I corrected the omission.

app/code/community/Bolt/Boltpay/controllers/ApiController.php
mccottry: use statement added

app/code/community/Bolt/Boltpay/controllers/OnepageController.php
mccottry: use statement added.

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
mccottry: Actually, I agree with this statement.  You are certainly right that something exhibiting webhook will always need to be signed.  I will update the code and the trait to be as such.  Also, in thinking about how people often override OnePage Checkout, I am going to restructure the override that happens to utilize traits in a way that makes it very simple for integration to extend without overriding any core methods.

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
mccottry: I've removed the optional parameter while not taking a the risk that I listed above in order to find the middle ground.

app/code/community/Bolt/Boltpay/Controller/Traits/ApiControllerTrait.php
mccottry: I was blind, but now I see.  signature removed as class data member.

app/code/community/Bolt/Boltpay/Controller/Traits/WebHookTrait.php
```diff
@@ -0,0 +1,101 @@
...
+ * Defines generalized actions associated with web hooks
+ *
+ * @method Mage_Core_Controller_Response_Http getResponse()
+ * @method Mage_Core_Model_Layout getLayout()
+ */
+trait Bolt_Boltpay_Controller_Traits_WebHookTrait {
+    use Bolt_Boltpay_BoltGlobalTrait;
+
+    /**
+     * @var string The body of the request made to this controller
```
daisy1754: you expect json as input, why not json_decode here?

mccottry: @daisy1754 I will take it as a sign that attempting to implement this nitpick is leading to spectacular  unit test failures:
https://circleci.com/gh/BoltApp/bolt-magento1/2799?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
https://circleci.com/gh/BoltApp/bolt-magento1/2803?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Moreover, we want the payload to be an honest representation of what the Bolt server sent in the request body.  Bolt is sending a json string, not a decoded PHP object.

However, for convenience, I will provide a utility public accessor method.

mccottry: > you expect json as input, why not json_decode here?

@daisy1754 I will take it as a sign that attempting to implement this nitpick is leading to spectacular  unit test failures:
https://circleci.com/gh/BoltApp/bolt-magento1/2799?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
https://circleci.com/gh/BoltApp/bolt-magento1/2803?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Moreover, we want the payload to be an honest representation of what the Bolt server sent in the request body.  Bolt is sending a json string, not a decoded PHP object.

However, for convenience, I have provided a utility public accessor method.

PR #376, opened 2019-05-10
mPR.09 - Remove obsoleted Order Fixer patch

mccottry:
https://app.asana.com/0/1118494470563243/1122570041435559

Pre-auth order creation validation makes the Order Fixer patch obsolete.  Different from Order fixer, it respects Magento's built-in system for inventory and order creation while not trying to radically alter the order.

This PR removes our temporary Order Fixer solution.


mccottry:
> https://app.asana.com/0/1118494470563243/1122570041435559
> 
> Pre-auth order creation validation makes the Order Fixer patch obsolete. Different from Order fixer, it respects Magento's built-in system for inventory and order creation while not trying to radically alter the order.
> 
> This PR removes our temporary Order Fixer solution.

@daisy1754 
The code is currently backwards compatible.  I will be very vocal with everyone if there is a change that is presented that is not.  This is not one of those changes.

"Order Fixer" code is not needed and it is very dangerous code that requires the client's consent.  "Order Fixer" says Bolt couldn't get the order right, and so we are going to override what is in Magento and put in what Bolt got.  This was a compromise that we made for a single M1 merchant or two before we moved to pre-auth.  This is not what we want in M1 or any plugin.  I think this is a very simple PR to move forward.

PR #380, opened 2019-05-15
Associate order to customer when placing on product detail page

johnmccombs1:
Asana task: https://app.asana.com/0/564264490825835/1123239410780665

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Order.php
+++ b/app/code/community/Bolt/Boltpay/Model/Order.php
@@ -315,6 +315,10 @@ class Bolt_Boltpay_Model_Order extends Bolt_Boltpay_Model_Abstract
         );
         ///////////////////////////////////////////////////////
 
+        if ($immutableQuote->getData('is_bolt_pdp') && Mage::getSingleton('customer/session')->isLoggedIn()) {
+            $this->associateOrderToCustomerWhenPlacingOnPDP($order->getData('increment_id'));
+        }
+
         if ($sessionQuoteId) {
             $checkoutSession = Mage::getSingleton('checkout/session');
             $checkoutSession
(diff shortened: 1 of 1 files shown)


app/code/community/Bolt/Boltpay/Model/Order.php
```diff
@@ -331,6 +335,24 @@ public function createOrder($reference, $sessionQuoteId = null, $isAjaxRequest =
...
+     * Associate order to customer when placing on product detail page
+     * @param $orderIncrementId
+     */
+    protected function associateOrderToCustomerWhenPlacingOnPDP($orderIncrementId){
+        $customer = Mage::getSingleton('customer/session')->getCustomer();
+
+        $order = Mage::getModel('sales/order')->loadByIncrementId($orderIncrementId);
+        $order->setCustomerId($customer->getId())
+            ->setCustomerEmail($customer->getEmail())
+            ->setCustomerFirstname($customer->getFirstname())
```
daisy1754: not sure if we want to override name. looks like we don't do this for standard bolt checkout?

johnmccombs1: > is it possible to add test?

@daisy1754 I wouldn't recommend it here as the changes are for saving data, and it doesn't really test functionality as much as it tests reading and writing data to the database. Let me know your thoughts on this.

Also, can you explain what you mean by "not sure if we want to override name. looks like we don't do this for standard bolt checkout?"

daisy1754: I was looking at order creation logic for standard bolt checkout - https://github.com/BoltApp/bolt-magento1/blob/f073380d2b8a1b6292a0ceb4155157632830f35b/app/code/community/Bolt/Boltpay/Model/Order.php#L104

Looks like we set name for guest checkout, but we don't set name for logged in checkout (I assume we prefer logged in user name in magento, as opposed to name entered in our checkout?). Not strong opinion here but I think it'd be good if we're consistent.

johnmccombs1: Yes, for guest checkout, it will use the name that the customer uses in the Bolt checkout modal and for logged in customer, it uses the name from the logged in customer (not the one that they input in the Bolt modal). I tested to confirm and the changes I made function the same way that regular Bolt checkout does.

PR #381, opened 2019-05-16
Re-renders button in mini-cart after it is ajax reloaded

mccottry:
Context:
Bolt button disappears from mini cart when quantity is updated (desktop and mobile)

https://app.asana.com/0/inbox/544708310157128/1122886631659761/1123291135314664

### DIFF
--- a/app/design/frontend/base/default/template/boltpay/replace.phtml
+++ b/app/design/frontend/base/default/template/boltpay/replace.phtml
@@ -167,6 +167,12 @@ $additionalClasses = $this->boltHelper()->getAdditionalButtonClasses();
 
         replaceCheckout();
 
+        if (typeof Minicart != "undefined") {
+            Minicart.prototype.hideOverlay = function(){
+                onDomReady();
+                $j(this.selectors.overlay).removeClass('loading');
+            };
+        }
     };
 
     if (document.addEventListener) {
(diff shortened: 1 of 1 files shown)


app/design/frontend/base/default/template/boltpay/replace.phtml
```diff
@@ -167,6 +167,12 @@ $additionalClasses = $this->boltHelper()->getAdditionalButtonClasses();
 
         replaceCheckout();
 
+        if (typeof Minicart != "undefined") {
+            Minicart.prototype.hideOverlay = function(){
+                onDomReady();
```
daisy1754: let's give this better name now that you call this function outside of `DOMContentLoaded` event. `initializeBoltButton`?

app/design/frontend/base/default/template/boltpay/replace.phtml
```diff
@@ -167,6 +167,12 @@ $additionalClasses = $this->boltHelper()->getAdditionalButtonClasses();
 
         replaceCheckout();
 
+        if (typeof Minicart != "undefined") {
+            Minicart.prototype.hideOverlay = function(){
```
daisy1754: add comment

mccottry: comment added

PR #382, opened 2019-05-16
Add unit test for the bugsnag client 

oldDriver:
The task - https://app.asana.com/0/0/1123361736071012/f


.circleci/config.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       - run:
           name: PHP 5.5 Magento 1.8
           command: |
-            PHPUNIT_PHAR=tests/unit/phpunit-4.8.36.phar MAGENTO_VERSION=magento-mirror-1.8.1.0 tests/scripts/ci.sh
+            PHPUNIT_PHAR=tests/integration/phpunit-4.8.36.phar MAGENTO_VERSION=magento-mirror-1.8.1.0 tests/scripts/ci.sh
```
daisy1754: I think having "integration" test in unit-php55-mage18 build is a bit weird. We have true integration test running webdriver on http://magento1-sandbox.bolt.me so I feel like every test in this directory is "unit" test in a sense it doesn't integrate with real Bolt APIs.

mccottry: As @daisy1754  mentioned, these are unit test, concerned with individual units of code and not concerned with each units integration with the other pieces of the code.  Integration is concerned with inter-dependencies, unit test are not.

tests/phpcs.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0"?>
```
mccottry: It is probably better to use the official Marketplace acceptance rule set which can be found here:
https://github.com/magento/marketplace-eqp/blob/master/MEQP1/ruleset.xml

We would then be interested in severity 10 level errors run as such
`vendor/bin/phpcs /path/to/your/extension --standard=MEQP1 --severity=10 --extensions=php,phtml`

perhaps even better if this was downloaded live by CI so as to make sure that we keep up to date, but if not, the above might be a good choice.

tests/unit/phpunit.xml
```diff
@@ -1,32 +1,22 @@
...
-         stopOnFailure="false"
-         syntaxCheck="true"
-         bootstrap="./autoload.php">
-    <testsuite name="Boltpay_Unit_Test">
-        <directory>testsuite</directory>
-    </testsuite>
+<?xml version="1.0" encoding="UTF-8"?>
+<phpunit bootstrap="bootstrap.php" colors="true" backupGlobals="true">
+    <testsuites>
+        <testsuite name="Tests location">
```
mccottry: Is `Tests location` a more meaningful name than `Bolt_Unit_Test`?  It does not seem as descriptive to me.

.circleci/config.yml
oldDriver: @mccottry , @daisy1754 you are right - in all documentation about Magento 1.* plugins such tests are called "unit". Really, integration testing is the phase in software testing in which individual software modules are combined and tested as a group. In order to run that "unit" tests we need to install Magento.
Unit testing means that we could test something without installing Magento

mccottry: Great @oldDriver .  Thanks for clarifying.  

For our purposes, we use PHP*Unit* here in are unit testing.  Also, we use the common OO perspective that the unit of focus will be our class methods that do require the PHP libraries but do not require that the methods are run in the context of the full application.  

So, @daisy1754 is requesting that we not make the change in name from `unit` to `integrataion` and I definitely fully agree.  There are already true integration test defined elsewhere and this change will make things a little bit confusing for others.

oldDriver: @mccottry @daisy1754 Could you propose name for real unit tests which could be run without Magento integration.
I could not put new unit test in the one folder with "unit/integration" tests, because they have different boostrap.
I could restore folder name to "unit" for integration tests, and put real unit tests into another folder. But i have no edea how to name it. May be "local", does it make sense?

mccottry: @oldDriver Please try 
1.) restoring the folder name to `unit` and keeping all PHP*Unit* test there.  
2.) Place any new integration test in a separate folder.

When you write new unit or integration test, you can demonstrate how you visualize pure unit test and also proper integration test should be written.

To me, it seems like you are wanting to challenge the validity of our current definition of unit test and integration test, which is that: 
- The units that we are interested in testing are the individual methods which may use standard Magento libraries and DB resources.  
- Integration test are concerned interactions between Bolt, web browser, DBs, Magento, and plugin code in near to true life context

I certainly understand that you would like that all Magento and DB dependencies be stubbed and mocked in unit test.  This point of view is also valid.  I suggest to pursue this, you should open this as a separate issue and present your proposal in Integration Weekly.  Another valid approach to your proposed change is to go test by test and stub out and mock out all DB and Magento dependencies so that they fit your desired definition of unit test.  You will then make PRs for each of these changes.

Hopefully, you can understand what I written above.  If not, feel free to give me a call and I can elaborate.

tests/unit/phpunit.xml
oldDriver: @mccottry This is just a label, not class or namespace

.gitignore
```diff
@@ -1,3 +1,6 @@
-.DS_Store
-.idea
-/vendor
+*
```
daisy1754: Hmm, I'm not big fan of astralisk. why it's needed?

oldDriver: * - ignor all files 
! - exclude from ignor
I noticed that in the Magento we have no /vendor folder and all Botl plugin's folders have name "Bolt", "bolt', "Boltpay" and "bolpay"
Such gitignore file allow to work with git when plugin is integrated
Does it make sense?

tests/phpcs.xml
oldDriver: This lite config file already show a lot of errors. We could replace it later with more strict

.gitignore
daisy1754: I know what astarisk means :) it sounds like you put plugin code within magento directory? you can avoid that by using modman - https://github.com/BoltApp/bolt-magento1#modman-guide

tests/phpcs.xml
mccottry: @oldDriver 
The point I wish to communicate is that our plugin is already in the Magento Marketplace
https://marketplace.magento.com/bolt-bolt.html

It is not optional that we pass the ruleset.xml listed above, it is mandatory that we pass that ruleset at severity level 10.  This is the only critically relevant code sniffing configuration for the M1 plugin.

tests/unit/phpunit.xml
mccottry: Can you make the label be more descriptive like "Bolt_Unit_Test" or "Bolt M1 Plugin Unit Test"?  That label seems to more quickly and clearly explain what is going on to any reader of the code.

.gitignore
mccottry: @oldDriver You have missed and have not been informed that we have already thoroughly focused on this problem and evolved to the modman solution.

Still, I'll give you some feedback on the solution that you presented.  The ignore all (*) except for  double asterisk directory approach will several critical files that do not lie in a sub-directory of something that says `bolt` or `boltpay`, (e.g. layout XML files, top level .md files, .sh files etc.) 

Compare it to the PR that replaced our previous asterisk solution:
https://github.com/BoltApp/bolt-magento1/pull/155/files

The problem with the previous solution was that it was ignoring new files and people were often forgetting to add the new files in their commits.  This PR will re-introduce that problem.

mccottry:
@oldDriver 
I'm not sure how you were assigned to this task, but I see that the title has changed from "Unit tests" to "Add unit test for the bugsnag client"

In that time the PR seems to have attempted to cover many separate things and not in a focused way. So far, the changes have been:
- attempting to rewrite the .gitignore file
- Defining a custom code sniffer ruleset
- moving existing unit test from a folder called 'unit' to one called 'integration'
- Writing unit test for static 3rd party library files (Bugsnag)

From my perspective, the focus seems to be misguided but I will leave that to @daisy1754 to decide.  Regardless, I strongly suggest that you talk with @daisy1754 and @OleksiiOnyshchenko and figure out how to best focus your efforts.  For this PR, it is very unclear to me what you are trying to achieve and that lack of clarity is something that we can work to resolve.

I also recommend that you start over and make at least two separate PRs for   
- Implementing the code sniffer ruleset
- Writing Unit test for a single class.  (I will add that I am having a hard time understanding the value of unit test for static 3rd party solutions, so, I'm not sure that is the best place to start.)

matt-thomason:
This PR is stale and unwieldy. You should take @mccottry 's advice up above and update and break it up. It also did not go through out usual process of approving tasks to do. Closing for now.

PR #384, opened 2019-05-19
mPR.11 - Pre-auth order canceling

mccottry:
When Bolt cancels a pre-auth pending order, we must 
1.) Release all resources like inventory.  This is done via cancel
2.) Remove the order.  This is done via delete
3.) expire the cache to trigger new Bolt order creation, otherwise the expired token will continue to be used by the Magento plugin.  This handles the expiration.

https://app.asana.com/0/1118494470563243/1123688922955296

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Observer.php
+++ b/app/code/community/Bolt/Boltpay/Model/Observer.php
@@ -121,6 +121,24 @@ class Bolt_Boltpay_Model_Observer
         $cartHelper->getCart()->truncate()->save();
     }
 
+    /**
+     * If the session quote has been flagged by having a parent quote Id equal to its own
+     * id, this will clear the cart cache, which, in turn, forces the creation of a new Bolt order
+     *
+     * event: controller_front_init_before
+     *
+     * @param Varien_Event_Observer $observer event contains front (Mage_Core_Controller_Varien_Front)
+     */
+    public function clearCartCacheOnOrderCanceled($observer) {
+        /** @var Mage_Sales_Model_Quote $quote */
+        $quote = Mage::getSingleton('checkout/session')->getQuote();
+        if ($quote && is_int($quote->getId()) && $quote->getId() === $quote->getParentQuoteId()) {
+            Mage::getSingleton('core/session')->unsCachedCartData();
+            // clear the parent quote ID to re-enable cart cache
+            $quote->setParentQuoteId(null);
+        }
+    }
+
     public function sendCompleteAuthorizeRequest($request)
     {
         return $this->boltHelper()->transmit('complete_authorize', $request);
--- a/app/code/community/Bolt/Boltpay/Model/Order.php
+++ b/app/code/community/Bolt/Boltpay/Model/Order.php
@@ -384,6 +384,36 @@ class Bolt_Boltpay_Model_Order extends Bolt_Boltpay_Model_Abstract
         return $rateDebuggingData;
     }
 
+    /**
+     * Removes a bolt order from the system.  The order is expected to be a Bolt order
+     *
+     * @param Mage_Sales_Model_Order $order
+     *
+     * @throws Mage_Core_Exception if the order cannot be canceled
+     */
+    public function removePreAuthOrder($order) {
+        if ($this->isBoltOrder($order)) {
+            if ($order->getStatus() !== 'canceled_bolt') {
+                $order->cancel()->setQuoteId(null)->setStatus('canceled_bolt')->save();
+            }
+            $previousStoreId = Mage::app()->getStore()->getId();
+            Mage::app()->setCurrentStore(Mage_Core_Model_App::ADMIN_STORE_ID);
+            $order->delete();
+            Mage::app()->setCurrentStore($previousStoreId);
+        }
+    }
+
+    /**
+     * Determines whether a given order is owned by Bolt
+     *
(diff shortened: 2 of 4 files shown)


app/code/community/Bolt/Boltpay/Model/Order.php
```diff
@@ -384,6 +384,36 @@ protected function getRatesDebuggingData($rates) {
...
+     *
+     * @throws Mage_Core_Exception if the order cannot be canceled
+     */
+    public function removePreAuthOrder($order) {
+        if ($this->isBoltOrder($order)) {
+            if ($order->getStatus() !== 'canceled_bolt') {
+                $order->cancel()->setQuoteId(null)->setStatus('canceled_bolt')->save();
+            }
+            $previousStoreId = Mage::app()->getStore()->getId();
+            Mage::app()->setCurrentStore(Mage_Core_Model_App::ADMIN_STORE_ID);
```
daisy1754: why you need this?

app/code/community/Bolt/Boltpay/Model/Observer.php
```diff
@@ -121,6 +121,24 @@ public function clearShoppingCart() {
         $cartHelper->getCart()->truncate()->save();
     }
 
+    /**
+     * If the session quote has been flagged by having a parent quote Id equal to its own
```
daisy1754: why do we have to clear the cache? isn't cart content same?

app/code/community/Bolt/Boltpay/Model/Order.php
mccottry: @daisy1754 
I see that you have a question about something but I am not sure what you have a question about.  I also do not see a change request here, but rather you desire an explanation of something.

When you say:
> why you need this?

What is the `this` that you are referring to? 
_personally, I only seem to need air, food, and water.  :)_

daisy1754: why do you need to set store to admin store?

mccottry: @daisy1754 
We must set the store to admin in order to delete the order.  In Magento, the admin area is required to delete this type of entity.  You may review the `Mage_Sales_Model_Order` class and trace the core code back to observe this.  Eventually, you will find this passage:
https://github.com/OpenMage/magento-mirror/blob/magento-1.9/app/code/core/Mage/Sales/Model/Order.php#L2322-L2331

From there, you will see that this code is invoked from an Abstract method:
```
    /**
     * Safeguard func that will check, if we are in admin area
     *
     * @throws Mage_Core_Exception
     */
    protected function _protectFromNonAdmin()
    {
        if (Mage::registry('isSecureArea')) {
            return;
        }
        if (!Mage::app()->getStore()->isAdmin()) {
            Mage::throwException(Mage::helper('core')->__('Cannot complete this operation from non-admin area.'));
        }
    }
```

Also, this will be true for Magento 2:
https://github.com/OpenMage/magento-mirror/blob/magento-1.9/app/code/core/Mage/Sales/Model/Order.php#L2322-L2331

app/code/community/Bolt/Boltpay/Model/Observer.php
mccottry: @daisy1754 
It is better to think of this here from the perspective of quotes and orders.  Remember, we are caching Bolt order tokens which are tied to a specific immutable quote.  Regardless of the content of the quote, what we have done prior to the point cancelling the pre-auth order is that we have converted the quote into an order.  Up to this point, we have not experimented with trying to 
- reuse a single quote to create multiple orders
- potentially have two orders exist using the same quote
- reverting a previously converted quote to the unconverted state

There are several things to be cautious of in pursuing the approach of reverting a quote from an order and not producing multiple duplicate orders and it is something to be pursued thoroughly in a future PR, (e.g. removing immutable quotes from our system and using only the native Magento quote.)  

For now, we do not try to over-complicate the issue and concern ourselves with the problem of quote reuse for multiple orders.  Instead, we live with the footprint of a new unused quote in the event of the customer choosing to refresh his browser after a payment failure.  If the customer does not refresh his browser, there will be no problems

app/code/community/Bolt/Boltpay/controllers/ApiController.php
```diff
@@ -237,4 +237,61 @@ protected function getCaptureAmount($transaction) {
...
+        $orderModel = Mage::getModel('boltpay/order');
+        $order =  $orderModel->getOrderByParentQuoteId($parentQuoteId);
+        if (!$order->isObjectNew()) {
+            $orderModel->removePreAuthOrder($order);
+        }
+        ////////////////////////////////////////////////////////////////////
+        /// We treat Bolt initiated failed payment cancels to be the same as a directive
+        /// to expire the cached immutable quote that is stored in the session,
+        /// otherwise, the Bolt checkout could result in a locked state where the end
+        /// user will repeatedly be told that his cart has expired and to refresh. However,
```
daisy1754: just to clarify I believe `failed` hook is only called for failed payment. If preauth endpoint returns 'cart expired' error, Bolt will not call hook

mccottry: Yes. I know.  There is no need to call a hook at that point as the order will not have been created and there is nothing to cancel.  We are talking here after the Bolt server has received the create_order success reply with the success URL

PR #385, opened 2019-05-20
mPR.12 - Define Bolt Pre-Auth statuses as constants

mccottry:
After a pre-auth pending order has existed for 15 minutes or more, we remove it from the system. 
Additionally, all pre-auth canceled orders should be removed from the system.

https://app.asana.com/0/home/544708310157128/1123841422280777

### DIFF
--- a/app/code/community/Bolt/Boltpay/sql/bolt_boltpay_setup/upgrade-1.4.0-2.0.0.php
+++ b/app/code/community/Bolt/Boltpay/sql/bolt_boltpay_setup/upgrade-1.4.0-2.0.0.php
@@ -33,14 +33,11 @@ $connection = $installer->getConnection();
 $statusTable = $installer->getTable('sales/order_status');
 $statusStateTable = $installer->getTable('sales/order_status_state');
 
-$pendingBolt = Bolt_Boltpay_Model_Payment::TRANSACTION_PRE_AUTH_PENDING;
-$canceledBolt = Bolt_Boltpay_Model_Payment::TRANSACTION_PRE_AUTH_CANCELED;
+$pendingBoltStatusDoesntExists = !is_object($connection->query("SELECT status FROM $statusTable WHERE status = 'pending_bolt'")->fetchObject());
+$pendingBoltMappingDoesntExists = !is_object($connection->query("SELECT status FROM $statusStateTable WHERE status = 'pending_bolt' AND state = 'pending_payment'")->fetchObject());
 
-$pendingBoltStatusDoesntExists = !is_object($connection->query("SELECT status FROM $statusTable WHERE status = '$pendingBolt'")->fetchObject());
-$pendingBoltMappingDoesntExists = !is_object($connection->query("SELECT status FROM $statusStateTable WHERE status = '$pendingBolt' AND state = 'pending_payment'")->fetchObject());
-
-$canceledBoltStatusDoesntExists = !is_object($connection->query("SELECT status FROM $statusTable WHERE status = '$canceledBolt'")->fetchObject());
-$canceledBoltMappingDoesntExists = !is_object($connection->query("SELECT status FROM $statusStateTable WHERE status = '$canceledBolt' AND state = 'canceled'")->fetchObject());
+$canceledBoltStatusDoesntExists = !is_object($connection->query("SELECT status FROM $statusTable WHERE status = 'canceled_bolt'")->fetchObject());
+$canceledBoltMappingDoesntExists = !is_object($connection->query("SELECT status FROM $statusStateTable WHERE status = 'canceled_bolt' AND state = 'canceled'")->fetchObject());
 
 if ($pendingBoltStatusDoesntExists) {
     // Insert statuses
@@ -51,7 +48,7 @@ if ($pendingBoltStatusDoesntExists) {
             'label'
         ),
         array(
-            array('status' => $pendingBolt, 'label' => 'Pending Bolt Authorization'),
+            array('status' => 'pending_bolt', 'label' => 'Pending Bolt Authorization'),
         )
     );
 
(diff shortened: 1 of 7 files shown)


mccottry:
> > After a pre-auth pending order has existed for 15 minutes or more, we remove it from the system. At this point, Bolt should have called the "failed_payment" webhook to do this. We are catching the anomaly cases where this mechanism fails.
> 
> I don't think Bolt is supposed to miss hook delivery. I don't think we should add complicated logic for never happening situation

@daisy1754 
I also have great confidence in the Bolt server-side code as well.  

However, as a fact, the situation has happened in production and is the reason why the code exists.  

If we are saying that we now have 100% confidence that this will not be happening again, (so, for instance, a failed_payment hook is sent by Bolt but never responded to... so Bolt retries indefinitely to cancel the order until it gets the proper response... or pre-auth is switched on and then off and failed hooks are assured to be sent and handled properly, or any other unexpected event with product page orders), then, sure, why not remove it.

On the other hand, this is pretty simple code with small footprint that was created to address a real-world-happening problem.   This, like many other future code implementations are in response to reality integrations and lessons learned from beta deploys.  It is not solely based on what theoretically should happen.  My recommendation is to keep the simple plugin-side safety code.

daisy1754:
> However, as a fact, the situation has happened in production and is the reason why the code exists.

can you point me to the asana task or whatever that documents that incident?

mccottry:
> > However, as a fact, the situation has happened in production and is the reason why the code exists.
> 
> can you point me to the asana task or whatever that documents that incident?

@daisy1754 
There are many long slack discussion threads on this, but the problem first started to show up like this:
https://app.asana.com/0/474713171217346/1119377556486228
https://app.asana.com/0/search/1124580642941269/1120054467871370

Just to be clear, I also have faith in the current plugin code base and that the Bolt server-side code will handle this case with the changes that we made. What I am pointing out is that this code was not added just because I think that it is "cool".  They are guided by real world problems and resolutions.

app/code/community/Bolt/Boltpay/sql/bolt_boltpay_setup/upgrade-1.4.0-2.0.0.php
```diff
@@ -33,11 +33,14 @@
 $statusTable = $installer->getTable('sales/order_status');
 $statusStateTable = $installer->getTable('sales/order_status_state');
 
-$pendingBoltStatusDoesntExists = !is_object($connection->query("SELECT status FROM $statusTable WHERE status = 'pending_bolt'")->fetchObject());
-$pendingBoltMappingDoesntExists = !is_object($connection->query("SELECT status FROM $statusStateTable WHERE status = 'pending_bolt' AND state = 'pending_payment'")->fetchObject());
+$pendingBolt = Bolt_Boltpay_Model_Payment::TRANSACTION_PRE_AUTH_PENDING;
```
daisy1754: In general I'm not big fan of using externally-defined value in migration. If we for some reason change the value of this constant, it'll change content of migration file, causing very hard to detect bug.

daisy1754:
> > > However, as a fact, the situation has happened in production and is the reason why the code exists.
> > 
> > 
> > can you point me to the asana task or whatever that documents that incident?
> 
> @daisy1754
> There are many long slack discussion threads on this, but the problem first started to show up like this:
> https://app.asana.com/0/474713171217346/1119377556486228
> https://app.asana.com/0/search/1124580642941269/1120054467871370
> 
> Just to be clear, I also have faith in the current plugin code base and that the Bolt server-side code will handle this case with the changes that we made. What I am pointing out is that this code was not added just because I think that it is "cool". They are guided by real world problems and resolutions.

I didn't fully follow that task, but looks like neither is related to missing hook?
First one seems plugin responded as success to "failed hook" so Bolt didn't retry. The second one seems stoneedge picked up order before failed hook arrives, which is addressed by a newly-added config of immediately sending failed hook.

app/code/community/Bolt/Boltpay/sql/bolt_boltpay_setup/upgrade-1.4.0-2.0.0.php
mccottry: @daisy1754 
The PR has changed and I see that you did not review the existing code.  Please review the existing code.

daisy1754: I'm confused. what do you mean by existing code? as far as I can tell only change made in this PR after latest commit is using constants `Bolt_Boltpay_Model_Payment::TRANSACTION_PRE_AUTH_PENDING` and `Bolt_Boltpay_Model_Payment:: TRANSACTION_PRE_AUTH_CANCELED` in migration, and that's exactly what I mention in my previous comment.

mccottry: @daisy1754 Actually, the major change was `removed order cleanup code and will trust in hooks` https://github.com/BoltApp/bolt-magento1/pull/385/commits/5e876f22a2c541b1793ade50bc766a673ca13e38

> In general I'm not big fan of using externally-defined value in migration. If we for some reason change the value of this constant, it'll change content of migration file, causing very hard to detect bug.

I did not see this above comment.  :O

Hmm.  Actually, I thought over this quite a bit prior to making the change thinking just as you had written above.  I believe that for poorly written migration code, you are correct.    However, for properly written migration code for updating the value which should always follow the pattern

- *if old row values exist* DELETE
- *if new row values do not exist* INSERT

Or if supported, UPSERT or in MySQL INSERT IGNORE, maybe this shouldn't be a problem.  I was thinking that an author updating table constants should be extremely conscious of this.

Ok.  Sounds good to me.  Let's use hard coded values.

PR #386, opened 2019-05-20
mPR.13 - Extra Config values for 2.0.0

mccottry:
Sets up the filters and validation for the hidden extra config values used in 2.0.0

https://app.asana.com/0/home/544708310157128/1123847373963084

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Admin/ExtraConfig.php
+++ b/app/code/community/Bolt/Boltpay/Model/Admin/ExtraConfig.php
@@ -14,7 +14,9 @@
  * @copyright  Copyright (c) 2019 Bolt Financial, Inc (https://www.bolt.com)
  * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
+
 require_once(Mage::getBaseDir('lib') . DS .  'Boltpay/DataDog/ErrorTypes.php');
+
 /**
  * Class Bolt_Boltpay_Model_Admin_ExtraConfig
  *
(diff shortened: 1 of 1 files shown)


app/code/community/Bolt/Boltpay/Model/Admin/ExtraConfig.php
```diff
@@ -184,15 +216,15 @@ function(hints) {
...
     /**
      * @param $rawConfigValue
      * @param array $additionalParams
      * @return string
      */
     public function filterDatadogKeySeverity($rawConfigValue, $additionalParams = array())
     {
-        $allExtraConfigs = (array)json_decode($this->normalizeJSON(Mage::getStoreConfig('payment/boltpay/extra_options')), true);
-        return (array_key_exists("datadogKeySeverity", $allExtraConfigs)) ? $rawConfigValue : Bolt_Boltpay_Helper_DataDogTrait::$defaultSeverityConfig;
+        return strlen(trim($rawConfigValue)) ? trim($rawConfigValue) : Bolt_Boltpay_Helper_DataDogTrait::$defaultSeverityConfig;
```
johnmccombs1: I don't think the change above will work when we want to disable sending logs to DataDog as default severity will always be used.

mccottry: @johnmccombs1 
As I understand from what you have written is that our method for disabling datadog is setting the `datadogKeySeverity` to an empty string, but if we want the default value, we omit defining that same key.

This is probably not as straight forward as having an enable setting or disabling by not providing a value for ``datadogKey`, but, we can discuss that in a separate PR.  For the sake of moving this PR forward, I'm reverting that above code.

app/code/community/Bolt/Boltpay/Model/Admin/ExtraConfig.php
```diff
@@ -193,6 +228,7 @@ public function filterDatadogKeySeverity($rawConfigValue, $additionalParams = ar
     {
         $allExtraConfigs = (array)json_decode($this->normalizeJSON(Mage::getStoreConfig('payment/boltpay/extra_options')), true);
         return (array_key_exists("datadogKeySeverity", $allExtraConfigs)) ? $rawConfigValue : Bolt_Boltpay_Helper_DataDogTrait::$defaultSeverityConfig;
+        // return strlen(trim($rawConfigValue)) ? trim($rawConfigValue) : Bolt_Boltpay_Helper_DataDogTrait::$defaultSeverityConfig; # preferable way to get default
```
daisy1754: no commented out code please

mccottry: This is here for future reference, but, I'll update the code.

mccottry:
> Do we already have merchants using these configs? If not I'm in general against adding new config because it'll complicate things

Yes, there are.  They have been expressed as useful.

As a side-note, these are unpublished configs that add no extra merchant facing complexity -- hidden precisely to not complicate things and not to expose things to merchants that they don't need to generally be concerned with.

We can and should however add some docs to notion about all extra config values.

daisy1754:
Do we document which hidden feature is used by who? If not, can you add comment here for each config (ideally with reasoning too) then we'll figure out where to document.

Even if it's hidden config from customer standpoint, it's still additional code we have to maintain.

mccottry:
> Do we document which hidden feature is used by who? 

The best place to document merchant specific configs continues to be here:
https://github.com/BoltApp/bolt-magento1-merchants/branches

This is a non-issue from the integration engineers perspective.

No

daisy1754:
you're right we shouldn't put merchant name here. can you please add it here https://app.asana.com/0/0/1124589487720104/f ?

mccottry:
> you're right we shouldn't put merchant name here. can you please add it here https://app.asana.com/0/0/1124589487720104/f ?

@daisy1754 absolutely!

PR #387, opened 2019-05-20
mPR.14 - Allows for option hiding of pre-auth orders from the Magento admin

mccottry:
A sage once told me, 'when doing Magento extension development, the best approach is to follow the old Boy Scout adage - "leave no trace"'.  I don't think he was referring to Datadog or Bugsnag.

Here we present the optional hiding of Bolt pre-auth orders from those listed in the Magento admin.  The goal here is not to clutter that board with orders which may not ultimately become authorized. If the merchant wishes to see these orders, they may turn this on via extra config.

For more context, see:
https://app.asana.com/0/1118494470563243/1123878904292346


mccottry:
> I'll review this after 11,12,13. looks like some change from these PRs are sneaked in here.
@daisy1754 
Maybe not a reason to reject the request.  I'll point this to mPR13 so that you can see the relevant changes.

app/code/community/Bolt/Boltpay/Model/Observer.php
```diff
@@ -298,6 +298,28 @@ public function setInitialOrderStatusAndDetails(Varien_Event_Observer $observer)
...
     }
 
+    /**
+     * Hides the Bolt Pre-auth order states from the admin->Sales->Order list
+     *
+     * event: sales_order_grid_collection_load_before
+     *
+     * @param Varien_Event_Observer $observer Observer event contains an orderGridCollection object
+     */
+    public function filterPreAuthOrders($observer) {
```
daisy1754: let's call this `hidePreAuthOrders` ? 'filter' can be intepreted either way - either you filter (hide) or filter (select).

mccottry: done

PR #388, opened 2019-05-22
Add new discount type

oldDriver:

PR #389, opened 2019-05-22
[WIP ]Multistore parameter

oldDriver:


tests/phpcs.xml
```diff
@@ -0,0 +1,351 @@
+<?xml version="1.0"?>
```
mccottry: This has nothing to do with Multistore support.  It belongs in a separate PR.

mccottry:
I am going to literally copy and paste something that I wrote you in PR #388 so that you can understand the normal procedure that we practice:
------------------------------------------------
@oldDriver
There should be a task in Asana for this that describes what and why you are making these changes. Additionally, your PR description should never be empty. At minimum, it should contain a link to the Asana task that it is addressing. It is even better to describe the changes directly in the description.

This pull request needs the same changes as described here:
https://github.com/BoltApp/bolt-magento1/pull/382#pullrequestreview-238500227
@daisy1754 Points out that 
- we need a description for the PR, and
- we shouldn't mix multiple changes in one PR.  (Here you are mixing codesniffing rulesets with a change for adding multi-store parameters)

You should expect that any PR that do not meet these requirements will be returned to you for update for the same reason.

Finally, all PRs for M1 and M2, at minimum should include @matt-thomason for review. Currently, it is common to add the following people for each platform and adding all is best practice.

M1: @matt-thomason, @mccottry, @johnmccombs1
M2: @matt-thomason, @danac-gs, @YevhenShBolt, @ethanwayda22
Woo: @mpatin(required), @Jiyuan-Bolt, @dumega, @vitaliyreznikov

The above list may change over time. Historically, all PRs for the above platforms also required review from @daisy1754. You may wish to consult with him directly regarding this requirement. For other platforms, also consult with @daisy1754.
------------------------------------------------

Note that while using the label [WIP] or Work-in-Progress, we do not necessarily need to have the reviewers listed above, it may be good practice for you to list all of them in your on-boarding process to develop much faster.  [WIP] implies to the reviewers that we are wanting early feedback on direction but that your work is not yet in final form.

matt-thomason:
Multistore has since been solved. Closing for now.

PR #390, opened 2019-05-23
mPR.15 - Pre-auth order success processing

mccottry:
This handles all things related to successful pre-auth order authorization


app/code/community/Bolt/Boltpay/controllers/ApiController.php
```diff
@@ -238,6 +238,40 @@ protected function getCaptureAmount($transaction) {
...
+     * @return string   The URL for which Bolt is to forward the browser.  It contains variables normally
+     *                  stored as session values as URL parameter
+     *
+     * @throws Mage_Core_Model_Store_Exception  if for any reason the store can not be found to generate the URL
+     */
+    private function createSuccessUrl($order, $immutableQuoteId) {
+        /* @var Mage_Sales_Model_Quote $immutableQuote */
+        $immutableQuote = Mage::getModel('sales/quote')->loadByIdWithoutStore($immutableQuoteId);
+        $recurringPaymentProfiles = $immutableQuote->collectTotals()->prepareRecurringPaymentProfiles();
+        $successUrlPath = $this->boltHelper()->getMagentoUrl(Mage::getStoreConfig('payment/boltpay/successpage')) ?: '/';
```
johnmccombs1: Might be able to clean this up a little by passing the query string parameters as a second argument to getMagentoUrl instead of concatenating the string like this? See the example here https://github.com/BoltApp/bolt-magento1/pull/348/files#diff-8a9f070472f68b882d410cebf95f5bd3R213

Not that important. Just saw it here and on line 260 and thought I'd point it out.

Otherwise LGTM

mccottry: Thanks!  I'll update it.

app/code/community/Bolt/Boltpay/controllers/ApiController.php
```diff
@@ -238,6 +238,40 @@ protected function getCaptureAmount($transaction) {
...
+
+        if ($successUrlPath[strlen($successUrlPath) - 1] === '/' ) $successUrlPath = substr( $successUrlPath, 0, -1);
+
+        $successUrlQueryString = "?lastQuoteId={$immutableQuote->getParentQuoteId()}&lastSuccessQuoteId={$immutableQuote->getParentQuoteId()}&lastOrderId={$order->getId()}&lastRealOrderId={$order->getIncrementId()}";
+
+        $recurringPaymentProfilesIds = array();
+        /** @var Mage_Payment_Model_Recurring_Profile $profile */
+        foreach((array)$recurringPaymentProfiles as $profile) {
```
daisy1754: given you already have immutableQuoteID in URL, why do we have to pass these recurringpaymentprofileiDs?

app/code/community/Bolt/Boltpay/Model/Observer.php
```diff
@@ -57,66 +56,35 @@ public function setBoltUserId($observer)
...
+     * event: bolt_boltpay_authorization_after
      *
-     * @param $observer
-     * @throws Exception
+     * @param Varien_Event_Observer $observer Observer event contains `quote`, `order`, and the bolt transaction `reference`
+     *
+     * @throws Mage_Core_Exception if the bolt transaction reference is an object instead of expected string
      */
-    public function verifyOrderContents($observer)
+    public function completeAuthorize($observer)
```
daisy1754: why is this called via bolt_boltpay_authorization_after event? given this is mostly operation around order to me it seems you could have this in Order.php

tests/unit/testsuite/Bolt/Boltpay/Model/ObserverTest.php
```diff
@@ -150,123 +164,12 @@ public function testAddMessageWhenCapture()
...
-        ));
-
-        $observerModel->verifyOrderContents($observerObject);
-
-    }
-
-    /**
-     * @inheritdoc
-     */
-    public function testSendOrderEmail()
```
daisy1754: looks like test around sending email was removed without good replacement?

mccottry: I created a replacement

app/code/community/Bolt/Boltpay/Model/Order.php
```diff
@@ -352,6 +352,97 @@ protected function getQuoteById($quoteId)
...
+     *
+     * @throws Mage_Core_Exception if there is a problem retrieving the bolt transaction reference from the payload
+     */
+    public function receiveOrder( $orderIncrementId, $payload ) {
+        /** @var Mage_Sales_Model_Order $order */
+        $order = Mage::getModel('sales/order')->loadByIncrementId($orderIncrementId);
+        $payloadObject = json_decode($payload);
+        $immutableQuote = $this->getQuoteFromOrder($order);
+
+        Mage::dispatchEvent('bolt_boltpay_order_received_before', array('order'=>$order, 'payload' => $payloadObject));
```
daisy1754: As I said in https://github.com/BoltApp/bolt-magento1/pull/396
1) please document these events in README
2) I'm against adding events unless they are absolute necessary

app/code/community/Bolt/Boltpay/Model/Order.php
```diff
@@ -352,6 +352,97 @@ protected function getQuoteById($quoteId)
...
 
+    /**
+     * Called after pre-auth order is confirmed as authorized on Bolt.
+     *
+     * @param string $orderIncrementId  customer facing order id
+     * @param object $payload           payload sent from Bolt
+     *
+     * @throws Mage_Core_Exception if there is a problem retrieving the bolt transaction reference from the payload
+     */
+    public function receiveOrder( $orderIncrementId, $payload ) {
```
daisy1754: if user is under bad network or close the page, this method may not be called even when payment was successful. Where do we handle such case?

app/code/community/Bolt/Boltpay/Model/Observer.php
```diff
@@ -185,119 +77,16 @@ public function addMessageWhenCapture($observer)
...
     }
-
-    /**
-     * Updates the Magento's interpretation of the Bolt transaction status on order status change.
-     * Note: this transaction status is not necessarily same as order status on the Bolt server.
-     * bolt_transaction_status field keeps track of payment status only from the magento plugin's perspective.
-     *
-     * @param $observer
-     */
-    public function updateBoltTransactionStatus($observer)
```
daisy1754: why can we remove this now?

mccottry: This is repeated, unnecessary code.  The webhook path handles this via the `Bolt_Boltpay_Model_Payment::handleTransactionUpdate` method.

https://github.com/BoltApp/bolt-magento1/blob/v2.0.0/mPR.15.order-success-processing/app/code/community/Bolt/Boltpay/Model/Payment.php#L426-L513

app/code/community/Bolt/Boltpay/Model/Order.php
mccottry: > 1.) please document these events in README
As you said, that README will be done as part of PR #396.  This PR, #390, doe not handle that

> 2.) I'm against adding events unless they are absolute necessary
Great.  To move away from core code overrides like `activateOrder` and `setBoltUserId` which we will not be giving developers the opportunity to do, we must give as much entry points at key events as possible and being generous in allowing a means to gracefully update behavior in a forward moving compatible way is preferred to limiting extend-ability by providing no means update at the method level.

app/code/community/Bolt/Boltpay/Model/Order.php
mccottry: Fortunately, this was thought about long ago and code was implement for this.
https://github.com/BoltApp/bolt-magento1/blob/975f73f7ffd27b8ba702653aaeb1aea841adca43/app/code/community/Bolt/Boltpay/controllers/ApiController.php#L71-L87


Unfortunately, by nature of the mPRs, that code is in the mPR that follows this and we did not catch the request to change this from being based on the event `bolt_boltpay_authorization_after` to a direct method call.  That update will be made in mPR.16

mccottry: updated.

PR #393, opened 2019-05-24
mPR.16 - Pre-auth order creation and validation

mccottry:
Pre-auth order creation and validation. 
- This code is not intended to be backwards compatible with 1.x.  Backwards compatibility is handled in a separate PR

https://app.asana.com/0/inbox/544708310157128/1124555236880429/1124421857405593


app/code/community/Bolt/Boltpay/Model/Order.php
```diff
@@ -285,49 +256,321 @@ public function createOrder($reference, $sessionQuoteId = null, $isAjaxRequest =
...
+     * Validates coupon codes
+     *
+     * @param Mage_Sales_Model_Quote $immutableQuote    Magento copy of Bolt order data
+     * @param object                 $transaction       Bolt copy of order data
+     *
+     * @throws Bolt_Boltpay_OrderCreationException  when coupon fails validation
+     */
+    protected function validateCoupons(Mage_Sales_Model_Quote $immutableQuote, $transaction) {
+
+        if (@$transaction->order->cart->discounts) {
```
daisy1754: let's early return if discounts is empty

app/code/community/Bolt/Boltpay/Model/Order.php
```diff
@@ -285,49 +256,321 @@ public function createOrder($reference, $sessionQuoteId = null, $isAjaxRequest =
...
+                throw new Bolt_Boltpay_OrderCreationException(
+                    OCE::E_BOLT_ITEM_PRICE_HAS_BEEN_UPDATED,
+                    OCE::E_BOLT_ITEM_PRICE_HAS_BEEN_UPDATED_TMPL,
+                    array($cartItem->getProductId(), $boltPrice, $magentoCalculatedPrice)
+                );
+            }
+        }
+
+        /////////////////////////////////////////////////////////////////////////
+        /// Historically, we have honored a price tolerance of 1 cent on
```
daisy1754: pre-auth is a great opportunity for us to do this right; can we be strict here?

daisy1754: maybe we can ignore (with warning log) some element's mismatch as long as total matches

app/code/community/Bolt/Boltpay/controllers/ApiController.php
```diff
@@ -209,27 +176,83 @@ public function hookAction()
...
+            $transaction = $this->getRequestData();
+            $immutableQuoteId = $this->boltHelper()->getImmutableQuoteIdFromTransaction($transaction);
+
+            /** @var  Bolt_Boltpay_Model_Order $orderModel */
+            $orderModel = Mage::getModel('boltpay/order');
+            $order = $orderModel->getOrderByQuoteId($immutableQuoteId);
+
+            if ($order->isObjectNew()) {
+                /** @var Mage_Sales_Model_Order $order */
+                $order = Mage::getModel('boltpay/order')->createOrder($reference = null, $sessionQuoteId = null, $isPreAuthCreation = true, $transaction);
```
daisy1754: use $orderModel ?

app/code/community/Bolt/Boltpay/Model/Order.php
```diff
@@ -285,49 +256,321 @@ public function createOrder($reference, $sessionQuoteId = null, $isAjaxRequest =
...
+
+        if (abs($totalMismatch) > $priceFaultTolerance) {
+            throw new Bolt_Boltpay_OrderCreationException(
+                OCE::E_BOLT_CART_HAS_EXPIRED,
+                OCE::E_BOLT_CART_HAS_EXPIRED_TMPL_GRAND_TOTAL,
+                array($boltGrandTotal, $magentoGrandTotal)
+            );
+        } else if ($totalMismatch) {
+            // Do order total correction if necessary so that the bottom line matches up
+            $order->setTaxAmount($order->getTaxAmount() + ($totalMismatch/100))
```
daisy1754: maybe call save just in case

daisy1754:
given strong demand for this release and this being taking very long, I'm ok adding unit test after releasing 2.0

mccottry:
> given strong demand for this release and this being taking very long, I'm ok adding unit test after releasing 2.0

I will add this on Monday and Tuesday.

app/code/community/Bolt/Boltpay/Model/Order.php
mccottry: > pre-auth is a great opportunity for us to do this right; can we be strict here?
One cent is fairly strict considering we have known rounding problems, but we can set the default to zero for new merchants.  Existing merchants should generally remain at 1 cent to not introduce any new noise with the upgrade.

> maybe we can ignore (with warning log) some element's mismatch as long as total matches
This is the current behavior barring the logging.  I've now added warning logs.  We ignore the difference for individual subtotals as long as it does not exceed the overall tolerance.  A final check is implemented for overall price totals here:
https://github.com/BoltApp/bolt-magento1/blob/548420e59342aa979b2f4667968bf614b46a3302/app/code/community/Bolt/Boltpay/Model/Order.php#L338-L409

app/code/community/Bolt/Boltpay/Model/Order.php
mccottry: aha.  I remember now, this is during the event: `sales_model_service_quote_submit_before`

By definition, these are actions performed precisely before the order is saved.  We do not want to save from this event.  We must respect that other plugins may link into this as well and perform some actions pre-save.  Magento will do the saving here. 

This is the same code section that is described above:
https://github.com/BoltApp/bolt-magento1/blob/548420e59342aa979b2f4667968bf614b46a3302/app/code/community/Bolt/Boltpay/Model/Order.php#L338-L409

app/code/community/Bolt/Boltpay/controllers/ApiController.php
mccottry: yep

PR #399, opened 2019-05-28
Ignore apple pay prefill

matt-thomason:
https://app.asana.com/0/941895179897714/1121581354248363/f

Quick fix to ignore dummy data the Apple Pay prefills. This occurs when someone clicks on Apple Pay, then cancels out of the apple pay order and uses Bolt Checkout. Many users just keep the dummy data there. Example attached below.

Screen Shot 2019-05-28 at 11 33 31 AM

### DIFF
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -347,7 +347,11 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
             $hints['virtual_terminal_mode'] = true;
         }
 
-        $hints['prefill'] = $prefill;
+        // Skip pre-fill for Apple Pay related data.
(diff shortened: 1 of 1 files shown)


app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -347,7 +347,11 @@ private function getAddressHints($quote, $checkoutType)
             $hints['virtual_terminal_mode'] = true;
         }
 
-        $hints['prefill'] = $prefill;
+        // Skip pre-fill for Apple Pay related data.
+        if (!$prefill['email'] == 'fake@email.com' && $prefill['phone'] == '1111111111') {
```
johnmccombs1: I don't think this will work as intended. The negation should appear in the comparison operator, or we should group the comparisons and negate the set. I think this update should accomplish the goal:
`if (!($prefill['email'] == 'fake@email.com' && $prefill['phone'] == '1111111111'))
`

matt-thomason: Shoot good catch.

PR #405, opened 2019-06-05
Removes unnecessary and error generating calls to Bolt Fetch by  the first successful call

mccottry:
The subsequent calls to the the fetch endpoint were failing.  The extra calls were not needed and we should be reusing the initial call to fetch's transaction

https://github.com/BoltApp/bolt-magento1/blob/bugfix/remove-unnecessary-calls-to-fetch-bolt-api/app/code/community/Bolt/Boltpay/controllers/ApiController.php#L57

https://github.com/BoltApp/bolt-magento1/blob/bugfix/remove-unnecessary-calls-to-fetch-bolt-api/app/code/community/Bolt/Boltpay/controllers/ApiController.php#L112

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Payment.php
+++ b/app/code/community/Bolt/Boltpay/Model/Payment.php
@@ -15,6 +15,8 @@
  * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 
+use Bolt_Boltpay_Model_Payment as Payment;
+
 class Bolt_Boltpay_Model_Payment extends Mage_Payment_Model_Method_Abstract
 {
     use Bolt_Boltpay_BoltGlobalTrait;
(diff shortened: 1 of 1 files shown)


app/code/community/Bolt/Boltpay/Model/Payment.php
```diff
@@ -444,7 +444,7 @@ public function handleTransactionUpdate(
         $newTransactionStatus,
         $prevTransactionStatus,
         $transactionAmount = null,
-        $transaction = null
```
daisy1754: This PR changes behaviour when handleTransactionUpdate is called w/o $transaction (eg., https://github.com/BoltApp/bolt-magento1/blob/2371d2b4e1633411962bf905a440c01f51a951c4/app/code/community/Bolt/Boltpay/Model/Payment.php#L422) 

Should we fetch transaction if caller is not passing it?

mccottry: @daisy1754 
The above call here is ok.  The behavior will be the same for placing a transaction on hold.  Your idea is full proof however and I will add this.



`handleTransactionUpdate` does not fetch the transaction.  If it is called without the transaction, no fetching will occur.  

The caller of `handleTransactionUpdate`  is responsible for deciding if fetching the transaction from bolt is necessary.  This method is normally called from the webhook controller.

mccottry: done

app/code/community/Bolt/Boltpay/Model/Payment.php
```diff
@@ -573,7 +581,7 @@ public function handleRefundTransactionUpdate(
                         $creditmemo->setPaymentRefundDisallowed(false);
                         $creditmemo->register()->save();
                     }
-                } else if (!$this->isPartialRefundFixingMismatch($order, $transactionAmount)) { // partial refund
+                } else if (!$this->isPartialRefundFixingMismatch($order, $transactionAmount, $transaction)) { // partial refund
```
daisy1754: boltTransaction ?

mccottry: @daisy1754 
Are you requesting that the name be changed?  In this method's signature, this variable is defined as `$transaction`, so there is no error here.

In the other method, the author was using the same variable name, $transaction, for two distinct objects, the Bolt transaction and the Magento transaction object.  I have never been a fan of that sort of reuse, so I changed the name to $boltTransaction for the Bolt transactions. 

The name change could be made, if that is what you are requesting.

mccottry: These objects are not the same type and were using the same variable name `$transaction` within the same method:
https://github.com/BoltApp/bolt-magento1/blob/develop/app/code/community/Bolt/Boltpay/Model/Payment.php#L473-L474

https://github.com/BoltApp/bolt-magento1/blob/develop/app/code/community/Bolt/Boltpay/Model/Payment.php#L502

The `$transaction` in 502, is actually a Bolt transaction object, 
https://github.com/BoltApp/bolt-magento1/blob/develop/app/code/community/Bolt/Boltpay/Model/Payment.php#L664-L667

where the 473-474 are Magento transaction objects.

daisy1754: ah ok I misunderstood this

PR #406, opened 2019-06-06
Fixes error generated when Apple Pay prefill values are not defined

mccottry:
Current code will generate errors in strict environments.   We suppress notices for undefined array indexes to address the problem.

### DIFF
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -348,7 +348,7 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
         }
 
         // Skip pre-fill for Apple Pay related data.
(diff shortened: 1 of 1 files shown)


app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -348,7 +348,7 @@ private function getAddressHints($quote, $checkoutType)
         }
 
         // Skip pre-fill for Apple Pay related data.
-        if (!($prefill['email'] == 'fake@email.com' || $prefill['phone'] == '1111111111')) {
+        if (!(@$prefill['email'] == 'fake@email.com' || @$prefill['phone'] == '1111111111')) {
```
mccottry: @matt-thomason  Do you think that this would be better as an **&&**?  Should both conditions necessarily be true to indicate Apple Pay?

matt-thomason: Either option represents prefilling dummy, non-real data - I think `||` is appropriate.

PR #407, opened 2019-06-06
Simple comprehensive approach to adding multi-store support

mccottry:
https://app.asana.com/0/544708310157130/1124073715655189

Ensure that Magento executes webhook and shipping and tax context of the appropriate store for multi-store approach.

By setting the store at the onset, all subsequent calls to config or any other store specific behaviors will use the store which is set instead of global default.


mccottry:
> Does this invalidate #389?
> Did you test this out with a test multistore?

Yes. I did have the opportunity to test this out on GrowAce.  Yes, it makes the approach in #389 obsolete.

PR #409, opened 2019-06-07
Fixes: for empty publishableKey and for "Hide On Checkout" setting

YevhenShBolt:
- remove Bolt button from Magento1 checkout page or OneStepCheckout page if setting "Hide On Checkout" set True
    - do not load script if publishable key is empty.

### DIFF
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -663,11 +663,11 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
 
     /**
      * Gets Publishable Key depending the other checkout type.
-     * -  shopping cart uses multi-step publishable keys
-     * -  firecheckout and onepage checkout uses a payment only publishable key
+     * - shopping cart, product page and etc uses multi-step publishable keys
+     * - firecheckout and onepage checkout uses a payment only publishable key
+     * - adminhtml uses a payment only back-office key
      *
      * @return string
-     * @throws Exception
      */
     public function getPublishableKeyForRoute()
     {
@@ -677,9 +677,11 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
         $checkoutType = static::CHECKOUT_TYPE_MULTI_PAGE;
         if ($routeName === 'adminhtml') {
             $checkoutType = static::CHECKOUT_TYPE_ADMIN;
-        } else if ( ($routeName === 'firecheckout') || ($routeName === 'checkout' && $controllerName === 'onepage') ) {
+        } else if ($routeName === 'checkout' && $controllerName === 'onepage') {
             $checkoutType = static::CHECKOUT_TYPE_ONE_PAGE;
         }
+
+        Mage::dispatchEvent('bolt_checkout_type_for_route', array('checkout_type' => $checkoutType));
 
         return $this->getPublishableKey($checkoutType);
     }
@@ -710,5 +712,12 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
         $isFireCheckoutPage = ($this->getRequest()->getRouteName() === 'firecheckout');
 
         return (!$isFireCheckoutPage && $this->isAllowedConnectJsOnCurrentPage());
+    }
+
+    public function getBoltModuleVersion()
+    {
+        $versionElm =  Mage::getConfig()->getModuleConfig("Bolt_Boltpay")->xpath("version");
+
+        return $versionElm[0];
     }
 }
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -683,7 +683,11 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
 
         $checkoutType =   $this->boltHelper()->doFilterEvent(
             'bolt_boltpay_filter_checkout_type',
-            $checkoutType
+            $checkoutType,
+            [
+                'route_name' => $routeName,
+                'controller_name' => $controllerName
+            ]
         );
 
         return $this->getPublishableKey($checkoutType);
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -714,6 +714,9 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
         return (!$isFireCheckoutPage && $this->isAllowedConnectJsOnCurrentPage());
     }
 
+    /**
+     * @return mixed
+     */
     public function getBoltModuleVersion()
     {
         $versionElm =  Mage::getConfig()->getModuleConfig("Bolt_Boltpay")->xpath("version");
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -681,7 +681,10 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
             $checkoutType = static::CHECKOUT_TYPE_ONE_PAGE;
         }
 
-        Mage::dispatchEvent('bolt_checkout_type_for_route', array('checkout_type' => $checkoutType));
+        $checkoutType =   $this->boltHelper()->doFilterEvent(
+            'bolt_boltpay_filter_checkout_type',
+            $checkoutType
+        );
 
         return $this->getPublishableKey($checkoutType);
     }
--- a/app/code/community/Bolt/Boltpay/Model/Observer.php
+++ b/app/code/community/Bolt/Boltpay/Model/Observer.php
@@ -164,6 +164,11 @@ class Bolt_Boltpay_Model_Observer extends Varien_Event_Observer
     }
 
     /**
+     * Checks route name for consistency with `firecheckout`. Firecheckout is 3rd party module and
+     * implement custom checkout for Magento.
+     *
+     * event: bolt_boltpay_filter_checkout_type
+     *
      * @param Varien_Event_Observer $observer
      * @return Varien_Object
      */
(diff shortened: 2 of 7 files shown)


app/code/community/Bolt/Boltpay/Model/Payment.php
```diff
@@ -230,7 +230,20 @@ public function fetchTransactionInfo(Mage_Payment_Model_Info $payment, $transact
...
         if(!empty($quote)) {
-            return $this->boltHelper()->canUseBolt($quote);
+            $canUseBolt = $this->boltHelper()->canUseBolt($quote);
+            $hidePayment = false;
+
+            $routeName= Mage::app()->getRequest()->getRouteName();
+            $moduleName = Mage::app()->getRequest()->getControllerModule();
+            $controllerName = Mage::app()->getRequest()->getControllerName();
+            if (
+                ($routeName === 'onestepcheckout' && $moduleName === 'Idev_OneStepCheckout')
```
mccottry: @YevhenShBolt 

Do you think that we can do something like 
```
$routeName= Mage::app()->getRequest()->getRouteName();
$controllerName = Mage::app()->getRequest()->getControllerName();

$onepageCheckoutUrl = Mage::app()->getUrl('checkout/onepage');
$requestedUrl = $routeName . '/' . $controllerName;

if ($onepageCheckoutUrl === $requestedUrl) {
     $hidePayment = Mage::getStoreConfigFlag('payment/boltpay/hide_on_checkout');
}
```

This may be a way to generically support all one page checkout extensions that override the standard `checkout/onepage`.

By the way, who uses Idev_OneStepCheckout?

matt-thomason: Yevhen created this fix for a live issue with Tonerprice

mccottry: @matt-thomason That makes sense, but it seems like this should be in there custom code as opposed to the plugin which could take the opportunity to solve this in a more robust and universal way.

app/design/frontend/base/default/template/boltpay/track.phtml
```diff
@@ -24,14 +24,19 @@ $controllerName = $this->getRequest()->getControllerName();
...
+        ? 'multi-page' : 'one-page');
+    ?>
+    <?php if (!empty($publishableKey)): ?>
+        <script
+                id="bolt-track"
+                type="text/javascript"
+                src="<?php echo $this->getTrackJsUrl(); ?>"
+                data-publishable-key="<?php echo $publishableKey; ?>">
+        </script>
+        <script>console.log("Bolt Version: <?php echo $version?>");</script>
```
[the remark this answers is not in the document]
YevhenShBolt: Done

YevhenShBolt:
Updates:
- reverted a correct console log message
- added an event to determine checkoutType for publishable key in Block.
- suggested to move 3rd party naming to the another 3rd party module.
- a bit refactored a track template.
-  added event to the method isAvailable in Model/Payment

app/code/community/Bolt/Boltpay/Model/Payment.php
YevhenShBolt: refactored.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -677,10 +677,12 @@ public function getPublishableKeyForRoute()
         $checkoutType = static::CHECKOUT_TYPE_MULTI_PAGE;
         if ($routeName === 'adminhtml') {
             $checkoutType = static::CHECKOUT_TYPE_ADMIN;
-        } else if ( ($routeName === 'firecheckout') || ($routeName === 'checkout' && $controllerName === 'onepage') ) {
+        } else if ($routeName === 'checkout' && $controllerName === 'onepage') {
             $checkoutType = static::CHECKOUT_TYPE_ONE_PAGE;
         }
 
+        Mage::dispatchEvent('bolt_checkout_type_for_route', array('checkout_type' => $checkoutType));
```
mccottry: I like the idea of an event here.  Two things:

1.) We have a pre-built construct here that would be more appropriate in this circumstance that will allow us to work with checkouts that we will not be official supporting.  https://github.com/BoltApp/bolt-magento1/blob/e03a9507faaca49109b15c4862867ba95bda8f93/app/code/community/Bolt/Boltpay/Helper/GeneralTrait.php#L162-L171

You would then use this in place of `Mage::dispactchEvent` and your return as follows:

```
return $this->getPublishableKey(
     $this->boltHelper()->doFilterEvent(
          'bolt_boltpay_filter_checkout_type',
          $checkoutType, 
          ['requested_path' => $requestedPath] # optionally, you could just pass $requestedPath w/o the array
     )
);
```

2.) We now have a very strict process and standard for the addition of any newly introduced events. 
https://www.notion.so/boltteam/Plugin-customization-guide-46d894ba914046da8cd0c533f4644385

YevhenShBolt:
I am a bit refactored code.
I added event listener `bolt_boltpay_filter_checkout_type` and returned firecheckout support for publishableKey.
Also, it will be a great example for other developer how to use event and easy determine checkoutType for another checkout modules.

YevhenShBolt:
@johnmccombs1 as you see we already have here an event and filter as suggests by @mccottry.
The main difference is that i moved the process of determine firecheckout to observer. I think that it will be hollisticly regarding our code and also easy to see that other 3rd party modules can be implemented through event.

johnmccombs1:
> @johnmccombs1 as you see we already have here an event and filter as suggests by @mccottry.
> The main difference is that i moved the process of determine firecheckout to observer. I think that it will be hollisticly regarding our code and also easy to see that other 3rd party modules can be implemented through event.

@YevhenShBolt the main reason to use a filter here is for the return value from the event. The code works and it's not bad, but I know we were looking to follow a new design pattern with this. I'm fine if we decide not to use it here as it may be overkill. @mccottry thoughts?

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -677,10 +677,15 @@ public function getPublishableKeyForRoute()
...
             $checkoutType = static::CHECKOUT_TYPE_ADMIN;
-        } else if ( ($routeName === 'firecheckout') || ($routeName === 'checkout' && $controllerName === 'onepage') ) {
+        } else if ($routeName === 'checkout' && $controllerName === 'onepage') {
             $checkoutType = static::CHECKOUT_TYPE_ONE_PAGE;
         }
 
+        $checkoutType =   $this->boltHelper()->doFilterEvent(
+            'bolt_boltpay_filter_checkout_type',
+            $checkoutType
+        );
```
mccottry: @YevhenShBolt You should give the filter some contextual additional parameters.  In the past code, you constructed a request path, and that seems appropriate.  However, to illustrate the concept using the now existing code, since you used `$routeName` and `$controllerName` as the basis of our standard logic, reason dictates that these same values will be extremely important for someone who needs to use this filter to override the behavior.  Therefore, you should include them in an associative array as the third parameter.

app/code/community/Bolt/Boltpay/Model/Observer.php
```diff
@@ -162,4 +162,21 @@ public function validateBeforeOrderCommit($observer) {
         $orderModel = Mage::getModel('boltpay/order');
         $orderModel->validateBeforeOrderCommit($observer);
     }
+
+    /**
+     * @param Varien_Event_Observer $observer
+     * @return Varien_Object
+     */
+    public function checkCheckoutTypeForFirecheckoutModule($observer)
```
mccottry: We don't want or need to create observers for our own self-created events.  If we find ourselves doing this, the question we should then ask ourselves is in which model or helper would this function better fit?

This pattern only needlessly obscures the control flow logic and makes it harder on code hinting via IDE's

A second observation is that if you are going to be adding event methods, in the comments, leave the name of the event or events that trigger its execution.

app/code/community/Bolt/Boltpay/Model/Payment.php
```diff
@@ -232,7 +232,24 @@ public function fetchTransactionInfo(Mage_Payment_Model_Info $payment, $transact
...
-            return $this->boltHelper()->canUseBolt($quote);
+            $canUseBolt = $this->boltHelper()->canUseBolt($quote);
+            $hidePayment = false;
+
+            $routeName= Mage::app()->getRequest()->getRouteName();
+            $moduleName = Mage::app()->getRequest()->getControllerModule();
+            $controllerName = Mage::app()->getRequest()->getControllerName();
+            $otherRoute = false;
+
+            Mage::dispatchEvent('bolt_is_available_payment', array('other_route' => $otherRoute));
```
mccottry: Here are some things to consider:

1.) It is not logical or normal to have a non-mutating event in a boolean getter method
2.) Even if such an event could be justified, the naming convention is incorrect.  It should begin with `bolt_boltpay_`
3.) A hard-code `false` parameter is of no use.  It appears that you are expecting to mutate it based on the code below.  However, you have only passed a copy of the the value of `$otherRoute` to your array.  You will therefore find that the value of $otherRoute in your if statement below, this will always be false.  Use our `doFilterEvent` construct instead.

mccottry:
> > @johnmccombs1 as you see we already have here an event and filter as suggests by @mccottry.
> > The main difference is that i moved the process of determine firecheckout to observer. I think that it will be hollisticly regarding our code and also easy to see that other 3rd party modules can be implemented through event.
> 
> @YevhenShBolt the main reason to use a filter here is for the return value from the event. The code works and it's not bad, but I know we were looking to follow a new design pattern with this. I'm fine if we decide not to use it here as it may be overkill. @mccottry thoughts?

@YevhenShBolt, @johnmccombs1 
Actually, Yevhen's version will not work as intended as explained above regarding passing `$otherRoute` to an array.  Yevhen, I definitely look at John's example as righteous clean, exemplifying how we want to use filters and style moving forward.  We also can use Notion, README.md, https://docs.bolt.com, or the merchants repository as a source for usage/training examples.  The official release is probably not the place for placing code purely as a teaching purposes.

app/code/community/Bolt/Boltpay/Model/Payment.php
YevhenShBolt: agree. updated.

YevhenShBolt:
@mccottry  @johnmccombs1 Good catch.I made several updates.

> We don't want or need to create observers for our own self-created events.

My main goal was divide our core logic from 3rd party modules. It is why i moved `firecheckout` to the observer instead of leave it directly in code. For my opinion it is more clear than previously and current part of code (determination a `firecheckout` route) divided from Bolt's business logic. The observer have pretty simple logic and i do not see a reason to move it inside helper or model.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -677,10 +677,12 @@ public function getPublishableKeyForRoute()
         $checkoutType = static::CHECKOUT_TYPE_MULTI_PAGE;
         if ($routeName === 'adminhtml') {
             $checkoutType = static::CHECKOUT_TYPE_ADMIN;
-        } else if ( ($routeName === 'firecheckout') || ($routeName === 'checkout' && $controllerName === 'onepage') ) {
+        } else if ($routeName === 'checkout' && $controllerName === 'onepage') {
```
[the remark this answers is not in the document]
YevhenShBolt: It does not remove any official support. I moved a 3rd party support to the observer model where is it place. We can easily add new 3rd party checkouts due the event. It will not break any production code because we have an event now.

YevhenShBolt:
Closed it for now.
I will create two separate PRs.

PR #410, opened 2019-06-10
Added Php code sniffer to the project

oldDriver:
Code style check with the PHP Code Sniffer tool, which validate code with a coding standard


tests/unit/.gitingore
```diff
@@ -1,2 +0,0 @@
-.DS_Store
```
mccottry: This is an unrelated change.  Remove it from this PR.

oldDriver: @mccottry it is typo "gitingor" instead of "gitignore"
May be it make sense stay it here?

mccottry: Open another PR for this.  And create the appropriate task and subtask in Asana.

PR #424, opened 2019-06-19
Allows ability to filter user note or override user note recording behavior

mccottry:
In the event the that a merchant does not wish to use our default customer note prefix, '[CUSTOMER NOTE]', or wishes to override our behavior as in the case of using Amasty Orderattr to store the user note in attributes and displaying it in another location, we provide this filter.

https://app.asana.com/0/544708310157130/1127699336702289


daisy1754:
This seems very specific. How about we remove prefix for all merchants (and potentially add prefix to other text or stop using usernote for non-user note)? To me it seems weird that we have prefix in the first place `$this->setOrderUserNote($order, '[CUSTOMER NOTE] ' . $transaction->order->user_note);`. 

>  Amasty Orderattr to store the user note in attributes and displaying it in another location
This can be done as part of other events, I think we have events for order creation and such

mccottry:
> I think this is fine, but @daisy1754 raises a good point here too. @mccottry

@matt-thomason and @daisy1754 

Magento does not have a native customer note UI.  Bolt adds this feature to M1 via our note interface.  We are using the order history for this and without a label, it is very hard to distinguish what came from the customer.  Our default label accomplishes this, making it standout obvious.

image

image

PR #429, opened 2019-06-24
Provides support for discount price changes at the time of shipping and tax estimations

mccottry:
Problem:
If the discount is changed after shipping and tax, for instance, if the Magento Admin is set to include tax as part of a percentage based discount, then pre-auth validation fails and the recorded Bolt totals will be incorrect.  Currently, this will block an order from being able to be processed.

Proposed Solution:
We have always utilized shipping tax in order to get the correct amount for total tax for similar reasons.  Now, we utilize the shipping cost estimate to include the price discount difference and disable subtotal validation in pre-auth.  We have bottom-line price validation that will catch any errors while we are able to preserve recording the proper tax amount.

This is an unblocker for M1 and core team decides on if and how to expand the shipping and tax endpoint to shipping, tax, and discount.

https://app.asana.com/0/941895179897704/1128289474493189


mccottry:
WIP - pending unit test

PR #430, opened 2019-06-24
BoltOrder.php file fixing warnings from the PHP code sniffer

oldDriver:
After this changes we have only two warnings

--------------------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
--------------------------------------------------------------------------------
 105 | WARNING | Function's cyclomatic complexity (13) exceeds 10; consider
     |         | refactoring the function
 541 | WARNING | Comment refers to a TODO task "evaluate necessity of this code
     |         | by auditing Bugsnag for corrections made"
--------------------------------------------------------------------------------

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/BoltOrder.php
+++ b/app/code/community/Bolt/Boltpay/Model/BoltOrder.php
@@ -23,20 +23,36 @@
  */
 class Bolt_Boltpay_Model_BoltOrder extends Bolt_Boltpay_Model_Abstract
 {
+    /**
+     * @var string
+     */
     const ITEM_TYPE_PHYSICAL = 'physical';
+
+    /**
+     * @var string
+     */
     const ITEM_TYPE_DIGITAL  = 'digital';
-    protected $itemOptionKeys = array('attributes_info', 'options', 'additional_options', 'bundle_options');
+
+    /**
+     * @var array
+     */
+    protected $_itemOptionKeys = array(
+        'attributes_info',
+        'options',
+        'additional_options',
+        'bundle_options'
+    );
 
     /**
      * @var int The amount of time in seconds that an order token is preserved in cache
      */
-    public static $cached_token_expiration_time = 360;
+    public static $cachedTokenExpirationTime = 360;
 
     /**
      * @var array  Store discount types, internal and 3rd party.
      *             Can appear as keys in Quote::getTotals result array.
      */
-    protected $discountTypes = array(
+    protected $_discountTypes = array(
         'discount',
         'giftcardcredit',
         'giftcardcredit_after_tax',
@@ -53,17 +69,19 @@ class Bolt_Boltpay_Model_BoltOrder extends Bolt_Boltpay_Model_Abstract
     /**
      * @var array  list of country codes for which we will require a region for validation to succeed.
      */
-    protected $countriesRequiringRegion = array(
-        'US', 'CA',
+    protected $_countriesRequiringRegion = array(
+        'US',
+        'CA',
     );
 
     /**
      * Generates order data for sending to Bolt.
      *
-     * @param Mage_Sales_Model_Quote        $quote      Magento quote instance
-     * @param bool                          $multipage  Is checkout type Multi-Page Checkout, the default is true, set to false for One Page Checkout
+     * @param Mage_Sales_Model_Quote $quote Magento quote instance
+     * @param bool $multipage  Is checkout type Multi-Page Checkout,
+     * the default is true, set to false for One Page Checkout
      *
-     * @return array            The order payload to be sent as to bolt in API call as a PHP array
+     * @return array The order payload to be sent as to bolt in API call as a PHP array
      */
     public function buildOrder($quote, $multipage)
     {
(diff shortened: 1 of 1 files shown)


mccottry:
@oldDriver 
Let's start like this:
- Please provide us with a link to the Asana task that you are addressing
- Please provide the command that we are running that generates these warnings
- In the Asana task or here, please provide the list of warnings sent out by the code sniffer.

oldDriver:
Asana task https://app.asana.com/0/0/1128279953521955/f
Comm and and list of warnings are in the asana task

oldDriver:
@mccottry 
MEQP1 codding standard is not installed in our system. The installed coding standards are Squiz, PSR1, PEAR, MySource, PSR2 and Zend.
Current phpcs.xml file emulate rules MEQP1, may be not strict but very close.
And the main parameter is "severity"
I could confirm, that with severity = 10 we have no warnings. This is true.
But is it good?
 Let us look into one file changing severity, for example app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
with default severity = 47 warnings
severity=1-6 -> 47 warnings
some warnings:
480 | WARNING | [ ] Visibility must be declared on method "getConfigSelectors"
 489 | WARNING | [ ] Visibility must be declared on method "isBoltOnlyPayment"
 498 | WARNING | [ ] Visibility must be declared on method "isEnableMerchantScopedAccount"
 537 | WARNING | [ ] Visibility must be declared on method "getPublishableKey"


severity=7,8 -> 5 warnings
107 | WARNING | Silencing errors is discouraged
 180 | WARNING | Silencing errors is discouraged
 310 | WARNING | Function's cyclomatic complexity (16) exceeds 10; consider refactoring the function
 351 | WARNING | Silencing errors is discouraged
 351 | WARNING | Silencing errors is discouraged

When we set severity parameter to lower we use more strict rules.
According Magento documentation severity=10 show only critical errors
https://docs.magento.com/marketplace/user_guide/sellers/code-validation.html
Recommended severity level is 6
"A severity level 6 warning does not prevent an extension from passing technical review. Developers are encouraged to resolve any issues that trigger a level 6 warning."
Does it make sense?

matt-thomason:
Hey @oldDriver 
`MEQP1` is installed via composer in our repo - it is part of the `magento/marketplace-eqp` package and is available at `vendor/magento/MEQP1`.

I can't view the Asana task you linked - is it private?

Yes, while severity 10 is the lowest, that is what we are following as per the technical market review guidelines: https://github.com/magento/marketplace-eqp

oldDriver:
Hi, @matt-thomason, @mccottry so, as i understood we don't need to decrease severity.
So i could remove this pull request. My idea was to do code more readable and as the result more flexible, without magic.

mccottry:
> Hi, @matt-thomason, @mccottry so, as i understood we don't need to decrease severity.
> So i could remove this pull request. 

@oldDriver 
There are ways to succeed and positively contribute to a team effort while gaining the trust, respect, and support of others ... and, there are ways to fail miserably.  

Let's focus on the successful option.  In this case, the successful option for this PR, which seems to be concerned with **PHP code sniffer**, is to make sure that we are following the official requirements for the Magento Marketplace.  That means to assure that we are using the official MEQP1 ruleset and not an approximation.  It also means to run it as officially required, which is with these commands: `--standard=MEQP1 --severity=10 --extensions=php,phtml`

> My idea was to do code more readable and as the result more flexible, without magic.

Regarding style changes, you were given a full opportunity to do this: 
https://app.asana.com/0/search/1128939441934017/1126393209784047

You failed to take advantage of the opportunity and I am not sure why.  We can not miss opportunities at clear paths toward success.

Bottom line is that we want to create and release the best code as possible.  To that end, there are many changes that I like in what you are proposing and I am going to reopen the styling task which provides another opportunity for us to properly succeed at that task, which, I must add, is separate from this PR.

oldDriver:
@mccottry You  could compare phpcs.xml file in the repo with MEQP1 ruleset.xml file.
https://github.com/magento/marketplace-eqp/blob/master/MEQP1/ruleset.xml
What are differences? In the current file removed (not added or changed) some special rules, such as MEQP1.Security.InsecureFunction. All other rules are the same.
I could create full list of that rules. What does it mean?  It means that if we add MEQP1 rules and run codesniffer with severity less that 10 (for example 8 or 6) we'll get same warnings and may be more.
So,  idea - current rules are less strict that MEQP1, because of some rules are disabled. 
I tried to do steps to in order to clear code, but you blocked all my attempts.
Using code sniffer with severity=10 does not have sense - it show only errors.
I could do following steps:
- install MEQP1 library
- restore ruleset.xml to strict MEQP1
- run phpcs with severity=10
It will return nothing - no errors
But if i'll decrease severity to 8 i'll get warnings, update pull request which will be blocked by your.
What steps i need to to if i want fixed code style?

mccottry:
@oldDriver 
You asked:
>   What steps i need to to if i want fixed code style?

The answer is the same as written before
>Let's focus on the successful option. In this case, the successful option for this PR, which seems to be concerned with PHP code sniffer, is to make sure that we are following the official requirements for the Magento Marketplace. That means to assure that we are using the official MEQP1 ruleset and not an approximation. It also means to run it as officially required, which is with these commands: --standard=MEQP1 --severity=10 --extensions=php,phtml

So, I would get the official and run it as stated.  While we could attempt to be clever and edit this down to what we think we need, in fact we only make things more difficult if the official standards change.  Better to plug and play with rulesets and let the codesniffer do the filtering than for us to be manually comparing the files after each edit to the ruleset.

matt-thomason:
This PR is abandoned following a discussion on coding standards. Closing this for now - if we want to talk about standards we can do it during mx weekly.

PR #436, opened 2019-07-03
Added data-was-replaced attribute for replaced button

oldDriver:
Asana task: https://app.asana.com/0/899238157173033/1116476869617025/f
This changes prevent double change buttons to Bolt button.
I added attribute, because we could not use style.display attribute of the element because of responsive design

### DIFF
--- a/app/design/frontend/base/default/template/boltpay/replace.phtml
+++ b/app/design/frontend/base/default/template/boltpay/replace.phtml
@@ -83,13 +83,13 @@ $additionalClasses = $this->boltHelper()->getAdditionalButtonClasses();
                                 /**
                                  * Escape double replace buttons
                                 **/
-                                var bolt = element.getAttribute("bolt");
-                                if (bolt != "replaced") {
+                                var wasReplaced = element.getAttribute("data-was-replaced");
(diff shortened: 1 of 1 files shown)


app/design/frontend/base/default/template/boltpay/replace.phtml
```diff
@@ -80,21 +80,29 @@ $additionalClasses = $this->boltHelper()->getAdditionalButtonClasses();
...
-                                <?php endif; ?>
-
-                                element.parentNode.insertBefore(bolt_button, element);
-
-                                if (element.parentNode.style.display === "") {
-                                    element.parentNode.style.display = "block";
+                                /**
+                                 * Escape double replace buttons
+                                **/
+                                var bolt = element.getAttribute("bolt");
```
mccottry: Can we use better variable names and also use the modern HTML `data-` attributes?
```
var wasReplaced = element.getAttribute('data-was-replaced');

if ('wasReplaced' !== 'true') {
   ...
   element.setAttribute("data-was-replaced", "true");
    
```

This way we can have both readable and valid HTML instead of this HTML which will not validate.

PR #439, opened 2019-07-05
Use varien http adapter curl

oldDriver:
Asana task https://app.asana.com/0/0/1129839896925771/f
Magento has Varien_Http_Adapter_Curl. It is a good idea to use it

### DIFF
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -607,14 +607,17 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
         return $locationInfo;
     }
 
+    /**
+     * @param string $url
+     * @return string
+     */
     public function url_get_contents($url)
     {
-        $ch = curl_init();
-        curl_setopt($ch, CURLOPT_URL, $url);
-        curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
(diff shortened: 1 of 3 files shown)


mccottry:
The code in general looks good and I would normally pass this.  However, I am also aware that this task was assigned to @johnmccombs1 and the solution decided was to use Guzzle via the official Bolt PHP API.

https://app.asana.com/0/941895179897714/1103542714570562

Since this is a refactor and is functionally not need, I'm going to defer this to @johnmccombs1 and @matt-thomason 

I am also going to recommend discussing things like this in #magento1-team Slack channel before reaching this point.  I believe this would have been pointed out there and you could have worked something out with John.

This is a similar situation to the CodeSniffer task that you could work out with @matt-thomason 
https://app.asana.com/0/941895179897714/1116581006500421

johnmccombs1:
@oldDriver To @mccottry's point, the code for Guzzle has already been implemented although it has been pending for testing for awhile due to fires that I've been working through with merchants. The implementation that I made was to modify the the Bolt PHP API and not the Bolt Magento API to keep the code clean and as reusable as possible across non-Magento implementations. I would prefer if we use the code I wrote for this. I'll try to get my implementation tested and out this week. I do appreciate you wanting to improve our codebase. In the future pinging us in the #magento1-team slack channel or discussing in the weekly meeting might be a good way to determine if something has already been discussed so we can reduce the amount of rework. Thanks!

matt-thomason:
I'm going to close out this PR in anticipation of @johnmccombs1 's guzzle implementation.

PR #446, opened 2019-07-11
updated integration tests reporter

arieznikov:
This will introduce better reporting for integration tests step
https://app.asana.com/0/746943642536769/1130800126316371/f

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,11 +59,19 @@ jobs:
               git clone git@github.com:BoltApp/integration-tests.git
               cd integration-tests
               npm install
-              mkdir -p /tmp/screenshots
-              SCREENSHOT_DIR=/tmp/screenshots TEST_ENV=sandbox npm run test-checkout-magento1
+              mkdir -p ./screenshots
+              JUNIT_REPORT_DIR=./test-results SCREENSHOT_DIR=./screenshots TEST_ENV=sandbox npm run test-checkout-magento1
             fi
+      - store_test_results:
+          path: ./test-results
       - store_artifacts:
-          path: /tmp/screenshots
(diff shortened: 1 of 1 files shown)


matt-thomason:
@arieznikov Does this have a corresponding Asana task? If so, please link it.

.circleci/config.yml
```diff
@@ -59,11 +59,19 @@ jobs:
               git clone git@github.com:BoltApp/integration-tests.git
               cd integration-tests
               npm install
-              mkdir -p /tmp/screenshots
-              SCREENSHOT_DIR=/tmp/screenshots TEST_ENV=sandbox npm run test-checkout-magento1
+              mkdir -p ./screenshots
```
matt-thomason: Why move the `screenshots` directory? Not that I object, but if we're changing our conventions there should probably be a reason.

.circleci/config.yml
```diff
@@ -59,11 +59,19 @@ jobs:
...
+          path: ./test-results
       - store_artifacts:
-          path: /tmp/screenshots
+          path: ./screenshots
+      - run:
+          name: Generate Report
+          command: npm run generate-report
+          when: always
+      - store_artifacts:
+          path: ./allure-report
```
matt-thomason: Do we need to `mkdir` this directory?

arieznikov: no, it is generated by generate-report

.circleci/config.yml
arieznikov: why is it in tmp? ci creates own space for each sub-task there is no need for tmp.

mccottry:
@arieznikov  Make sure to add reviews as subtask in Asana.  Thanks!

PR #448, opened 2019-07-12
Fix path for export artifacts for integration tests

arieznikov:
https://app.asana.com/0/746943642536769/1130800126316371/f


.circleci/config.yml
```diff
@@ -63,15 +63,17 @@ jobs:
               JUNIT_REPORT_DIR=./test-results SCREENSHOT_DIR=./screenshots TEST_ENV=sandbox npm run test-checkout-magento1
             fi
       - store_test_results:
-          path: ./test-results
+          path: ./integration-tests/test-results
       - store_artifacts:
-          path: ./screenshots
+          path: ./integration-tests/screenshots
```
matt-thomason: Can you update the line `mkdir -p ./screenshots` above?

mccottry:
@arieznikov 
Is this more properly labeled: _**fix path for export artifacts**_ and attributed to https://app.asana.com/0/746943642536769/1131005536528125/f

Is there something broken about these paths or is this just a refactor?

.circleci/config.yml
arieznikov: there is no need for that, mkdir is already happening inside /integration-tests dir

matt-thomason: Gotcha, I see that now.

arieznikov:
@mccottry yes, catch was that step "run" is downloading external repository and diving into it but all next steps defaults context to root directory

PR #450, opened 2019-07-16
Replace curl with guzzle

johnmccombs1:


lib/Boltpay/Init.php
```diff
@@ -0,0 +1,4 @@
+<?php
+require_once dirname(__FILE__) . DIRECTORY_SEPARATOR . 'Guzzle/ApiClient.php';
+require_once(Mage::getBaseDir() . DS .  'vendor/autoload.php');
```
matt-thomason: Looks like this line here is no bueno - killed all the circleci tests. Can you fix @johnmccombs1 ?

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -26,6 +26,7 @@
  * create the order on Bolt side and set up the javascript BoltCheckout.configure process with cart and hint data.
  *
  */
+require_once(Mage::getBaseDir('lib') . DS .  'Boltpay/Init.php');
```
mccottry: This is not needed.  See below.

app/code/community/Bolt/Boltpay/Helper/ApiTrait.php
```diff
@@ -14,7 +14,7 @@
  * @copyright  Copyright (c) 2019 Bolt Financial, Inc (https://www.bolt.com)
  * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-
+require_once(Mage::getBaseDir('lib') . DS .  'Boltpay/Init.php');
```
mccottry: This is not needed.  See below.

app/code/community/Bolt/Boltpay/controllers/ConfigurationController.php
```diff
@@ -20,6 +20,7 @@
  *
  * Check Configuration for Bolt
  */
+require_once(Mage::getBaseDir('lib') . DS .  'Boltpay/Init.php');
```
mccottry: This is not needed.  See below.

lib/Boltpay/Bugsnag/Notification.php
```diff
@@ -171,16 +171,13 @@ public function postJSON($url, $data)
             return;
         }
 
-        // Prefer cURL if it is installed, otherwise fall back to fopen()
-        // cURL supports both timeouts and proxies
-
         try {
-            if (function_exists('curl_version')) {
-                $this->postWithCurl($url, $body);
+            if (method_exists('GuzzleHttp\Client','sendAsync')) {
```
mccottry: Since you are choosing to modify the 3rd party Bugsnag library by removing the `postWithCurl`, which is the aggressive approach, there are a few observations to point out.  

First, since we are including guzzle with our software, then Guzzle must exist, right?  That would make this check unnecessary.  We should always `postWithGuzzle`.

Second, internally, guzzle is performing the same check to see if `curl_multi_exec` and `curl_exec` exists and whether `ini_get('allow_url_fopen')` is set to true.  Therefore, `postWithFopen` would also be unnecessary because Guzzle will do that internally.

https://github.com/guzzle/guzzle/blob/3b0452a3a0347e74c42e9b9a2e9b11158531ce14/src/functions.php#L95-L124

You could also choose to take the more conservative approach which would be to check for `method_exists('GuzzleHttp\Client','sendAsync')` to exist and use it, but keeping the original cascade intact by then checking for `function_exists('curl_version')` and not removing `postWithCurl` or the next check for  `ini_get('allow_url_fopen')` and `postWithFopen`

PR #451, opened 2019-07-17
Remove redundant parallel

arieznikov:
https://app.asana.com/0/436430054069245/1130680516655715/f

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,7 +83,6 @@ workflows:
                 - master
                 - develop
                 - circleci
-                - remove-redundant-parralel
       - unit-php55-mage18
       - unit-php55-mage19
       - unit-php56-mage18
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,20 +48,17 @@ jobs:
     docker:
       - image: circleci/openjdk:latest-node
       - image: selenium/standalone-chrome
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -48,20 +48,17 @@ jobs:
     docker:
       - image: circleci/openjdk:latest-node
       - image: selenium/standalone-chrome
-    parallelism: 2
```
arieznikov: this was causing job to run same tests 2 times, causing increased response times

.circleci/config.yml
```diff
@@ -48,20 +48,17 @@ jobs:
...
     steps:
       - checkout
       - run:
           name: Integration test (only for master/develop/circleci branch)
-          command: |
-            if [ "$CIRCLE_BRANCH" = "master" ] || [ "$CIRCLE_BRANCH" = "develop" ] || [ "$CIRCLE_BRANCH" = "circleci" ]; then
+          command: |
```
arieznikov: this is the same as filter down below

.circleci/config.yml
mccottry: @arieznikov 
This is interesting and a great find.  Would this parallelism be useful in running the unit test?

arieznikov: units are already divided on job level

PR #454, opened 2019-07-18
Add filter if discount applies to shipping

johnmccombs1:
I added a filter to allow custom discount modifications without the need to override the entire addDiscounts function. This is needed in cases where the default handling of discounts needs to be modified.

For example, let's say that a customer has a $100 subtotal in their cart and they have a $1,000 gift card. At the time of page load the calculated discount would be $100 since there is no shipping applied to the cart total yet. But if the discount is allowed to be applied to shipping as well, then when the shipping is applied during the shipping and tax estimation step, the $100 discount won't be enough to cover the shipping even though the customer has enough on the gift card. In this case we actually need to customize the discount amount at page load to be the full amount (e.g. $1,000) instead of the default amount (e.g. $100) so that if the shipping amount is affected by the discount, the discount will have enough to affect it. In the example above let's say that the shipping was $25. A discount of $1,000 can be applied to a $125 grand total whereas the default discount functionality would only apply to $100 leaving $25 still required by the customer to pay.  Historically we would override the addDiscounts function in the local code pool and apply the full amount of the discount there. However, with the new Filter functionality, we can simply apply a filter to customize the discount. This code simply allows us to do that now without needing to override the entire function.

Fixes: https://app.asana.com/0/932047656377869/1108045192467066

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ If you would like to pull the latest Bolt code from the Git repo and update Mage
 
 | Name | Area | Description | Parameters | Filtered Value | 
 | --- | --- | --- | --- | --- |
+| bolt_boltpay_filter_discount_amount | global | Allows changing of individual discount amount that is displayed in Bolt | **quote**<br>_Mage_Sales_Model_Quote_<br>the Magento cart copy of the Bolt order<br><br>**discount**<br>string<br>the Magento totals array index/label for the discount |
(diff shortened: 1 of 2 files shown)


matt-thomason:
@johnmccombs1 The value of this PR is evident from the contents, but there should be some sort of description of the motivation here - either in the PR or in the Asana task (that isn't linked: https://app.asana.com/0/932047656377869/1108045192467066). There is a *little* bit of context in Asana but nothing that leads directly to the action taken in the PR - we should be more explicit going further.

mccottry:
@johnmccombs1  Can you address @matt-thomason's concerns so we can get this merged?

PR #456, opened 2019-07-23
refactor the discount filter code to allow for skipping discount (i.e. amount = 0)

mccottry:
Modifications to PR #454 
This prevents the discount from being processed if the returned filter value is 0


mccottry:
> Looks good. Were there cases where we were applying $0.00 discounts?

We can think of this in two ways.  The first is as in the description above -- that is, if the custom logic determines there is no discount, the processing should not do any further processing for that discount.  This is already in the original code as `&& $amount = $totals[$discount]->getValue()` in the if statement.  So, we are maintaining status quo.

The second way to think of it is that we want to allow away to fully override/skip the default behavior for a special sort of discount -- perhaps something like points, or maybe a negative balance.

mccottry:
@johnmccombs1 I'm going to adjust this code to do an early continue instead of the large if block.

PR #465, opened 2019-07-28
Address Pre-auth timeouted out orders, and v1.x orders

mccottry:
# Description
Adds the concept of orphaned transactions to 2.0.  This does two things:
  1.)  Allows for orders to be created via webhook if the create_order endpoint times out 
  2.)  Allows for non-pre-auth (legacy) orders to be processed by webhook

Additionally, this adds the user experience code of forwarding the user to the success_url when the Bolt create_order mechanism does not provide the forwarding URL

Fixes: https://app.asana.com/0/436430054069245/1132981769648030

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/app/code/community/Bolt/Boltpay/Helper/Data.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Data.php
@@ -164,6 +164,7 @@ class Bolt_Boltpay_Helper_Data extends Mage_Core_Helper_Abstract
                 }
             }"
             : "function(transaction, callback) {
+                window.bolt_transaction_reference = transaction.reference;
                 $successCustom
                 callback();
             }";
@@ -190,16 +191,33 @@ class Bolt_Boltpay_Helper_Data extends Mage_Core_Helper_Abstract
                     }
                     ";
                 break;
-            case Bolt_Boltpay_Block_Checkout_Boltpay::CHECKOUT_TYPE_FIRECHECKOUT:
-                $javascript .=
-                    "
-                    isFireCheckoutFormValid = false;
-                    initBoltButtons();
-                    ";
-                break;
             case Bolt_Boltpay_Block_Checkout_Boltpay::CHECKOUT_TYPE_PRODUCT_PAGE:
                 $quoteId = Mage::getSingleton('checkout/session')->getQuoteId();
                 $successUrl = $this->getMagentoUrl(Mage::getStoreConfig('payment/boltpay/successpage'), array('checkoutType' => $checkoutType, 'session_quote_id' => $quoteId));
+                break;
+            case Bolt_Boltpay_Block_Checkout_Boltpay::CHECKOUT_TYPE_FIRECHECKOUT:
+                $javascript .=
+                    "
+                    $closeCustom
+                    isFireCheckoutFormValid = false;
+                    initBoltButtons();
+                    ";
+                $closeCustom = '';
+            default:
+                $appendChar = (strpos($successUrl, '?') === false) ? '?' : '&';
+
+                $javascript .=
+                    "
+                    $closeCustom
+                    if (window.bolt_transaction_reference) {
+                         setTimeout(
+                              function() {
+                                   window.location = '$successUrl'+'$appendChar'+'bolt_transaction_reference='+window.bolt_transaction_reference;
+                              },
+                              500
+                         );
+                    }
+                    ";
         }
 
         return $javascript;
--- a/app/code/community/Bolt/Boltpay/Helper/Data.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Data.php
@@ -203,19 +203,17 @@ class Bolt_Boltpay_Helper_Data extends Mage_Core_Helper_Abstract
                     initBoltButtons();
                     ";
                 $closeCustom = '';
+                // fall-through
             default:
+                // Backup success page forwarding for Firecheckout, Onepage Checkout, Multi-Checkout/Mini-Cart
+                // Generally all checkouts should fall-through to this
                 $appendChar = (strpos($successUrl, '?') === false) ? '?' : '&';
 
                 $javascript .=
                     "
                     $closeCustom
                     if (window.bolt_transaction_reference) {
-                         setTimeout(
-                              function() {
-                                   window.location = '$successUrl'+'$appendChar'+'bolt_transaction_reference='+window.bolt_transaction_reference;
-                              },
-                              500
-                         );
+                         window.location = '$successUrl'+'$appendChar'+'bolt_transaction_reference='+window.bolt_transaction_reference;
                     }
                     ";
         }
--- a/app/code/community/Bolt/Boltpay/Model/Observer.php
+++ b/app/code/community/Bolt/Boltpay/Model/Observer.php
@@ -122,7 +122,7 @@ class Bolt_Boltpay_Model_Observer
 
             /** @var Mage_Sales_Model_Order $order */
             $order = Mage::getModel('sales/order')->loadByIncrementId($incrementId);
-            $orderEntityId = (!$order->isObjectNew()) ? $order->getId() : 4294967295;  # use required max sentinel id if order not yet created
+            $orderEntityId = (!$order->isObjectNew()) ? $order->getId() : Bolt_Boltpay_Model_Order::MAX_ORDER_ID;  # use required max sentinel id if order not yet created
 
             $requestParams['lastQuoteId'] = $requestParams['lastSuccessQuoteId'] = $immutableQuoteId;
             $requestParams['lastOrderId'] = $orderEntityId;
--- a/app/code/community/Bolt/Boltpay/Model/Observer.php
+++ b/app/code/community/Bolt/Boltpay/Model/Observer.php
@@ -74,37 +74,63 @@ class Bolt_Boltpay_Model_Observer
      */
     public function setSuccessSessionData($observer) {
 
+        /** @var Mage_Checkout_Model_Session $checkoutSession */
+        $checkoutSession = Mage::getSingleton('checkout/session');
         $requestParams = Mage::app()->getRequest()->getParams();
 
         // Handle only Bolt orders
-        if (!isset($requestParams['bolt_payload'])) {
+        if (!isset($requestParams['bolt_payload'])  && !isset($requestParams['bolt_transaction_reference'])) {
             return;
         }
 
-        $payload = @$requestParams['bolt_payload'];
-        $signature = base64_decode(@$requestParams['bolt_signature']);
+        if (isset($requestParams['bolt_payload'])) {
+            $payload = @$requestParams['bolt_payload'];
+            $signature = base64_decode(@$requestParams['bolt_signature']);
 
-        if (!$this->boltHelper()->verify_hook($payload, $signature)) {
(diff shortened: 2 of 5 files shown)


app/code/community/Bolt/Boltpay/Helper/Data.php
```diff
@@ -190,16 +191,33 @@ public function buildOnCloseCallback($closeCustom, $checkoutType)
...
-                $successUrl = $this->getMagentoUrl(Mage::getStoreConfig('payment/boltpay/successpage'), array('checkoutType' => $checkoutType, 'session_quote_id' => $quoteId));
+                $closeCustom = '';
+            default:
+                $appendChar = (strpos($successUrl, '?') === false) ? '?' : '&';
+
+                $javascript .=
+                    "
+                    $closeCustom
+                    if (window.bolt_transaction_reference) {
+                         setTimeout(
```
daisy1754: why setTimeout is need?

app/code/community/Bolt/Boltpay/Helper/Data.php
```diff
@@ -190,16 +191,33 @@ public function buildOnCloseCallback($closeCustom, $checkoutType)
...
+                    $closeCustom
                     isFireCheckoutFormValid = false;
                     initBoltButtons();
                     ";
-                break;
-            case Bolt_Boltpay_Block_Checkout_Boltpay::CHECKOUT_TYPE_PRODUCT_PAGE:
-                $quoteId = Mage::getSingleton('checkout/session')->getQuoteId();
-                $successUrl = $this->getMagentoUrl(Mage::getStoreConfig('payment/boltpay/successpage'), array('checkoutType' => $checkoutType, 'session_quote_id' => $quoteId));
+                $closeCustom = '';
+            default:
```
daisy1754: can you add comment when execution comes default?

app/code/community/Bolt/Boltpay/Helper/Data.php
```diff
@@ -190,16 +191,33 @@ public function buildOnCloseCallback($closeCustom, $checkoutType)
...
                     "
+                    $closeCustom
                     isFireCheckoutFormValid = false;
                     initBoltButtons();
                     ";
-                break;
-            case Bolt_Boltpay_Block_Checkout_Boltpay::CHECKOUT_TYPE_PRODUCT_PAGE:
-                $quoteId = Mage::getSingleton('checkout/session')->getQuoteId();
-                $successUrl = $this->getMagentoUrl(Mage::getStoreConfig('payment/boltpay/successpage'), array('checkoutType' => $checkoutType, 'session_quote_id' => $quoteId));
+                $closeCustom = '';
```
daisy1754: is removing `break` intentional? if so let's add explicit comment eg.,
> // fall-through

matt-thomason:
Please change base branch back to master now that it's all settled :)

app/code/community/Bolt/Boltpay/Helper/Data.php
mccottry: Welcome back, Kazuki!

Context:
```
leon  6 days ago
Out of curiosity, what URL will you try to forward the user to in the case of timeout?

niraj  6 days ago
Great question, we have to rely on frontend redirect again

niraj  6 days ago
We won't have a URL from merchant backend now

leon  6 days ago
Therefore, I think it will be best to handle that from the plugin side

niraj  6 days ago
Yeah it will have to be done from plug-in side

leon  6 days ago
plugin developers will once again have to rely on the BoltCheckout.success/BoltCheckout.close method for fallback

niraj  6 days ago
Basically, I view pre-authorization as an optimistic process but we need to have fallbacks if it fails for basically things that we no longer have

niraj  6 days ago
That's right

leon  6 days ago
sounds good.

leon  6 days ago
In the success callback, be need an indicator to whether pre-auth order creation succeed or if this is a "forced success"

niraj  6 days ago
Well if pre auth order creation succeeded presumably we would have the reidrect url

leon  6 days ago
however, one problem that I always have had and have been silent about is that the BoltCheckout.success is still called -- as it should be.

leon  6 days ago
Without reverting to hacky like code, what this means is that if the success implementations goal is to send a directive to the plugin to create the order the 1.x context, then we are in a bad space.

niraj  6 days ago
Sorry not following that part

leon  6 days ago
I know... that is why I am still typing :slightly_smiling_face:

leon  6 days ago
First and foremost, explanation aside, in the javascript success method, we need an indication of whether this is a pre-auth success or not.

leon  6 days ago
Ideally, the logic works like this in BoltCheckout.success:
-- If this was a successful pre-auth order do nothing (BoltCheckout will forward us)
but
-- if this success was generated any other way perform the necessary logic to create the conditions for a success URL and then forward us to that page

niraj  6 days ago
When we redirect right now from pre auth, do we call the onSuccess?

leon  6 days ago
Without an indication from Bolt at the success method level, instead the plugin developer would have to talk back and forth with the merchant shopping cart server to figure out if it should do the forwarding logic.
As is, if preauth is successful, it is a racing conditions hoping for the modal to forward to the success URL before executing the code in success\close.

niraj  6 days ago
No I don't think it's a race condition I think we always redirect to the success URL first. IIRC we either automatically do the URL redirect or send the message from our iframe indicating we can do onSuccess

leon  6 days ago
I have observed differently in the early days of preauth.

niraj  6 days ago
Okay let me figure this out when I get to a place I can use my laptop

niraj  6 days ago
Thanks again! I'll do some testing later today

leon  6 days ago
Also, you shouldn't redirect before calling the success/onSuccess method.  If that happens, the success method would not be triggered and it violates the contract written in the docs.

leon  6 days ago
Also, the docs are poorly written in the way that we are using to names success and onSuccess to refer to the same method.  In the doc, both terms are used
- https://docs.bolt.com/docs/step-2-load-bolt-checkout
- https://docs.bolt.com/docs/step-2-load-bolt-checkout#section-call-boltcheckout-configure

niraj  6 days ago
Cool yeah maybe I'm wrong I'll take another look

leon  6 days ago
M1, M2, Woo and all early plugins use `success`
That api is changing all the time.  Maybe they are now aliases.
like `BoltCheckout` used to be `BoltConfigure`

leon  6 days ago
I've confirmed that there absolutely will be a race condition between forwarding code placed in the success/onSuccess and close BoltCheckout methods and the forward called as a result of pre-auth.  Furthermore, there is no indication in the transaction parameter passed to the method that this was a natural pre-auth success.
As I've indicated, I can write some JS synchronous/blocking code to achieve what I need, but it would be great to know directly from the success/onSuccess parameters the context in which we are in, if possible.
If we need a call on this, I am available.  Perhaps, Monday will be the best day.

niraj  6 days ago
Got it, yeah Monday or Tuesday is fine I am flying back to SF on Monday night

leon  6 days ago
perfect!
```

So, in summary, we don't want the `BoltCheckout.close` method doing the forwarding in the normal pre-auth context.  There currently is no simple way from the success or close method to determine if this is a pre-auth order and will be automatically forwarded, if it is a pre-auth create_order call that had a timeout or unusual response and needs to be forwarded by plugin code, or if it is an order with the pre-auth flag turned off, which also needs to be forwarded by the plugin.  We we wait (`setTimeout`) a reasonable and very small amount of time, a half second, to allow BoltCheckout the opportunity to forward the page to the natural success URL.  If after that amount of time that directive has not come, it is certain that we are in an unusual context (e.g. pre-auth flag turned off, create_order timeout or unusual response), and then we execute the code to forward the browser.

**Comment is being added to the code**

app/code/community/Bolt/Boltpay/Helper/Data.php
mccottry: yes, I can.

app/code/community/Bolt/Boltpay/Helper/Data.php
mccottry: Yes, it is intentional.

app/code/community/Bolt/Boltpay/Model/Observer.php
```diff
@@ -74,38 +74,64 @@ public function clearCartCacheOnOrderCanceled($observer) {
      */
     public function setSuccessSessionData($observer) {
 
+        /** @var Mage_Checkout_Model_Session $checkoutSession */
+        $checkoutSession = Mage::getSingleton('checkout/session');
         $requestParams = Mage::app()->getRequest()->getParams();
 
         // Handle only Bolt orders
-        if (!isset($requestParams['bolt_payload'])) {
+        if (!isset($requestParams['bolt_payload'])  && !isset($requestParams['bolt_transaction_reference'])) {
```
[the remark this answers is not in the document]
mccottry: @matt-thomason 
It should be `&&`.  Neither of the conditions should be true for an early return.  And in practice, only one of the two things will be true for Bolt orders.

If I wanted to wrap the whole method body instead of doing an early return, I could use `||` like this
```
if (isset($requestParams['bolt_payload']) || isset($requestParams['bolt_transaction_reference'])){
     // ... all the rest of the method
}
```

But, there has been a lot of preference in our code for the early return pattern, so it is done the way above for this reason.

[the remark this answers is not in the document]
matt-thomason: I misread the lack of `!` in the following if statements, got it. Reviewed too late in the night, sorry 👍

app/code/community/Bolt/Boltpay/Helper/Data.php
daisy1754: I'm still unclear on this one. what happens if you havepre auth on, order was created successfully but redirect takes longer than 0.5 sec?

app/code/community/Bolt/Boltpay/Model/Observer.php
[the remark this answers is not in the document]
mccottry: i feel that late night working.

app/code/community/Bolt/Boltpay/Helper/Data.php
mccottry: BoltCheckout is written and performing excellently, so:
1.) This has never been observed to happened so far.
2.) It would imply something faulty in BoltCheckout that would need fixing.

But, currently this is not the case.

If, however, this scenario somehow managed to occur, 

- that is to say that pre-auth order creation succeeded at Magento 
- and the Bolt server side received confirmation and then
- directed the plugin to call the `success`, and therefore invoke the `callback()` function
- and then call `close`
- but somehow got delayed in initiating the success URL forwarding

if all this occured, then at this point the plugin code will no longer be concerned with the pre-auth flow and the front-end will handle the success screen by utilizing the `transaction` element passed to the `success` function.  At the shopping cart platform server-side, the relevant session information will be taken from fetching the full Bolt transaction.  And finally, considering that the order has already been created, the success screen will then be displayed in the same way as in pre-auth forward, with the same URL, but different URL parameters utilizing the already created order.

Effectively, we get the same less optimized outcome with the extra processing time coming from the time it takes to fetch a transaction from the Bolt server and the less-than-a-blink-of-an-eye 0.5 second mishap that somehow occurred within the BoltCheckout JS, (which has never been observed to occur).

To further explore this, the M1 code is written in such a way to always deliver a customer facing successful outcome.  Of course, the ideal and most speedy path is with pre-auth success and BoltCheckout initiated forwarding, however, regardless of the hiccups, the M1 software is anticipated to resolve gracefully.

PR #467, opened 2019-07-29
Remove authcapture parameter

johnmccombs1:
# Description
Currently configuration for auth capture can be set on magento backend, and this will override the Bolt Merchant admin configuration. Removing this will allow us to use the configuration on the merchant admin.

Fixes: https://app.asana.com/0/564264490825835/1132958531677445

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -130,18 +130,14 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
      * Generate cart data
      *
      * @param      $orderCreationResponse
-     * @param null $checkoutType
      * @return array
      */
-    public function buildCartData($orderCreationResponse, $checkoutType = null)
+    public function buildCartData($orderCreationResponse)
     {
-        $authCapture = $checkoutType === 'admin' ? true : Mage::getStoreConfigFlag('payment/boltpay/auto_capture');
-
         //////////////////////////////////////////////////////////////////////////
         // Generate JSON cart and hints objects for the javascript returned below.
         //////////////////////////////////////////////////////////////////////////
         $cartData = array(
-            'authcapture' => $authCapture,
             'orderToken' => ($orderCreationResponse) ? $orderCreationResponse->token: '',
         );
 
--- a/app/code/community/Bolt/Boltpay/Controller/Traits/OrderControllerTrait.php
+++ b/app/code/community/Bolt/Boltpay/Controller/Traits/OrderControllerTrait.php
@@ -79,7 +79,7 @@ trait Bolt_Boltpay_Controller_Traits_OrderControllerTrait {
 
         if (!$cart_data) {
             $immutableQuote = $this->cloneQuote($sessionQuote, $checkoutType);
-            $cart_data = $boltpayCheckout->buildCartData($boltOrder->getBoltOrderToken($immutableQuote, $checkoutType),$checkoutType);
+            $cart_data = $boltpayCheckout->buildCartData($boltOrder->getBoltOrderToken($immutableQuote, $checkoutType));
         }
 
         $boltOrder->cacheCartData($cart_data, $sessionQuote, $checkoutType);
--- a/app/code/community/Bolt/Boltpay/etc/system.xml
+++ b/app/code/community/Bolt/Boltpay/etc/system.xml
@@ -46,16 +46,6 @@
               <show_in_website>1</show_in_website>
               <show_in_store>1</show_in_store>
             </test>
-            <auto_capture translate="label">
-              <label>Automatic Capture Mode</label>
-              <comment>If enabled, payment is automatically captured when Bolt approves orders. Otherwise, payment is captured when you create an invoice.</comment>
-              <frontend_type>select</frontend_type>
-              <source_model>adminhtml/system_config_source_yesno</source_model>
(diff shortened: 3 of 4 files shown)


mccottry:
@johnmccombs1  Why is this still a draft AND more interestingly, HOW DID YOU MAKE THIS A DRAFT?

johnmccombs1:
> @johnmccombs1 Why is this still a draft AND more interestingly, HOW DID YOU MAKE THIS A DRAFT?

@mccottry I found this option https://www.screencast.com/t/vwox9c3L3uj9. I created it this way so I could review the new template before publishing. I was trying to determine if I should add a unit test for this before publishing it.

matt-thomason:
Draft PRs are really neat and a relatively new feature 😄 https://github.blog/2019-02-14-introducing-draft-pull-requests/

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -135,8 +135,13 @@ protected function isAdminAndUseJsInAdmin($checkoutType)
      */
     public function buildCartData($orderCreationResponse, $checkoutType = null)
     {
-        $authCapture = $checkoutType === 'admin' ? true : Mage::getStoreConfigFlag('payment/boltpay/auto_capture');
-
+        if($checkoutType === 'admin') {
```
matt-thomason: This is good but we have so many `extraConfig` settings. I'm going to get Mitch's opinion here on to how to best document these externally.

matt-thomason:
@johnmccombs1 Have you tested how this interacts with the autocapture setting for the backoffice division in the admin dashboard?

mpatin:
I thought we wanted to move away from the plugins defining `auto-capture` vs `auth&capture`? I believe if we don't define a value, the division setting will be used.

mccottry:
@mpatin, if this is the way we are going, we should probably remove this from the plugins while everyone is upgrading, right?

johnmccombs1:
> @mpatin, if this is the way we are going, we should probably remove this from the plugins while everyone is upgrading, right?
> 
> @matt-thomason @johnmccombs1
> Also, I don't think that this is an Extra-Config value. If we are going to add it, it should be prominently displayed in the Magento admin. A similar setting that could use this move is Bolt button color. I think of Extra Config as something that I really don't want to have to explain to the merchant. It should only come up out of absolute necessity.

@mccottry I'm ok with this being an admin setting. I made it an Extra Config because I assumed that there was a good reason for not allowing it in the admin to begin with. It seemed odd that this is the first time that a merchant requested it so I assumed that not allowing it was status quo. I'll go ahead and make it a setting in the admin as I agree that it's odd that we have a setting for frontend but not admin.

mpatin:
@mccottry @johnmccombs1 - I agree, we should begin to remove it from all plugins. It's not something that should be changing regularly, and I believe it's more straightforward to centralize as and a Merchant Division setting. Also having it in two places can and will cause confusion about which one takes precedence.

Prior to removing it for all plugins and merchants, we need to verify that merchants don't have differing configurations in the plugin settings vs. the admin dashboard.

https://app.asana.com/0/1127468708335972/1115005290123321/f

johnmccombs1:
> @mccottry @johnmccombs1 - I agree, we should begin to remove it from all plugins. It's not something that should be changing regularly, and I believe it's more straightforward to centralize and a Merchant Division setting. Also having it in two places can and will cause confusion about which one takes precedence.
> 
> Prior to removing it for all plugins and merchants, we need to verify that merchants don't have differing configurations in the plugin settings vs. the admin dashboard.
> 
> https://app.asana.com/0/1127468708335972/1115005290123321/f

@mpatin just to confirm, we are opting to have this handled in the admin dashboard rather than in Magento. If that's the case then I'll remove this PR and confirm it works with the Bolt admin config.

mpatin:
@johnmccombs1 Yep, let's take out the auto-capture logic in for backoffice orders. Can you update (instead of remove) this PR to reflect this change? That way we keep the discussion above attached with the code change.

mpatin:
@matt-thomason is verifying that all M1 backoffice merchant divisions have autocatpure enabled in the Bolt db (shown in Admin). This will ensure that if other merchants update, the capture behavior will stay the same.

https://docs.google.com/spreadsheets/d/1HVWQgpW0Xd70Qo6afAFeTl7Vqg1o8chXg9i6XG-Qy8I/edit#gid=0

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -135,16 +135,17 @@ protected function isAdminAndUseJsInAdmin($checkoutType)
...
-
         //////////////////////////////////////////////////////////////////////////
         // Generate JSON cart and hints objects for the javascript returned below.
         //////////////////////////////////////////////////////////////////////////
         $cartData = array(
-            'authcapture' => $authCapture,
             'orderToken' => ($orderCreationResponse) ? $orderCreationResponse->token: '',
         );
 
+        if($checkoutType !== self::CHECKOUT_TYPE_ADMIN) {
```
mccottry: @johnmccombs1, @matt-thomason 

If we are moving auto-capture setting to the Bolt Merchant Dashboard, then this should not be included in the Bolt order creation data, correct?

Also, you can remove the configuration from the Magento admin in configuration.

matt-thomason: Agreed here @johnmccombs1

PR #483, opened 2019-08-09
Frontend non-multistore URL pre-processing

mccottry:
# Description
In the multi-store context and in the caching context, Magento uses a placeholder "___SID" to store a session id on URLs.  This is post-processed upon rendering of the HTML and in single store context, or when it is not needed, it will be removed.  Because we are checking our required URL outside of the rendered context, we get differences between the final outputed URL and the pre-processed URL.  For our purposes, we want to process the URL prior to comparisons.

Fixes: https://app.asana.com/0/544708310157130/1135096267113402

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/app/code/community/Bolt/Boltpay/Helper/Data.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Data.php
@@ -177,8 +177,8 @@ class Bolt_Boltpay_Helper_Data extends Mage_Core_Helper_Abstract
      */
     public function buildOnCloseCallback($closeCustom, $checkoutType)
     {
-        // For frontend URLs, we want to session id process the URL to get the correct URL
-        // in the non-multi-store context
+        // For frontend URLs, we want to "session id process" the URL to get the 
+        // final format URL which may or may not contain the __SID=(S|U) parameter
         $successUrl = Mage::getModel('core/url')->sessionUrlVar(
             $this->getMagentoUrl(Mage::getStoreConfig('payment/boltpay/successpage'))
         );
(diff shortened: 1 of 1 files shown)


app/code/community/Bolt/Boltpay/Helper/Data.php
```diff
@@ -177,7 +177,11 @@ public function buildOnSuccessCallback($successCustom, $checkoutType)
      */
     public function buildOnCloseCallback($closeCustom, $checkoutType)
     {
-        $successUrl = $this->getMagentoUrl(Mage::getStoreConfig('payment/boltpay/successpage'));
+        // For frontend URLs, we want to session id process the URL to get the correct URL
```
johnmccombs1: Can you clarify this comment a little? I know what you're doing but the comment doesn't make sense. Good work.

mccottry: @johnmccombs1 
Would it make more sense with quotations like this:
> For frontend URLs, we want to _"session id process"_ the URL to get the final format URL which may or may not contain the `__SID=SU`

mccottry: Also, if I edit the comment, it will have to re-reviewed.  So, the comment will come in a separate PR

matt-thomason: @mccottry I think it makes more sense to just edit the comment in this PR. It results in the same number of reviews, but 1 less PR and less confusion

PR #497, opened 2019-08-28
Fixes error of Cart has expired after failed_payment hook

mccottry:
# Description
This replaces PR #495 

Quotes are not being blocked for processing after failed_payment hooks.  After a create preauth order is deleted due to failed_payment, (after 10 minutes or immediately if that option is enabled), the quote would be reported as expired due to the order having been created. The manifestation of this in orphaned transactions is a failed webhook with the same error message.

This unblocks the quotes to allow orders to be re-created after failed_payments. 

Fixes: https://app.asana.com/0/inbox/544708310157128/1136200477398193/1136201003916971

# Type of change

# How Has This Been Tested?

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Order.php
+++ b/app/code/community/Bolt/Boltpay/Model/Order.php
@@ -60,19 +60,10 @@ class Bolt_Boltpay_Model_Order extends Bolt_Boltpay_Model_Abstract
 
             $immutableQuoteId = $this->boltHelper()->getImmutableQuoteIdFromTransaction($transaction);
             $immutableQuote = $this->getQuoteById($immutableQuoteId);
-            $immutableQuote->setParentId($transaction->order->cart->order_reference);
+            $immutableQuote->setParentQuoteId($transaction->order->cart->order_reference);
             Mage::app()->setCurrentStore($immutableQuote->getStore());
             $parentQuote = $this->getQuoteById($transaction->order->cart->order_reference);
             benchmark( "Looked up quotes." );
-
-            if (!$parentQuote->getIsActive()) {
-                throw new Exception(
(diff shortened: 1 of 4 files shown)


app/code/community/Bolt/Boltpay/Model/Order.php
```diff
@@ -927,13 +921,30 @@ protected function getRatesDebuggingData($rates) {
...
+            ##########################################################
+            $parentQuote = $this->getParentQuoteFromOrder($order);
+            $parentQuote->setIsActive(true)->save();
+            ###########################################################
+
+            ##############################################################
+            # Cancel order for restocking inventory and triggering events
+            ##############################################################
             if ($order->getStatus() !== 'canceled_bolt') {
                 $order->cancel()->setQuoteId(null)->setStatus('canceled_bolt')->save();
```
johnmccombs1: @mccottry should we move the unblock quote logic to after the order is deleted from the system, otherwise, it might cause some unexpected issues (e.g. sending abandon cart email to the customer, etc).

mccottry: @johnmccombs1 I'm a little hesitant to change the order because in testing, I observed order was important -- though logically, I can not see why this would be the case.  I'll run a few test with the modified order that you suggest because the previous order should be gone before the new order is created, however, I also do not see why canceling or deleting an order would trigger an abandoned cart email.  Technically, those orders are fulfilled.  Have you observed this problem?

PR #511, opened 2019-09-12
Prevent accidentally changing a processing order status to a pending bolt authorization status

johnmccombs1:
This prevents the following case:

The administrator creates an order which shows the current order status as Pending Bolt Authorization. They then comment on the order after Bolt has changed the status to something other than Pending Bolt Authorization and it changes the status back to Pending Bolt Authorization when the comment is submitted. This then causes the order to no longer show up in the admin order grid.

Fixes: https://app.asana.com/0/564264490825835/1139660501761068

# Type of change

# How Has This Been Tested?

# Checklist:

- [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Payment.php
+++ b/app/code/community/Bolt/Boltpay/Model/Payment.php
@@ -127,8 +127,13 @@ class Bolt_Boltpay_Model_Payment extends Mage_Payment_Model_Method_Abstract
     {
         $stateObject
             ->setState(Mage_Sales_Model_Order::STATE_PENDING_PAYMENT)
-            ->setStatus('pending_bolt')
             ->setIsNotified(false);
+
+        if ($this->isAdminArea()){
+            $stateObject->setStatus('pending');
+        } else {
+            $stateObject->setStatus('pending_bolt');
+        }
 
         return parent::initialize($paymentAction, $stateObject);
     }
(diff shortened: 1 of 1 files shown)


mccottry:
Hey @johnmccombs1.  I have a clarifying question.  Is this problem happening only for admin created orders?  I believe that this is the case.

johnmccombs1:
> Hey @johnmccombs1. I have a clarifying question. Is this problem happening only for admin created orders? I believe that this is the case.

@mccottry Yes, this only occurs for admin created orders.

mccottry:
@johnmccombs1 and I discussed changing the initial state on admin created orders over a slack call.  This should fix the problem at it's root as admin orders should never have the pending payment state and pending_bolt status.

app/code/community/Bolt/Boltpay/Model/Observer.php
```diff
@@ -291,4 +291,48 @@ public function validateBeforeOrderCommit($observer) {
...
     }
+
+    /**
+     * Sets the order's initial status according to Bolt and annotates it with creation and payment meta data
+     *
+     * @param Varien_Event_Observer $observer the observer object which contains the order
+     *
+     * @throws Mage_Core_Exception
+     */
+    public function setInitialOrderStatusAndDetails(Varien_Event_Observer $observer) {
```
mccottry: I see no need for this.

app/code/community/Bolt/Boltpay/etc/config.xml
```diff
@@ -198,6 +198,17 @@
         </boltpay>
       </updates>
     </layout>
+    <events>
```
mccottry: This is unnecessary and probably likely very incorrect.  Set any initial status in the payment `initialize` method.  I recommend setting the order state and status to "pending"

PR #523, opened 2019-09-30
Restrict bolt by customer group

oldDriver:
# Description
Ability to Restrict Payment Method by Customer Group

Fixes: https://app.asana.com/0/941895179897714/1142266855481737/f

# Type of change

# How Has This Been Tested?

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -457,7 +457,16 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
     {
         return Mage::getStoreConfig('payment/boltpay/skip_payment');
     }
-
+    
+    /**
+     * Return disabled customer groups for Bolt
+     * @return array
+     */
+    public function disableCustomerGroups()
+    {
+        return explode(',', Mage::getStoreConfig('payment/boltpay/bolt_disabled_customer_groups'));
+    }
+    
     /**
      * Returns whether enable merchant scoped account.
      * @return string
@@ -613,6 +622,13 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
      */
     private function isAllowedOnCurrentPageByRoute()
     {
+        $isAllowed = false;
+        $quote = $this->getQuote();
+        $customerGroupId = $quote->getCustomerGroupId();
+        $disabledGroups = $this->disableCustomerGroups();
+        if ($customerGroupId && in_array($customerGroupId, $disabledGroups)) {
+            return $isAllowed;
+        }
         $routeName = $this->getRequest()->getRouteName();
         $controllerName = $this->getRequest()->getControllerName();
 
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -459,14 +459,26 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
     }
     
     /**
-     * Returns disabled customer group Ids for the Bolt plugin
+     * Returns array of disabled customer group Id integers for the Bolt plugin
      * @return array
      */
     public function getDisabledCustomerGroups()
     {
         return explode(',', Mage::getStoreConfig('payment/boltpay/bolt_disabled_customer_groups'));
     }
-    
+
+    /**
+     * Check if customer group ID in the disabled groups fo the Bolt Plugin
+     *
+     * @param integer $customerGroupId
+     * @return boolean
+     */
+    private function isCustomerGroupDisabled($customerGroupId)
+    {
+        $disabledCustomerGroups = $this->getDisabledCustomerGroups();
+        return in_array($customerGroupId, $disabledCustomerGroups);
+    }
+
     /**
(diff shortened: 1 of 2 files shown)


app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -457,7 +457,16 @@ function isBoltOnlyPayment()
     {
         return Mage::getStoreConfig('payment/boltpay/skip_payment');
     }
-
+    
+    /**
+     * Return disabled customer groups for Bolt
+     * @return array
+     */
+    public function disableCustomerGroups()
```
mccottry: When creating a "getter" method, by convention in M1, the method names start with `get`.  Therefore this 
 should be named `getDisabledCustomerGroups`.

However, considering what you are trying to do and the comment below about corrupting/overriding a well defined method, it is better that you create a `private` method here and call it `isCustomerGroupAllowed` and returning a boolean true|false.  This method should then be called and `&&` to this condition:

**https://github.com/BoltApp/bolt-magento1/blob/master/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php#L664** in the `isAllowedConnectJsOnCurrentPage` method of this same class.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -457,7 +457,16 @@ function isBoltOnlyPayment()
     {
         return Mage::getStoreConfig('payment/boltpay/skip_payment');
     }
-
+    
+    /**
+     * Return disabled customer groups for Bolt
+     * @return array
```
mccottry: We can do a better job documenting that we are returning an array of customer group Id integers.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -613,6 +622,13 @@ public function canUseBolt()
      */
     private function isAllowedOnCurrentPageByRoute()
     {
+        $isAllowed = false;
```
mccottry: This is the wrong place for this code.  You are basically corrupting and overriding a well defined method with something completely unrelated.  This code should be removed from here and added to a private method as described above.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
oldDriver: @mccottry add new condition in the method isAllowedConnectJsOnCurrentPage - bad practice.
Now in this method we call independently two methods. First method could return true and really we don't need second method. This is on one hand. Just for performance it's better create chain of rules. If first rule returns false (Bolt is not allowed) we don't need do any other calculations. Because of this i added method isCustomerGroupDisabled in the method isAllowedOnCurrentPageByRoute
On the other hand this is not a problem for boolean logic get result based on two parameters, but much more complex get result based on three parameters.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -453,16 +453,37 @@ function getConfigSelectors()
...
+    public function isBoltOnlyPayment()
     {
         return Mage::getStoreConfig('payment/boltpay/skip_payment');
     }
+    
+    /**
+     * Returns array of disabled customer group Id integers for the Bolt plugin
+     * @return array
+     */
+    public function getDisabledCustomerGroups()
```
roopakv: why make this function if it only accessed in this class?

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -453,16 +453,37 @@ function getConfigSelectors()
      * Returns the Skip Payment Method Step configuration.
      * @return string
      */
-    function isBoltOnlyPayment()
+    public function isBoltOnlyPayment()
```
roopakv: why public? is it accessed by other classes?

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -453,16 +453,37 @@ function getConfigSelectors()
...
+        $disabledCustomerGroups = $this->getDisabledCustomerGroups();
+        return in_array($customerGroupId, $disabledCustomerGroups);
+    }
 
     /**
      * Returns whether enable merchant scoped account.
      * @return string
      */
-    function isEnableMerchantScopedAccount()
+    public function isEnableMerchantScopedAccount()
```
roopakv: leave as private unless really needed to expose?

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: @roopakv We definitely should be looking to use much more strict encapsulation in our code.  

Remember, unlike in Java when a method is declared without an access modifier, it does not go into `package` scope or `private` scope like in maybe some other languages, but in PHP, they are `public` by default.  So, the code above is just explicitly declaring what is implicitly invoked.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
oldDriver: @roopakv in the PHP by default functions are public. As @mccottry mentioned,  the code above is just explicitly declaring what is implicitly enforced.

PR #500, opened 2019-08-28
Replace curl with guzzle

johnmccombs1:
# Description
Replace curl with guzzle

Fixes: https://app.asana.com/0/564264490825835/1137562463503385

# Type of change

- [ ] Bug fix (change which fixes an issue)
- [x] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?
Please validate that you have tested your change in at least one of the following areas:

- [x] Successfully tested locally (or docker image)
- [ ] Successfully tested on a staging or sandbox server
- [ ] Successfully tested on a merchant's staging server

# Checklist:

- [x] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have created or modified unit tests to sufficiently cover my changes.

PR #536, opened 2019-10-29
Fixed case, when all groups are allowed, that means that disable gropus

oldDriver:
array with empty value, but not empty arrray

# Description
if user unlogged and all groups are allowed bolt is disabled. This is not right

Fixes: https://app.asana.com/0/941895179897714/1142266855481737/f

\#changelog

# Type of change

- [+] Bug fix (change which fixes an issue)

# How Has This Been Tested?

- [+] Successfully tested locally (or docker image)
- [+] Successfully tested on a staging or sandbox server
- [+] Successfully tested on a merchant's staging server

# Checklist:

### DIFF
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -466,7 +466,12 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
      */
     private function isCustomerGroupDisabled($customerGroupId)
     {
-        $disabledCustomerGroups = explode(',', Mage::getStoreConfig('payment/boltpay/bolt_disabled_customer_groups'));
+        $disabledCustomerGroups = Mage::getStoreConfig('payment/boltpay/bolt_disabled_customer_groups');
+        // Case ot empty groups
+        if (empty($disabledCustomerGroups)) {
+            return false;
+        }
+        $disabledCustomerGroups = explode(',', $disabledCustomerGroups);
         return in_array($customerGroupId, $disabledCustomerGroups);
     }
 
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -467,7 +467,7 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
     private function isCustomerGroupDisabled($customerGroupId)
     {
         $disabledCustomerGroups = Mage::getStoreConfig('payment/boltpay/bolt_disabled_customer_groups');
-        // Case ot empty groups
+        // Case of empty groups
         if (empty($disabledCustomerGroups)) {
             return false;
         }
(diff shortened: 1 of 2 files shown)


oldDriver:
@daisy1754 sure, sir!

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
```diff
@@ -245,4 +254,368 @@ public function testGetSaveOrderURL()
 
         $this->assertEquals($expected, $result);
     }
-}
\ No newline at end of file
+
+    /**
+     * @test
```
mccottry: I am definitely not a fan of breaking from universally established long-running naming conventions for test and using annotations, but, it is good to use all the resources available and perhaps we can change to this proposed convention.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -466,7 +466,12 @@ public function isBoltOnlyPayment()
      */
     private function isCustomerGroupDisabled($customerGroupId)
     {
-        $disabledCustomerGroups = explode(',', Mage::getStoreConfig('payment/boltpay/bolt_disabled_customer_groups'));
+        $disabledCustomerGroups = Mage::getStoreConfig('payment/boltpay/bolt_disabled_customer_groups');
+        // Case ot empty groups
```
mccottry: // Case **of** empty groups

Also, it is better to use this comment to explain why you must return false here, which is really because you are preventing the case of the user not belonging to a customer group and no customer groups being disabled.

Possibly, easier to understand code would be:
```
     // False if the customer is not part of any groups or no groups are disabled
     if (empty($customerGroupId) || empty($disabledCustomerGroups) {
          return false;
     }
     $disabledCustomerGroups = explode(',', $disabledCustomerGroups);
     return in_array($customerGroupId, $disabledCustomerGroups);
```

oldDriver: @mccottry PHP has magic!
explode(',', '') returns array('') and this array is not empty!
in_array(0, array('')) returns true
Type will be fixed.

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
oldDriver: @mccottry annotations allow use cases in one test.
Also, it is clear what method is covered by this test.
I think five tests named something like
"testIsCustomerGroupDisabledWithEmptyCustomerAndEmptyGroups" is not so clear as simple case
Using "@group" allow to run not all tests but some range.

mccottry: The idea is following existing convention.  If all test methods start with test, and then you decide instead to use a different type of name like `getConfigSelectors` as a new test and use instead the annotation @test, then you have introduced unnecessary disharmony.  A casual reader might interpret `getConfigSelectors` as a local getter method and not a test.  In general, honoring existing convention is encouraged in software engineering in team environments.

However, as I mentioned, we should be open to adopting new and better conventions when easy to adopt.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: correct.  In the code listed above, as you have written the method, the `empty($disabledCustomerGroups)` is not testing an array, but rather a string.  If you review the code above, you will see that you will never be allowed the opportunity to compare `in_array(0, array(''))`.

Thank you for updating the spelling.

mccottry:
@oldDriver please try checking the boxes after you create your pull request.  Observer the difference in #500

PR #539, opened 2019-10-31
Adds a defensive pre-auth order cron cleanup in case other mechanisms fail

mccottry:
# Description
Initially proposed as PR #385

After a pre-auth pending order has existed for 15 minutes or more, we remove it from the system.
At this point, Bolt should have called the "failed_payment" webhook to do this. We are catching
the anomaly cases where this mechanism fails.  This will resolve any outstanding pending_payment Bolt orders on production servers.

Fixes: https://app.asana.com/0/1125081140214268/1146968193990481/f

#changelog  Internal order artifact cleanup routines

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Cron.php
+++ b/app/code/community/Bolt/Boltpay/Model/Cron.php
@@ -24,6 +24,12 @@ class Bolt_Boltpay_Model_Cron
 {
     use Bolt_Boltpay_BoltGlobalTrait;
 
+    /*
+     * We will have a conservative tolerance of 6 hours for non-confirmed pre-auth orders before we allow the system
+     * to remove them
+     */
+    const PRE_AUTH_STATE_TIME_LIMIT_MINUTES = 600;
+
     /**
      * After an immutable quote / unused generated session quote on PDP has existed for 2 weeks or more, we remove it from the system.
      * At this point, only the order object is relevant for converted orders and any immutable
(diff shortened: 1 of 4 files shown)


app/code/community/Bolt/Boltpay/etc/config.xml
```diff
@@ -295,6 +295,14 @@
           <model>boltpay/cron::cleanupQuotes</model>
         </run>
       </immutable_quote_cleanup>
+      <preauth_order_cleanup>
+        <schedule>
+          <cron_expr>*/15 * * * *</cron_expr>
```
ryan-timothy-albert: Does this mean we are running every 15 minutes. If not what is the time frame and what is the reason we chose this interval.

nirajjayantbolt: @mccottry I think 15mins might be too aggressive too. This is really for AKIRA where we want to stop Failed Webhooks and has a shitty server. Maybe we can do every hour?

app/code/community/Bolt/Boltpay/Model/Order.php
```diff
@@ -966,6 +966,8 @@ public function removePreAuthOrder($order) {
             if ($this->boltHelper()->getExtraConfig('keepPreAuthOrders')) {return;}
             $previousStoreId = Mage::app()->getStore()->getId();
             Mage::app()->setCurrentStore(Mage_Core_Model_App::ADMIN_STORE_ID);
+            $adminEntry = Mage::getModel('sales/order_grid')->load($order->getId());
```
nirajjayantbolt: can you explain what this part is?

app/code/community/Bolt/Boltpay/etc/config.xml
nirajjayantbolt: Oh I see I see, okay lets maybe keep it at 20mins to give us a little bit of a buffer

app/code/community/Bolt/Boltpay/Model/Cron.php
```diff
@@ -44,4 +44,44 @@ public function cleanupQuotes() {
 
         $connection->query($sql);
     }
+
+    /**
+     * After a pre-auth pending order has existed for 15 minutes or more, we remove it from the system.
+     * At this point, Bolt should have called the "failed_payment" webhook to do this. We are catching
```
daisy1754: let's explicitly mention that this is accounting for case like magento returns 500 but Bolt proceed to auth?

app/code/community/Bolt/Boltpay/Model/Cron.php
```diff
@@ -44,4 +44,44 @@ public function cleanupQuotes() {
...
+     * the anomaly cases where this mechanism fails.
+     */
+    public function cleanupOrders() {
+        $expiration_time = Mage::getSingleton('core/date')->gmtDate('Y-m-d H:i:s', time()-(60*15));  // Magento uses GMT to save in DB
```
daisy1754: let's move this 15 to constant. Also I suggest we move this to somthing more generous like 6 hours.  we can still run cron every 15min but let's start with less aggressive with deleting data

app/code/community/Bolt/Boltpay/Model/Cron.php
```diff
@@ -44,4 +44,44 @@ public function cleanupQuotes() {
...
+
+        $ordersToRemove = $expiredPendingPaymentOrderCollection->getItems();
+
+        /** @var Bolt_Boltpay_Model_Order $orderModel */
+        $orderModel = Mage::getModel('boltpay/order');
+
+        /** @var Mage_Sales_Model_Order $order */
+        foreach($ordersToRemove as $order) {
+            try {
+                $orderModel->removePreAuthOrder($order);
```
daisy1754: do we need separate treatment for this case? in failed_payment case we want to reactivate quote but in 500 case it's not (?)

app/code/community/Bolt/Boltpay/Model/Cron.php
```diff
@@ -44,4 +44,44 @@ public function cleanupQuotes() {
 
         $connection->query($sql);
     }
+
+    /**
+     * After a pre-auth pending order has existed for 15 minutes or more, we remove it from the system.
+     * At this point, Bolt should have called the "failed_payment" webhook to do this. We are catching
+     * the anomaly cases where this mechanism fails.
+     */
+    public function cleanupOrders() {
```
daisy1754: should we have top-level try-catch just as a safe mechanism? like what happens if this some logic fails? we have no access to server-side log

app/code/community/Bolt/Boltpay/etc/config.xml
ryan-timothy-albert: bump

nirajjayantbolt:
@mccottry can you respond to Kazukis comments and ping us when ready?

app/code/community/Bolt/Boltpay/etc/config.xml
mccottry: 20 minutes it is

app/code/community/Bolt/Boltpay/Model/Order.php
mccottry: @nirajjayantbolt 
In Magento, deleting orders is not enough to prevent them from being displayed in the admin.   There is a separate table for this exclusively for presentation of the orders.  In some databases, constraints will keep this and the orders table in sync.  In other cases, the constraints are not enabled and this is needed to remove the order from being listed in the admin.  This code is for that case.

app/code/community/Bolt/Boltpay/Model/Cron.php
mccottry: sounds good.

app/code/community/Bolt/Boltpay/Model/Cron.php
mccottry: @daisy1754 I don't understand your reasoning here.  I don't see a separate case.  There is only one case here: *if an order is still in the pending_bolt state after an unreasonable amount of time.*  In all cases, these orders should be handled regardless of under which conditions the order was created ( 500, timeout, abnormal response.)  How do you see this differently?

app/code/community/Bolt/Boltpay/Model/Cron.php
mccottry: There is an individual try/catch per remove order attempt, but let's add a catch-all just in case for abnormal failures.

app/code/community/Bolt/Boltpay/Model/Order.php
nirajjayantbolt: Oh okay got it thanks!

app/code/community/Bolt/Boltpay/Model/Cron.php
mccottry: done

app/code/community/Bolt/Boltpay/Model/Cron.php
daisy1754: tbh I didn't too much thought (it's more of question than suggestion). I think what you say makes sense

app/code/community/Bolt/Boltpay/Model/Cron.php
mccottry: @daisy1754 I thought about this a little more.  I accidentally set this at 10 hours instead of 6, and when going to change it, I believe that 6 hours is way too long.  The original 15 minute idea comes from the fact that we are holding up inventory... and in this case, holding inventory on an order that is likely never to succeed.  I believe a much more reasonable time would be 30 minutes.

What do you think?

nirajjayantbolt: I agree with Leon here - lets start with 30 minutes and we can monitor to tweak it?

PR #556, opened 2019-11-27
Improves the cleanup of artifacts created from by test

mccottry:
# Description
Many test have failed locally due to improper cleanup.  This addresses the issue.

Fixes: https://app.asana.com/0/544708310157130/1151537717819185

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/tests/unit/testsuite/Bolt/Boltpay/OrderHelper.php
+++ b/tests/unit/testsuite/Bolt/Boltpay/OrderHelper.php
@@ -128,9 +128,10 @@ class Bolt_Boltpay_OrderHelper
         $resource = Mage::getSingleton('core/resource');
         /** @var Magento_Db_Adapter_Pdo_Mysql $writeConnection */
         $writeConnection = $resource->getConnection('core_write');
-        $table = $resource->getTableName('sales/order');
+        $orderTable = $resource->getTableName('sales/order');
+        $orderGridTable = $resource->getTableName('sales/order_grid');
 
-        $query = "DELETE FROM $table WHERE increment_id = :increment_id";
+        $query = "DELETE ot, ogt FROM $orderTable ot LEFT JOIN $orderGridTable ogt ON ot.increment_id = ogt.increment_id WHERE ot.increment_id = :increment_id";
         $bind = array(
             'increment_id' => (int)$incrementId
         );
--- a/tests/unit/testsuite/Bolt/Boltpay/ProductProvider.php
+++ b/tests/unit/testsuite/Bolt/Boltpay/ProductProvider.php
@@ -103,9 +103,10 @@ class Bolt_Boltpay_ProductProvider
         $resource = Mage::getSingleton('core/resource');
         /** @var Magento_Db_Adapter_Pdo_Mysql $writeConnection */
         $writeConnection = $resource->getConnection('core_write');
-        $table = $resource->getTableName('catalog/product');
+        $productTable = $resource->getTableName('catalog/product');
+        $inventoryTable = $resource->getTableName('cataloginventory/stock_item');
 
-        $query = "DELETE FROM ".$table." WHERE entity_id = :productId";
+        $query = "DELETE pt, it FROM ".$productTable." pt LEFT JOIN $inventoryTable it ON it.product_id = pt.entity_id WHERE pt.entity_id = :productId";
         $bind = array(
             'productId' => (int) $productId
         );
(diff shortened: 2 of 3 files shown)


tests/unit/testsuite/Bolt/Boltpay/ProductProvider.php
```diff
@@ -103,9 +103,10 @@ public static function  deleteDummyProduct($productId)
         $resource = Mage::getSingleton('core/resource');
```
oldDriver: @Leon, is it better to use Magento style?
$product = Mage::getModel('catalog/product')->load($productId);
        Mage::register('isSecureArea', true);
        $product->delete();
        Mage::unregister('isSecureArea');
And product will be deleted.

tests/unit/testsuite/Bolt/Boltpay/OrderHelper.php
```diff
@@ -128,9 +128,10 @@ public static function deleteDummyOrderByIncrementId($incrementId)
         $resource = Mage::getSingleton('core/resource');
```
oldDriver: @mccottry the same with order, you could use 
Mage::register('isSecureArea', true);
$order->delete();
Mage::unregister('isSecureArea');

tests/unit/testsuite/Bolt/Boltpay/ProductProvider.php
mccottry: @oldDriver
That is a very good question and we can use this opportunity to discuss using **_Magento ORM_ (model)** constructs to create and delete objects versus using an approach that _directly queries the database_.  There are to execution flows that we should be aware of for this discussion which have different ideal approaches.  These two execution flows are:

1. **Production** - from within a working store which may have several custom modules
2. **Unit Testing** - an execution of code ideally designed to be focused and isolated

__Production Execution__

In production execution, it is extremely important that when we create, delete, or modify orders or entities, that we notify all interested extensions of our actions so that they may execute their code in response.  When we use _Magento models_ and functions like `delete`, what we are effectively doing is

1. checking external conditions to verify if the requested action is allowed
2. creating a DB query that deletes that specific models underlying table entry
3. relying on DB defined constraints to handle related data entries
4. trigger the execution of event handlers for registered observers like 3rd party modules

So, especially for the reason of triggering observed event handlers and also checking if the operation should be allowed, it is ideal in production execution to use Magento models -- what you refer to as "Magento style"

In this execution flow, if you were to use the Magento provided interfaces for _direct queries_, (which as a side-note, is also "Magento style"), we would certainly fail to trigger the listening 3rd party software.  This is no good!

So, when in the production execution flow, we do use the ORM as much as possible.

__Unit Testing__

In the unit test context, things are different. Here we are definitely not interested in triggering unexpected events.  We want to do a very specific and defined thing.  In other words, ideally, we want to do just enough for our test to be effective and nothing more.

In this context, using the direct query approach is similar to using stubbed mocks where we only expose and touch what we need.  For example, if one only needs an order in the database that can be looked up with the ID of 987654321 with the specific state of "pending_payment", then ideally, we just create that entry with some sql and do trigger side events like execute reward points, trigger conversion analytics listeners, deal with inventory, etc.  This is similar to the opposite direction.  If I created a dummy entry in the DB as described above, and I now wish to delete it, I ideally only have to clear that single entry in a single table with one concise query rather then the ORM doing things like triggering refunds in ERPs, re-indexing the DB, or whatever other unexpected things maybe listening.

To take it a step further, imagine if we wanted to create and delete 1000s of dummy entries.  Having the weight of an ORM, especially with Magento's EAV open schema model, becomes prohibitive.

So, in the context of unit testing code that is coupled with the DB, ideally we want to use concise DB queries and avoid using the ORM whenever it is practical.

__The Point of the Pull Request__

So, to answer the original question, __using the ORM is preferred in real world production context, while using direct concise queries is preferred in the unit testing context.__

**Now, here is the kicker!:** The point of this pull request has nothing to do with using an ORM versus using direct queries.  We have already pointed out that the by using the ORM for something like:
```
$product->delete();
```
will generate a database query like this
```
DELETE FROM `catalog_product_entity` WHERE `entity_id` = 987654321
```
We also mentioned that we are still now relying on the DB defined constraints to handle removal and cleanup of all other related entries.  **The point is after doing a bunch of test and development manipulating the database in so many ways, the DB constraints are no longer working!**  As such, the ORM approach does nothing to solve the problem that the PR was created to solve.

I've investigated and seen that many others have ran into this problem with no explanation of how this happens, and that there is even a tool for repairing this:

https://docs.magento.com/m1/ce/user_guide/magento/database-repair-tool.html

To this point, I have manually run queries to get the DB back to the proper state and I do not find constantly running this tool as a practical solution for a development environment.  So instead, I am proposing being more concise in our cleanup queries.  With the proposed changes, I am running and re-running test locally with no issues.

PR #558, opened 2019-11-27
Fixes 'hidden' failures in Bolt_Boltpay_Model_CronTest

mccottry:
# Description
Due to an oversight with output buffering affecting the reporting of unit test failures, several failed test have been going unnoticed under the radar.  These changes fix the test and/or tested object to restore M1 Unit test is good standing.

Fixes: https://app.asana.com/0/544708310157130/1151537717819187

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Cron.php
+++ b/app/code/community/Bolt/Boltpay/Model/Cron.php
@@ -66,17 +66,17 @@ class Bolt_Boltpay_Model_Cron
      */
     public function cleanupOrders() {
         try {
-            $expiration_time = Mage::getSingleton('core/date')
-                ->gmtDate('Y-m-d H:i:s', time()-(60*PRE_AUTH_STATE_TIME_LIMIT_MINUTES));  // Magento uses GMT to save in DB
+            $expiration_time = gmdate('Y-m-d H:i:s', time()-(60*self::PRE_AUTH_STATE_TIME_LIMIT_MINUTES));  // Magento uses GMT to save in DB
 
             /* @var Mage_Sales_Model_Resource_Order_Collection $orderCollection */
             $orderCollection = Mage::getModel('sales/order')->getCollection();
 
+             $orderCollection
+                ->addFieldToFilter('created_at', array( 'gteq' => $expiration_time))
+                ->setOrder('created_at', 'ASC');
+
             /** @var Mage_Sales_Model_Order $deletePendingPaymentOrdersBeforeThis */
-            $deletePendingPaymentOrdersBeforeThis = $orderCollection
-                ->addFieldToFilter('created_at', array( 'gte' => $expiration_time))
-                ->setOrder('created_at', 'ASC')
-                ->getFirstItem();
+            $deletePendingPaymentOrdersBeforeThis = $orderCollection->getFirstItem();
 
             /* @var Mage_Sales_Model_Resource_Order_Collection $expiredPendindOrderCollection */
             $expiredPendingPaymentOrderCollection = Mage::getModel('sales/order')->getCollection();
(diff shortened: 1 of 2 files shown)


app/code/community/Bolt/Boltpay/Model/Cron.php
```diff
@@ -66,17 +66,17 @@ public function cleanupQuotes() {
      */
     public function cleanupOrders() {
         try {
-            $expiration_time = Mage::getSingleton('core/date')
-                ->gmtDate('Y-m-d H:i:s', time()-(60*PRE_AUTH_STATE_TIME_LIMIT_MINUTES));  // Magento uses GMT to save in DB
+            $expiration_time = gmdate('Y-m-d H:i:s', time()-(60*self::PRE_AUTH_STATE_TIME_LIMIT_MINUTES));  // Magento uses GMT to save in DB
```
adentong: Just wondering, why did you get rid of the getSIngleton?

app/code/community/Bolt/Boltpay/Model/Cron.php
```diff
@@ -66,17 +66,17 @@ public function cleanupQuotes() {
...
+             $orderCollection
+                ->addFieldToFilter('created_at', array( 'gteq' => $expiration_time))
+                ->setOrder('created_at', 'ASC');
+
             /** @var Mage_Sales_Model_Order $deletePendingPaymentOrdersBeforeThis */
-            $deletePendingPaymentOrdersBeforeThis = $orderCollection
-                ->addFieldToFilter('created_at', array( 'gte' => $expiration_time))
-                ->setOrder('created_at', 'ASC')
-                ->getFirstItem();
+            $deletePendingPaymentOrdersBeforeThis = $orderCollection->getFirstItem();
```
adentong: Is there a reason why you split this instead of adding a single "q"?

tests/unit/testsuite/Bolt/Boltpay/Model/CronTest.php
```diff
@@ -4,6 +4,10 @@
 
 class Bolt_Boltpay_Model_CronTest extends PHPUnit_Framework_TestCase
 {
+    /**
+     * A time in minutes padded to test entry timestamps to make them eligible for cleanup
+     */
+    const MINUTES_PADDING_FOR_TEST_ENTRY_TIMESTAMPS = 3;
```
matt-thomason: clever

app/code/community/Bolt/Boltpay/Model/Cron.php
mccottry: @adentong 
The mode was doing some locale based conversion on the dates that was not necessary as the time() function is returning things in GMT already.  Using the model function is an error in this context.

https://github.com/OpenMage/magento-mirror/blob/magento-1.9/app/code/core/Mage/Core/Model/Date.php#L137-L164

https://www.php.net/manual/en/function.time.php

adentong: Interesting!

app/code/community/Bolt/Boltpay/Model/Cron.php
mccottry: @adentong 
There a few factors into this.

Firstly, in M1, by default, Bolt orders that have not been paid for are created without a timestamp.  This serves to hide the order from ERPs that import orders base on the `created_at` timestamp.  Because of this, we use the timestamps of converted orders to find the lowest id from which it is safe to delete Bolt pending payment orders.

Second, we are interested in order model objects.  When we delete, we want to use the ORM so that we trigger all the  observers of the deletion event so that they execute their code and the removal of the orders is cleaned up the right way for all 3rd party plugins.

This brings me to the third factor.  If I were simply writing a plain SQL query to get the ids of the orders, then it would be a "not-even-thinking-twice-about-it" writing a single SQL statement.  However, since I am using their ORM and query builder, making the joins using it seems much more complex and harder to understand for future developers than to breaking things down into two very simple and understandable collection object queries.  Moreover, I am trusting that their implementation of this is more efficient than me first querying for the order IDs and then loading the order objects in a loop one-by-one in order to delete them.

So, for the the end goal, I believe like this is simpler to understand and easier to implement and easier to maintain.

PR #565, opened 2019-11-28
Change Discount from 2001003 to 2001006

mccottry:
# Description
A change in the spec has been made that coupon discounts use code 2001006 for more informative displays to customers in case of errors

Fixes: https://app.asana.com/0/544708310157130/1147324810049189

# Type of change

- [ ] Test

# How Has This Been Tested?

# Checklist:

- [x] New and existing unit tests pass locally with my changes.

### DIFF
--- a/tests/unit/testsuite/Bolt/Boltpay/Model/OrderTest.php
+++ b/tests/unit/testsuite/Bolt/Boltpay/Model/OrderTest.php
@@ -6,7 +6,7 @@ require_once('MockingTrait.php');
 use Bolt_Boltpay_TestHelper as TestHelper;
 
 /**
- * Class Bolt_Boltpay_Model_OrderTest
+ * @coversDefaultClass Bolt_Boltpay_Model_Order
  */
 class Bolt_Boltpay_Model_OrderTest extends PHPUnit_Framework_TestCase
 {
@@ -18,7 +18,7 @@ class Bolt_Boltpay_Model_OrderTest extends PHPUnit_Framework_TestCase
     protected $testClassName = 'Bolt_Boltpay_Model_Order';
 
     /**
-     * @var Bolt_Boltpay_Model_Order  The mocked instance the test class
+     * @var Bolt_Boltpay_Model_Order|PHPUnit_Framework_MockObject_MockObject  The mocked instance the test class
      */
     private $testClassMock;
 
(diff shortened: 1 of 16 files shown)


tests/unit/testsuite/Bolt/Boltpay/Model/OrderTest.php
```diff
@@ -157,4 +157,197 @@ public function testActivateOrder()
...
+
+        $mocks = $this->validateTotalsSetup(20, .75, .50, 1);
+
+        $mockBoltHelper = $mocks['mockBoltHelper'];
+        $mockImmutableQuote = $mocks['mockImmutableQuote'];
+        $mockTransaction = $mocks['mockTransaction'];
+
+        $mockBoltHelper->expects($this->never())->method('logWarning');
+
+        try {
```
matt-thomason: @mccottry I'd much rather be using the `expectException` workflow - can we switch over to that workflow?
https://thephp.cc/news/2016/02/questioning-phpunit-best-practices

mccottry: @matt-thomason I am definitely aware of the `expectException` method and workflow and considered that first.  Unfortunately, that method was introduced in PHPUnit 5.2 and half of our created benchmarks use PHPUnit 4.8.36, so that method is not an option.

The next "modern" approach would be to use the annotation `@expectedException` and its family.  My problem with this approach is that it does not give control over the timing in which the exception is thrown.  While this may be a minor concern in this situation, it definitely gave me pause to think.  If not using the older return and fail, this is the route that I was thinking of going.

The third consideration was the older, `setExpectedException` method and it's family.  I chose not to use that as they are deprecated in the other PHPUnit suite that we use, PHPUnit 5.7.9.  

So, no matter which way we go here, we are caught between rocks and a hard places.

1.) Uglier code - where we are now (catch, return, fail) - (pro - Exact code and handling)
2.) Code that won't run in 4.8 - `expectException` - (pro - the most elegant)
3.) Unpredictable timing on when watching exceptions - annotations - (pro-they seem cooler)
4.) Deprecated methods in 5.7.9 - `setExpectedException` - (pro - similar to expectException)

Which path do you suggest we take?

tests/unit/testsuite/Bolt/Boltpay/Model/OrderTest.php
```diff
@@ -157,4 +157,195 @@ public function testActivateOrder()
         $this->assertEquals('TEST-BOLT-REFE-TWO2', $payment->getAdditionalInformation('bolt_reference'));
         ////////////////////////////////////////////////////////////
     }
+
+    /**
+     * @test
+     * That when the discount price is the same at Bolt and Magento, no exception is thrown
+     *
+     * @covers ::validateTotals
```
oldDriver: @mccottry according documentation https://phpunit.de/manual/4.8/en/appendixes.annotations.html#appendixes.annotations.covers
@cover ::validateTotals specifies that the annotated test method covers the specified global function.

oldDriver: @mccottry please also add group annotation - this allow in future run not all unit tests.
Hint: each method could belong to several groups, it make sense add such groups:
@group Model - in order to test only models
@group Order - to run tests for Orders
@group ModelOrder - to run tests from one file

oldDriver:
@mccottry  @matt-thomason propose you to change test according best practice.
It will look like
    /**
     * @test
     * That when the discount price is different at Bolt than Magento beyond the price fault tolerance,
     * a general discount mismatch exception is thrown
     * @expectedException Bolt_Boltpay_OrderCreationException
     * @expectedExceptionCode 2001006
     * @expectedExceptionMessage Discount total has changed
     * @group ModelOrder
     * @covers ::validateTotals
     */
    public function validateTotals_discountDiffersBeyondTolerance() {

        $mocks = $this->validateTotalsSetup(20, .75, .50, 1);

        $mockBoltHelper = $mocks['mockBoltHelper'];
        $mockImmutableQuote = $mocks['mockImmutableQuote'];
        $mockTransaction = $mocks['mockTransaction'];
        $mockBoltHelper->expects($this->never())->method('logWarning');
        TestHelper::callNonPublicFunction(
            $this->testClassMock,
            'validateTotals',
            [$mockImmutableQuote, $mockTransaction]
        );
    }
I tested this and it works

tests/unit/testsuite/Bolt/Boltpay/Model/OrderTest.php
mccottry: @oldDriver  use --filter

PR #571, opened 2019-12-03
Bugfix/track js on normal pages

mccottry:
# Description
track.js is was not loaded on pages that where neither the product page nor the shopping cart page when the one page checkout publishable key was not present.  This is addressed as follows

```
    /**
     * Returns the One Page or Multi-Page checkout Publishable key depending on the requested
     * page and which key has been set in the admin.  Cart and Product pages prioritize the multi-step key
     * and falls back to the payment only key.  All other pages prioritize the payment only key and falls back
     * to the multi-step key.
     */
```

Fixes: https://app.asana.com/0/544708310157130/1150319197435017

# Type of change

- [x] Test

# How Has This Been Tested?

# Checklist:

- [x] New and existing unit tests pass locally with my changes.

### DIFF
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -516,6 +516,9 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
      * page and which key has been set in the admin.  Cart and Product pages prioritize the multi-step key
      * and falls back to the payment only key.  All other pages prioritize the payment only key and falls back
      * to the multi-step key.
+     *
+     * @return string   The publishable key to be used on this page
+     * @throws Bolt_Boltpay_BoltException when neither a multi-step nor a payment-only publishable key is configured
      */
     public function getPublishableKeyForThisPage() {
         $routeName = $this->getRequest()->getRouteName();
@@ -534,7 +537,7 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
         if (!$multiStepPublishableKey && !$paymentOnlyPublishableKey) {
             $noPublishableKeyException = new Bolt_Boltpay_BoltException("No publishable key has been configured.");
             $this->boltHelper()->logException($noPublishableKeyException);
-            $this->boltHelper()->notifyException($noPublishableKeyException,[],'error');
+            $this->boltHelper()->notifyException($noPublishableKeyException, [], 'error');
             throw $noPublishableKeyException;
         }
 
--- a/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
+++ b/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
@@ -512,6 +512,36 @@ class Bolt_Boltpay_Block_Checkout_Boltpay extends Mage_Checkout_Block_Onepage_Re
     }
 
     /**
+     * Returns the One Page or Multi-Page checkout Publishable key depending on the requested
+     * page and which key has been set in the admin.  Cart and Product pages prioritize the multi-step key
+     * and falls back to the payment only key.  All other pages prioritize the payment only key and falls back
+     * to the multi-step key.
+     */
+    public function getPublishableKeyForThisPage() {
+        $routeName = $this->getRequest()->getRouteName();
+        $controllerName = $this->getRequest()->getControllerName();
+
+        $multiStepPublishableKey = $this->getPublishableKey('multi-page');
+        $paymentOnlyPublishableKey = $this->getPublishableKey('one-page');
+
+        $thisPagesPublishableKey =
+            $controllerName === 'cart'
+            || ($routeName === 'catalog' && $controllerName === 'product')
+                ? $multiStepPublishableKey
+                : $paymentOnlyPublishableKey
+        ;
+
+        if (!$multiStepPublishableKey && !$paymentOnlyPublishableKey) {
+            $noPublishableKeyException = new Bolt_Boltpay_BoltException("No publishable key has been configured.");
+            $this->boltHelper()->logException($noPublishableKeyException);
+            $this->boltHelper()->notifyException($noPublishableKeyException,[],'error');
+            throw $noPublishableKeyException;
+        }
+
+        return $thisPagesPublishableKey ?: max($multiStepPublishableKey, $paymentOnlyPublishableKey);
+    }
+
+    /**
      * Returns the One Page / Multi-Page checkout Publishable key.
      *
      * @param  string $checkoutType  'multi-page' | 'one-page' | 'admin'
--- a/tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
+++ b/tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
@@ -45,7 +45,7 @@ class Bolt_Boltpay_Block_Checkout_BoltpayTest extends PHPUnit_Framework_TestCase
         Mage::unregister('_helper/boltpay');
         $this->helperMock = $this->getMockBuilder('Bolt_Boltpay_Helper_Data')
             ->setMethods(
-                ['notifyException', 'logException']
+                array('notifyException', 'logException')
             )
             ->getMock();
 
--- a/tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
+++ b/tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
@@ -530,7 +530,7 @@ class Bolt_Boltpay_Block_Checkout_BoltpayTest extends PHPUnit_Framework_TestCase
         $actualPaymentOnlyKey = $this->app->getStore()->getConfig('payment/boltpay/publishable_key_onepage');
 
         $this->currentMock->getPublishableKeyForThisPage();
-        $this->assertFalse($actualMultiStepKey || $actualPaymentOnlyKey);
+        $this->assertFalse($actualMultiStepKey || $actualPaymentOnlyKey, "No key should be configured: multi [$actualMultiStepKey], paymentOnly [$actualPaymentOnlyKey]");
         $this->assertEquals($multiStepKey, $actualMultiStepKey);
         $this->assertEquals($paymentOnlyKey, $actualPaymentOnlyKey);
     }
--- a/tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
+++ b/tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
@@ -7,6 +7,9 @@ require_once('TestHelper.php');
  */
 class Bolt_Boltpay_Block_Checkout_BoltpayTest extends PHPUnit_Framework_TestCase
 {
+    /**
+     * @var Mage_Core_Model_App Used to manipulate the Magento application environment
+     */
     private $app = null;
 
     /**
@@ -287,20 +290,16 @@ class Bolt_Boltpay_Block_Checkout_BoltpayTest extends PHPUnit_Framework_TestCase
     }
 
     /**
-     * @test
-     * that a proper product key is returned for every config or an exception is thrown
-     *
-     * @covers ::getPublishableKeyForThisPage
-     * @dataProvider getPublishableKeyForThisPageProvider
+     * Sets dependencies used by test for {@see Bolt_Boltpay_Block_Checkout_Boltpay::getPublishableKeyForThisPage()}
      *
      * @param string|null $multiStepKey             the multi-step publishable key, empty string or null indicates not set
      * @param string|null $paymentOnlyKey           the payment-only publishable key, empty string or null indicates not set
      * @param string      $routeName                The magento path route name
      * @param string      $controllerName           The magento path controller name
-     * @param string|null $predictedReturnValue     The key that is expected to be returned by the mock or null if not returned
+     *
+     * @throws Exception if there is a failure retrieving the focus mock's Request object
      */
-    public function getPublishableKeyForThisPage($multiStepKey, $paymentOnlyKey, $routeName, $controllerName, $predictedReturnValue )
-    {
+    private function setUp_getPublishableKeyForThisPage($multiStepKey, $paymentOnlyKey, $routeName, $controllerName) {
         $this->currentMock = $this->getMockBuilder('Bolt_Boltpay_Block_Checkout_Boltpay')
             ->setMethods(array('getPublishableKey'))
             ->disableOriginalConstructor()
@@ -317,21 +316,6 @@ class Bolt_Boltpay_Block_Checkout_BoltpayTest extends PHPUnit_Framework_TestCase
         $this->app->getStore()->setConfig('payment/boltpay/publishable_key_multipage', $multiStepKey);
         $this->app->getStore()->setConfig('payment/boltpay/publishable_key_onepage', $paymentOnlyKey);
 
-        switch ($controllerName) {
-            case 'cart':
-                $routeName = 'catalog';
-                #intentionally falling through because this his handled the same way as catalog/product
-            case 'product':
-                if ($routeName === 'catalog') {
-                    $expectedReturnedValue = $multiStepKey ?: $paymentOnlyKey;
-                    break;
-                }
-                # intentionally falling through because the correct route was not found
-            default:
-                $expectedReturnedValue = $paymentOnlyKey ?: $multiStepKey;
-
-        }
-
         $this->currentMock->expects($this->exactly(2))->method('getPublishableKey')
             ->withConsecutive(
                 [$this->equalTo('multi-page')],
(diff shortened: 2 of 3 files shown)


app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -511,6 +511,36 @@ public function getCartURL()
         return $this->boltHelper()->getMagentoUrl('checkout/cart');
     }
 
+    /**
+     * Returns the One Page or Multi-Page checkout Publishable key depending on the requested
+     * page and which key has been set in the admin.  Cart and Product pages prioritize the multi-step key
+     * and falls back to the payment only key.  All other pages prioritize the payment only key and falls back
```
ryan-timothy-albert: What is the reason for all other pages prioritizing the payment only key

ryan-timothy-albert: Wouldn't you want all pages on a storefront to prioritize multi-step

matt-thomason: I agree with @ryan-timothy-albert here the intuition is to prioritize multi-step as that's our core focus.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -511,6 +511,36 @@ public function getCartURL()
...
+            $controllerName === 'cart'
+            || ($routeName === 'catalog' && $controllerName === 'product')
+                ? $multiStepPublishableKey
+                : $paymentOnlyPublishableKey
+        ;
+
+        if (!$multiStepPublishableKey && !$paymentOnlyPublishableKey) {
+            $noPublishableKeyException = new Bolt_Boltpay_BoltException("No publishable key has been configured.");
+            $this->boltHelper()->logException($noPublishableKeyException);
+            $this->boltHelper()->notifyException($noPublishableKeyException,[],'error');
```
matt-thomason: nit: consistent comma spacing

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
```diff
@@ -255,6 +286,91 @@ public function testGetSaveOrderURL()
...
+                [$this->equalTo('multi-page')],
+                [$this->equalTo('one-page')]
+            )
+            ->willReturnOnConsecutiveCalls(
+                $mockProxy->getPublishableKey('multi-page'),
+                $mockProxy->getPublishableKey('one-page')
+            )
+        ;
+
+        if ($predictedReturnValue) {
```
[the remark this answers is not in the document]
matt-thomason: @mccottry My guess here is that you did this so that you could DRY the function testing here, and it's written out because you can't really test exception and non-exception code with a dataprovider's output. I would much rather have two unit test functions here instead of one for clarity though - one for an exception flow and one for a non-exception. @oldDriver is right in that we should try to avoid logic of this kind in the test.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: @ryan-timothy-albert and @matt-thomason 
While at first pass, it seems that multi-step should be our default, and of course I was thinking that while writing this code, it is very important to ask oneself questions to like, 

_**"hmm... the code originally is defaulting to one-page/payment only.  I wonder why this was writtenin  that way and what would be the consequences of me changing this default and deploying it to existing merchants?"**_

https://github.com/BoltApp/bolt-magento1/blob/4eecb3edf0fd0c1571bf66f360e10681e763c556/app/design/frontend/base/default/template/boltpay/track.phtml#L31-L33

**So, I am going to weigh heavy on the fact that the logic is longstanding in production without issue before trying to change that logic and deploy it to our current merchants.**

If we wish to explore changing this logic, I move that we pursue it in a different pull request where the risk and impact are properly assessed as changing this logic is not the focus of this PR.  **For now, I believe that it should suffice that if no payment-only publishable key is issued to the merchant, then the multi-step key will always be used which is in alignment with our core focus.**

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: @matt-thomason 
Yep.

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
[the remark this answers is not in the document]
mccottry: I agree with this assessment and the implementation was bothering me as I was thinking about a unit test for a unit test, which absolutely makes no sense.

@matt-thomason You guess is correct and preference is what will now be implemented.

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
```diff
@@ -618,4 +734,36 @@ public function getQuoteData()
...
         );
     }
+
+    /**
+     * Creates diverse scenarios of data for {@see Bolt_Boltpay_Block_Checkout_BoltpayTest::getPublishableKeyForThisPage()}
+     *
+     * @return array[] in the format of [$multiStepKey, $paymentOnlyKey, $routeName, $controllerName, $predictedReturnValue]
+     */
+    public function getPublishableKeyForThisPageProvider() {
+        return [
```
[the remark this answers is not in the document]
mccottry: I have considered this and do not like this approach.  I do like the named parameter aspect of this and I will implement an alternative that addresses my concerns of both clarity of reading in the test and data provider.

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
```diff
@@ -511,6 +511,39 @@ public function getCartURL()
...
+        ;
+
+        if (!$multiStepPublishableKey && !$paymentOnlyPublishableKey) {
+            $noPublishableKeyException = new Bolt_Boltpay_BoltException("No publishable key has been configured.");
+            $this->boltHelper()->logException($noPublishableKeyException);
+            $this->boltHelper()->notifyException($noPublishableKeyException, [], 'error');
+            throw $noPublishableKeyException;
+        }
+
+        return $thisPagesPublishableKey ?: max($multiStepPublishableKey, $paymentOnlyPublishableKey);
```
oldDriver: using max function has no sense. This is like random choice.

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
```diff
@@ -2,17 +2,34 @@
 
 require_once('TestHelper.php');
```
oldDriver: Please remove this

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
```diff
@@ -25,9 +42,26 @@ public function setUp()
             ->getMock()
         ;
 
+        Mage::unregister('_helper/boltpay');
+        $this->helperMock = $this->getMockBuilder('Bolt_Boltpay_Helper_Data')
+            ->setMethods(
+                ['notifyException', 'logException']
```
oldDriver: we use "array()" in the project. It's better to follow one style.

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
```diff
@@ -255,6 +289,401 @@ public function testGetSaveOrderURL()
...
+     * Provides data for a multi-step key, or a payment-only key, or for both publishable keys in the context of
+     * the standard cart page, a custom cart page, the standard checkout page, a standard product page, and any
+     * unexpected page identified by route and controller name for
+     * {@see Bolt_Boltpay_Block_Checkout_BoltpayTest::getPublishableKeyForThisPage_withAtLeastOneKeyConfigured()}
+     *
+     * @return array[] in the format of [$multiStepKey, $paymentOnlyKey, $routeName, $controllerName, $predictedReturnValue]
+     */
+    public function getPublishableKeyForThisPage_withAtLeastOneKeyConfiguredProvider()
+    {
+        return [
```
oldDriver: Please use "array()"

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
```diff
@@ -255,6 +289,401 @@ public function testGetSaveOrderURL()
...
+     *
+     * @param string|null $multiStepKey             the multi-step publishable key, empty string or null indicates not set
+     * @param string|null $paymentOnlyKey           the payment-only publishable key, empty string or null indicates not set
+     * @param string      $routeName                The magento path route name
+     * @param string      $controllerName           The magento path controller name
+     * @param string      $expectedReturnValue      The key that is expected to be returned by the mock
+     *
+     * @throws Exception if there is a problem in setting up test dependencies
+     */
+    public function getPublishableKeyForThisPage_withAtLeastOneKeyConfigured($multiStepKey, $paymentOnlyKey, $routeName, $controllerName, $expectedReturnValue)
```
oldDriver: It's better to use one variable< for example $case as array. 
The original method has no such parameters.

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
```diff
@@ -255,6 +289,401 @@ public function testGetSaveOrderURL()
...
+    public function getPublishableKeyForThisPage_withAtLeastOneKeyConfiguredProvider()
+    {
+        return [
+            [
+                # When both keys exist on standard cart page, multi-step is used
+                "multiStepKey" => "multi+payOnly-key-on-standard-cart",
+                "paymentOnlyKey" => "payOnly+multi-key-on-standard-cart",
+                "routeName" => "checkout",
+                "controllerName" => "cart",
+                "predictedReturnValue" => "multi+payOnly-key-on-standard-cart"
```
oldDriver: Usually in best practice used key 'expect' or 'expects'

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
```diff
@@ -255,6 +289,401 @@ public function testGetSaveOrderURL()
...
+     * @expectedExceptionMessage No publishable key has been configured.
+     *
+     * @param string|null $multiStepKey             the multi-step publishable key, empty string or null indicates not set
+     * @param string|null $paymentOnlyKey           the payment-only publishable key, empty string or null indicates not set
+     * @param string      $routeName                The magento path route name
+     * @param string      $controllerName           The magento path controller name
+     *
+     * @throws Exception if there is a problem in setting up test dependencies
+     */
+    public function getPublishableKeyForThisPage_whenNoKeyIsConfigured($multiStepKey, $paymentOnlyKey, $routeName, $controllerName)
```
oldDriver: The same, use one case variable

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
```diff
@@ -255,6 +289,401 @@ public function testGetSaveOrderURL()
...
+        $this->setUp_getPublishableKeyForThisPage($multiStepKey, $paymentOnlyKey, $routeName, $controllerName);
+
+        $this->helperMock->expects($this->once())->method('logException');
+        $this->helperMock->expects($this->once())->method('notifyException');
+
+        $actualMultiStepKey =  $this->app->getStore()->getConfig('payment/boltpay/publishable_key_multipage');
+        $actualPaymentOnlyKey = $this->app->getStore()->getConfig('payment/boltpay/publishable_key_onepage');
+
+        $this->currentMock->getPublishableKeyForThisPage();
+        $this->assertFalse($actualMultiStepKey || $actualPaymentOnlyKey);
```
oldDriver: Bro, you made me cry!!!
Unit test could not contains logic!

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
mccottry: I will humor these comments as it may prove valuable for you and for others of our team who may wish to speed learn PHP and our M1 plugin.

First, @oldDriver, it is not that the code does not make sense, it is that you do not understand it.  In that situation, there is no problem asking for explanation and please do, as it is preferred.  

The code does concisely what it is suppose to do. Here we can only have three cases:

1.) Both publishable keys are configured and the preferred one was set to `$thisPagesPublishableKey`;
2.) Only the preferred publishable key was configured and is set to `$thisPagesPublishableKey`.
3.) Only the non-preferred publishable key was configured and `$thisPagesPublishableKey` is an empty string,`''`, or `null`

If both keys were either null or empty, the  `$noPublishableKeyException` would have been thrown.

In case #1 and #2, `$thisPagesPublishableKey` is returned with the preferred key.

In case #3, we know that the non-preferred key is a string and what we now want to return and we also know that the preferred string set to `$thisPagesPublishableKey` is an empty string or `null`.  The `max` function will now compare the keys.  Because PHP type juggles, it will now convert `null` to and empty string to compare the same types.

https://www.php.net/manual/en/language.types.string.php#language.types.string.casting

The `max` documentation is very clear about its usage in string comparisons. 

image

https://www.php.net/manual/en/function.max.php

So, the end result will the non-empty string will be returned, which we know to be the non-preferred key, which is concisely what we are after.

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
mccottry: `require_once('TestHelper.php');` was not added in this pull request and would fall under the category of an unrelated change, which we try very hard not to do.  I will not add that here, but will create a separate PR to remove this if I find that it is unnecessary.  You may also do the same.

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
mccottry: Actually, in our code base, we have been using both the old 5.3 and below syntax, `array()` and the 5.4 and above syntax of `[]` for a long time.  One example is in the Product page checkout code that you are familiar with.

https://github.com/BoltApp/bolt-magento1/blob/master/app/code/community/Bolt/Boltpay/Block/Catalog/Product/Boltpay.php#L67-L81

Our coding standards are enforced by the official Magento 1 MEQP1 rule set at severity Level 10.  This allows us to be officially allowed in the Magento 1 Marketplace.  It does not restrict the use of short array syntax.

https://github.com/magento/marketplace-eqp

At any rate, we can unify the syntax in a separate PR if it is for some reason deemed important.

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
mccottry: https://github.com/BoltApp/bolt-magento1/pull/571#discussion_r354516020

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
mccottry: Actually, I prefer the style which I am using.  It is more clear in the test and in the documentation, especially with being able to declare types.  The signature of the original method is not relevant here.

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
mccottry: Thanks.  I actually wanted to name this the same as I did in the test parameters, which is `expectedReturnValue`.  This, I have updated!

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
mccottry: https://github.com/BoltApp/bolt-magento1/pull/571#discussion_r354529022

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
mccottry: Hmmm.  There is probably no reason to cry here, especially if we accustomed to recognizing some general characteristics of `&&` and `||`.

Let's say that I have four conditions, `$cond1`, `$cond2`, `$cond3`, and `$cond4`.  Now, I want to **assert that all the conditions are false** in order for the test to pass.

Then the follow two approaches accomplish the same thing --

- Long Approach - 
```
$this->assertFalse($cond1);
$this->assertFalse($cond2);
$this->assertFalse($cond3);
$this->assertFalse($cond4);
```

- Short Approach -
```
$this->assertFalse($cond1 || $cond2 || $cond3 || $cond4);
```

And if I were interested in asserting that all the conditions were true, I could do this:
```
$this->assertTrue($cond1 && $cond2 && $cond3 && $cond4);
```

If I wanted to be verbose, I could add a message to the assertion:
```
$this->assertFalse($cond1 || $cond2 || $cond3 || $cond4, "All should be false: $cond1,$cond2,$cond3,$cond4, " );
```

app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php
oldDriver: @mccottry just one dummy question: Could merchant start with Multistep key and later add OnePage step?
What will be in this case?
Great thanks for the "php.net" link! never see before

mccottry: @oldDriver, @matt-thomason, @ryan-timothy-albert 

> @mccottry just one dummy question: Could merchant start with Multistep key and later add OnePage step?
> What will be in this case?

In this case, on the shopping cart page, and all product pages, the multi-step key is presented.  For all other cases, it will be assumed that it is a one-page checkout and so that key will be presented.  Also note, that in current use, this only is used for track.js and not connect.js.

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
mccottry: will not change

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
mccottry: will not change

tests/unit/testsuite/Bolt/Boltpay/Block/Checkout/BoltpayTest.php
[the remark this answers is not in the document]
mccottry: will not change

PR #703, opened 2020-03-13
Remove BoltCheckout.configure from check callback in Admin and Firecheckout

mccottry:
# Description
Recently, changes were made to BoltCheckout.configure which breaks the use of calling this method from the check callback.  We therefore take an alternative approach to deferred configuration for Admin and Firecheckout where we must first validate the forms before BoltCheckout.configure can can be provided with proper data.

Note: This is a functional implementation of the request to be available at the soonest release and it is not for style points.  This code will be refactored in the future to simplify it's understanding.

Fixes: https://app.asana.com/0/544708310157130/1164240591833289

#changelog Remove BoltCheckout.configure from check callback in Admin and Firecheckout

# Type of change

- [ ] Test
- [x] Refactor

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/app/code/community/Bolt/Boltpay/Helper/Data.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Data.php
@@ -96,10 +96,12 @@ class Bolt_Boltpay_Helper_Data extends Mage_Core_Helper_Abstract
     }
 
     /**
-     * @param      $checkoutType
+     * @param string $checkoutType
      * @param bool $isVirtualQuote
      *
      * @return string
+     *
+     * @throws Mage_Core_Model_Store_Exception
      */
     public function buildOnCheckCallback($checkoutType, $isVirtualQuote = false)
     {
@@ -124,7 +126,7 @@ class Bolt_Boltpay_Helper_Data extends Mage_Core_Helper_Abstract
                         resolvePromise(response.responseJSON.cart_data);
                     }                   
                 },
-                 onFailure: function(error) { rejectPromise(error); }
+                onFailure: function(error) { rejectPromise(error); }
             }); 
         ";
         switch ($checkoutType) {
--- a/app/code/community/Bolt/Boltpay/Helper/Data.php
+++ b/app/code/community/Bolt/Boltpay/Helper/Data.php
@@ -62,10 +62,8 @@ class Bolt_Boltpay_Helper_Data extends Mage_Core_Helper_Abstract
 
         return "{
                   check: function() {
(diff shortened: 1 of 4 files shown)


app/code/community/Bolt/Boltpay/Helper/Data.php
```diff
@@ -98,13 +96,39 @@ public function getBoltCallbacks($checkoutType, $isVirtualQuote = false)
...
+        $ajaxCall = "
+            new Ajax.Request('$checkoutTokenUrl', {
+                method:'post',
+                parameters: '',
+                onSuccess: function(response) {
+                    if(response.responseJSON.error) {                                                        
+                        rejectPromise(response.responseJSON.error_messages);
+                        
+                        // BoltCheckout is currently not doing anything reasonable to alert the user of a problem, so we will do something as a backup
+                        alert(response.responseJSON.error_messages);
```
matt-thomason: What does this alert look like, do you have a screenshot?
I think the lack of external messaging is an intentional decision, cc:ing @daisy1754 here for his opinion.

daisy1754: yeah let's not do alert - alert is such an unpolished UI. what are the example of errors that "please try again" don't fit well? we can chat with product to define some error type. As Matt said not showing error message as is probably intentional decision, as often these errors are not necessarily user friendly

mccottry: I considered a few things and had questions for one of them.

@daisy1754, @matt-thomason  
- What does Bolt now do by default?  When this code was first written, Bolt did nothing so, I thought to keep it.  But, not showing anything is an option.

- Regarding unpolished, it is actually the same interface used by standard magento, so it will appear to be common to the end user.

- We have our own pop-up that we can use.  As is, this is a let's make Mike A. happy PR on a tight timeline.  We can use the same popup that we use for refunds if there is suitable message.

mccottry: ok.  Then we will give no indication to the merchant at this time.  We can revisit this at a future date if we decide we wish to inform the merchant with a user-friendly popup.

daisy1754: oh i thought this PR is for firecheckout. for admin I think it's ok to use alert but firecheckout I'd vote against. we actually had merchant complained us in the past about using alert saying because it's not sophisticated ux

PR #710, opened 2020-03-14
100% code coverage

adentong:
Removed an unused method that was ignored by code coverage. Also rewrote a line in ShippingController because it was never executed before since exitImmediately defaults to true. Also added the missing test for cron.

#changelog 100% code coverage

# Type of change

- [ ] Test

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/app/code/community/Bolt/Boltpay/Model/Cron.php
+++ b/app/code/community/Bolt/Boltpay/Model/Cron.php
@@ -42,6 +42,8 @@ class Bolt_Boltpay_Model_Cron
      *
      * As an artifact, we leave the parent quotes, (i.e. all Magento created quotes), and converted
      * immutable quotes. We delegate cleanup responsibility of these to the merchants.
+     *
+     * @codeCoverageIgnore Not really testable. For one thing sfo.entity_id is the primary key, so should never be NULL.
      */
     public function cleanupQuotes() {
         try {
--- a/app/code/community/Bolt/Boltpay/Model/Cron.php
+++ b/app/code/community/Bolt/Boltpay/Model/Cron.php
@@ -43,7 +43,7 @@ class Bolt_Boltpay_Model_Cron
      * As an artifact, we leave the parent quotes, (i.e. all Magento created quotes), and converted
      * immutable quotes. We delegate cleanup responsibility of these to the merchants.
      *
-     * @codeCoverageIgnore Not really testable. For one thing sfo.entity_id is the primary key, so should never be NULL.
+     * @codeCoverageIgnore Temporarily ignoring to reach 100%. Very difficult to test.
      */
     public function cleanupQuotes() {
         try {
--- a/app/code/community/Bolt/Boltpay/Model/Cron.php
+++ b/app/code/community/Bolt/Boltpay/Model/Cron.php
@@ -42,8 +42,6 @@ class Bolt_Boltpay_Model_Cron
      *
      * As an artifact, we leave the parent quotes, (i.e. all Magento created quotes), and converted
      * immutable quotes. We delegate cleanup responsibility of these to the merchants.
-     *
-     * @codeCoverageIgnore Temporarily ignoring to reach 100%. Very difficult to test.
      */
     public function cleanupQuotes() {
         try {
--- a/app/code/community/Bolt/Boltpay/controllers/ShippingController.php
+++ b/app/code/community/Bolt/Boltpay/controllers/ShippingController.php
@@ -126,7 +126,11 @@ class Bolt_Boltpay_ShippingController
                 benchmark('Notified bugsnag of address error');
                 $this->sendResponse(
                     422,
-                    array('status' => 'failure','error' => $addressErrorDetails)
+                    array('status' => 'failure', 'error' => $addressErrorDetails),
+                    call_user_func(function () {
+                        benchmark('Sending shipping address failure to Bolt');
+                        return true;
+                    })
                 );
             }
             ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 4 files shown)


app/code/community/Bolt/Boltpay/Model/Cron.php
```diff
@@ -42,6 +42,8 @@ class Bolt_Boltpay_Model_Cron
      *
      * As an artifact, we leave the parent quotes, (i.e. all Magento created quotes), and converted
      * immutable quotes. We delegate cleanup responsibility of these to the merchants.
+     *
+     * @codeCoverageIgnore Not really testable. For one thing sfo.entity_id is the primary key, so should never be NULL.
```
mccottry: @adentong This is a decent temporary solution.   I will write a unit test for this next week.

app/code/community/Bolt/Boltpay/controllers/ShippingController.php
```diff
@@ -128,7 +128,6 @@ public function indexAction()
                     422,
                     array('status' => 'failure','error' => $addressErrorDetails)
```
mccottry: Great insight!
To preserve the log, if important at all, perhaps we could do this
```suggestion
                    422,
                    array('status' => 'failure','error' => $addressErrorDetails)	                    array('status' => 'failure','error' => $addressErrorDetails),
                   call_user_func(function(){benchmark('Sending shipping address failure to Bolt');return true;})
```

app/code/community/Bolt/Boltpay/Model/Cron.php
adentong: Ah yes you're definitely right.

adentong: Another difficulty I found is whenever I try to update the updated time to more than two weeks ago it gets automatically updated to the current time. I couldn't figure out how to disable the automatic updating of updated time.

roopakv:
for what it is worth, having 100% test coverage is not something that is every truly possible, and 100% is also never really true, because of the way tools test. While it is ok to not count tests for a specific set of files, i think this isnt a solution we should pursue in the long term. We should rather understand that some things cant be tested and try stay above 80%.

Adding files to not testable will invariably lead to someone else adding another file to non testable just because they want to maintain 100% (i've seen this happen) and this isnt really useful, and more likely a pain.

adentong:
Yea I know. It's just that we've gotten so close to 100% that I'm just kinda bothered by it lol. Also for what it's worth this is I think the only ignored method that's not deprecated or unused.

mccottry:
@adentong and @roopakv 
Using `@codeCoverageIgnore` is unnecessary.  And after reading Roopak's comments, I will backtrack at my previous comment calling it a "good temporary solution".  I was trying to be diplomatic.  Using these sort of tricks will necessarily cheapen the monumental effort that it took us to reach 99.8% code coverage without sweeping things under the rug to get an artificial number.

As I mentioned in the comment above, early next week, I will complete the test for this final method.  Therefore, when we announce that we were able to bring to be M1 the first of many more Bolt plugins to follow to reach 100% code coverage, it will be a legit claim.  Whether or not it is truly possible will remain a question of if one believes in the testing platforms and how they report, and not our implemented test which legitimately plows through every conditional path presented, deliberately.  The only question that could remain is whether we tested thoroughly enough, (i.e. presented all the relevant test case scenarios that could hit the code.)

Which brings me to the main point that I have said from the start, and Roopak is rightly bringing up here: **Unit test must be useful.**  Numbers mean nothing if we are not confident that we actually tested and confirmed what we were trying to do. A 99.8% score for meaningful test trumps a 100% score for padded results.

app/code/community/Bolt/Boltpay/Model/Cron.php
```diff
@@ -42,6 +42,8 @@ class Bolt_Boltpay_Model_Cron
      *
      * As an artifact, we leave the parent quotes, (i.e. all Magento created quotes), and converted
      * immutable quotes. We delegate cleanup responsibility of these to the merchants.
+     *
+     * @codeCoverageIgnore Temporarily ignoring to reach 100%. Very difficult to test.
```
mccottry: Let's remove this.  We don't need to get to 100% this way.  We'll do it the difficult way so that our 100% means 100%, and not 100%*

app/code/community/Bolt/Boltpay/Model/Cron.php
mccottry: I would imagine that if no triggers were defined in the DB, a direct query would be the approach to solve this.  You probably tried that, but, I will keep this in mind if I get to solving this on Monday before you do.

app/code/community/Bolt/Boltpay/Model/Cron.php
adentong: Got it.

adentong:
@mccottry There's no need to be diplomatic with me. I would in fact really aporeciate it if you could call out any changes you feel are unnecessary or inappropriate, such as in this case. Also I'm greatly interested to see how you will tackle this final unit test given the difficulties I mentioned above.

app/code/community/Bolt/Boltpay/Model/Cron.php
adentong: Removed.

app/code/community/Bolt/Boltpay/Model/Cron.php
adentong: I have in fact not tried a direct query. Thanks for informing me of that possibility.

adentong:
@mccottry After reading your comments I think I can try to figure out that last test case by myself. My bad for trying to take the easy way out.

mccottry:
> @mccottry After reading your comments I think I can try to figure out that last test case by myself. My bad for trying to take the easy way out.

@adentong  A big Bolt to you.  If you don't get a chance, I will take care of it tomorrow.

adentong:
@mccottry The last unit test has been added. We've now achieved true, unadulterated 100% test coverage.

mccottry:
> @mccottry The last unit test has been added. We've now achieved true, unadulterated 100% test coverage.

@adentong Awesome work, man!  You did it.  100% here means 100%... no asterisk necessary!

